Issue #19, opened 2020-09-08
response schema openapi


thelittlebug:
the docs state that the response schema gets documented but im unable to see it.

am i doing something wrong?

begyy:
I have this problem too :( 
![image](https://user-images.githubusercontent.com/44405438/96616264-c7ad4c00-131b-11eb-8d15-af8f71da60cc.png)
![image](https://user-images.githubusercontent.com/44405438/96616287-d09e1d80-131b-11eb-9054-a62603210ab4.png)
@vitalik did not fix this problem?

vitalik:
fixed

Please use `pip install django-ninja==0.8.0`

PR #22, opened 2020-10-03
Add response schema for swagger

fojetin:
Fix for #19


fojetin:
@vitalik

vitalik:
Hi @fojetin 

Unfortunately I will not have much time coming week, would you be able to add few test cases for this PR ?
(like create operation with response model and then check if openapi schema resulted expected details)

Thanks

Issue #48, opened 2020-12-07
Hide validation errors in prodaction env


fojetin:
Hi! I need to hide {"detail": errors} on prod env. It's error after `_get_values(request, kw)`

I think, we should add some param to NinjaAPI, like hide_errors, to hide detail in error.
@vitalik do you approve this solution?

vitalik:
this will be possible with custom response classes (along with #6)

vitalik:
@fojetin 

This is now possible with [response renderers](https://django-ninja.rest-framework.com/tutorial/response-renderers/)

```Python
from ninja.renderers import JSONRenderer

class MyRenderer(JSONRenderer):

    def render(self, request, data, *, response_status):
        if not settings.DEBUG and response_status == 422:
                data = {"error": True}
        return super().render(request, data, response_status=response_status)


api = NinjaAPI(renderer=MyRenderer())

```

Issue #6, opened 2020-07-24
Can I add a default json renderer class?


begyy:
![image](https://user-images.githubusercontent.com/44405438/88390986-ba42fd00-cdd2-11ea-9b79-998f00d92f30.png)
and django also supports ujson and orsjon?  
![image](https://user-images.githubusercontent.com/44405438/88391194-21f94800-cdd3-11ea-8f4d-db9ebcb7aa4f.png)

vitalik:
@begyy 

Yes, it's a good option, it's in my backlog

vitalik:
Hey @begyy 

what are you thoughts...

Do you need to set response class for each individual operation?  `@api.get("", response_class=ORJsonResponse)`

or is it enough to set one global like `api = NinjaAPI(..., json_renderer=UJSON)` - that will handle both deserialising input and serialising output ?

I do not see any case for myself where I would want one  operation to have ujson or other one stdlib json lib

begyy:
Hi @vitalik  better it api = NinjaAPI(..., json_renderer=UJSON)

vitalik:
Hi @begyy 

Please try latest version with support for response renderers

https://django-ninja.rest-framework.com/tutorial/response-renderers/

begyy:
ok

begyy:
@vitalik good

PR #49, opened 2020-12-08
Hide values validation error with param

fojetin:
Solution for #48 issue


vitalik:
thanks for the patch

but have to refuse it...

basically there is a development for the response class parameter (like the issue in #6 )

where you will be able to set this type of logic as well


as for temporary solution you can use meanwhile a middleware that will check - if path is in /api/* and response-code == 422 - it will return some custom response message

PR #57, opened 2020-12-25
fix reverse error under nested namespace

valbendan:
Under nested namespace, we only return last part for register, and use full namespace todo `reverse`.


vitalik:
Thank you

how do I reproduce the error you got before this fix?

valbendan:
I create a django demo app to reproduce the error.

repo link: https://github.com/valbendan/django-ninja-demo

vitalik:
@valbendan 
thank you

PR #64, opened 2021-01-11
Removed usage of broad excepts and some print statements

hannseman:
Thanks for this project! 🌟 

I found some usages of broad excepts where `ValueError` can be used instead. It's nice to avoid too broad excepts when it's not needed, let me know if I'm missing something here. 

I also found a couple of `print` statements, I understand that the exception that's printed is caught and silenced but we should probably try and use the logging module if we want some debugging-output for the user.

### DIFF
--- a/ninja/params.py
+++ b/ninja/params.py
@@ -27,7 +27,6 @@ class Param(FieldInfo):
         # param_type: Any = None,
         **extra: Any,
     ):
-        # print('alias = ', alias)
         self.deprecated = deprecated
         # self.param_name: str = None
         # self.param_type: Any = None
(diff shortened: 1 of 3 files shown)


vitalik:
Hi @hannseman 

thank for pull request, but unfortunately there is a big chunk of new functionality coming in `development` branch, that should land on master any day now


CleanShot 2021-01-12 at 12 40 05

your patch seems not compatible anymore

but I will check and apply it on the dev branch

Thank your for your effort

hannseman:
Your comment disappeared after my force push but I see your point. I've reverted back to catching `Exception` on the body parser.

vitalik:
this is now done in separate functionality

PR #67, opened 2021-01-15
Fix routers inheritance

fojetin:


vitalik:
how to reproduce this bug on old version ?

fojetin:
I added a test for this case

PR #69, opened 2021-01-21
fixed two docs typos

duplxey:
In the `path-params` part of the tutorial I'm pretty it should be `api` decorator instead of `app`.


vitalik:
@duplxey Thank you

PR #71, opened 2021-02-01
Added model.dict() arguments

fojetin:


Adamantish:
Nice! We began using Ninja about 3 weeks ago and I went looking for exactly this feature (`exclude_unset`) yesterday.

Issue #61, opened 2021-01-05
[proposal] router parameters


laith43d:
Hi vitalik,

This project is magical, I cannot imagine myself developing APIs using anything else from now on, I'm actually in the process of migrating two projects to Ninja already! 🙂

I have an enhancement proposal for the router, I repetitively configure each route for auth, tags, response, etc. if it is possible to configure these generally in the router.

Examples:
instead of
```python
staff_router = Router()


@staff.get('/orders',
           tags=['staff'],
           auth=IsAuthorized(staff_auth),
           response={200: List[InventoryOrderSchemaOut]}
           )

@staff.get ...
```

we do
```python
staff = Router(
           tags=['staff'],
           auth=IsAuthorized('STAFF'),
)


@staff.get('/orders',
           response={200: List[InventoryOrderSchemaOut]}
           )
```

vitalik:
Hi @aith43d 

Thank you! glad to hear this


Indeed, I have also experience some issues with with routers tagging

but the approach of dealing with this I think a bit different - the api.add_router should have these options...

for you case it can be something like this:

```Python
from some_module import staff # router

api = NinjaAPI()

api.add_router("/staff", staff, auth=IsAuthorized('STAFF'), tags=["staff"])

```

PR #73, opened 2021-02-09
Router auth

Stamper:
Cool project! That's what I need in FastAPI.

This feature could be very handy in my current project.

Overrides Global auth, but might be overridden by Operation auth.


vitalik:
Hi @Stamper 

Thank you
well there is development going for this issue https://github.com/vitalik/django-ninja/issues/61 
which actually does the same, but I will try to add as well your approach for the Router `__init__` argument for auth as well 

will keep this PR open for now...

vitalik:
done in 0.12

Issue #70, opened 2021-01-25
Operation ids from same function name


vitalik:
```Python
@api.get('/foo')
def operation(request):  #!
     pass

@api.get('/bar')
def operation(request):  #!
     pass
```
^ this is currently a valid and working code, but it generates two identical operation_IDs

it should validate if that happens and throw and error that `operation` id is already taken - pelase rename your function or set id manually

yedpodtrzitko:
Hi,
should this really be a responsibility of the framework? To me it looks like a job for linting/static analysis tools. Both `mypy` or `pylint` (the most common tools for this purpose) are catching this kind of error. Seems good enough to me:

```
$ cat collide.py
def foo(): pass
def foo(): pass

$ mypy collide.py
collide.py:2: error: Name 'foo' already defined on line 1

$ pylint collide.py
collide.py:2:0: E0102: function already defined line 1 (function-redefined)
```

yedpodtrzitko:
> well it is

sorry I'm not sure which things are you answering to - that it is part of framework's responsibilities or the CI tools?

> but technically it still a valid python code..

yes. that's why the tools are able to detect and report such errors which are relatively easy to make.

> larger problem is here actually in operation id's - if they are the same - swagger ui basically blocks you from opening 2nd operation

Maybe I'm missing here some piece of information - is `operation` a magic name which is expected to be present in order for the API docs to be generated? As such, more API endpoints in the same file would need to have the same name to get their docs generated? If not, then again - why not leave this to the CI tools?

vitalik:
> why not leave this to the CI tools?

because people do not do this :) and I already got lot of issues when  functions are named the same


basically there two solutions here:

 - either throw exceptoin or warning when same function name detected
 -  or OpenId operationId in sequence -> `operation, operation_1, operation_2, , operation_3` automatically

vitalik:
> Maybe I'm missing here some piece of information - is operation a magic name which is expected to be present in order for the API docs to be generated?

https://swagger.io/docs/specification/paths-and-operations/

![CleanShot 2021-03-09 at 11 23 41](https://user-images.githubusercontent.com/95222/110448519-f0b35b00-80c9-11eb-91c9-886b1dc3114b.png)

vitalik:
0.12.3 - showing a warning
0.13.0 - will throw an error

PR #80, opened 2021-02-22
Make operation ids unique and raise error on overlap

EmilioCarrion:
This counts as a first contribution so any suggestions with the PR definition are welcomed.

Implementation of the feature described in https://github.com/vitalik/django-ninja/issues/70

Changed the `Operation` class to have a not `None` `operation_id`. It will be the custom one passed as a keyword argument or a generated one using the module and the `view_func` name. 

Added the restriction to check that an API instance can only have one instance of each `operation_id`.

Removed the operation name fallback in the schema generation (the operation now has always an id).

Edited old tests to follow the new restriction.

The biggest doubt about this PR is where to set the tests, I created a test_decorator as as placeholder. Where would be the best location?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 *.pyc
-.venv
+*venv
+*.sqlite3
+
 .vscode
 .mypy_cache
 .coverage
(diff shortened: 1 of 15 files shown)


vitalik:
Hi 

Thank you for PR
Will take a look, but on very first glance I see that you removed public NinjaApi.get_operation_id

Which is used in some projects..

vitalik:
in 0.12.3  django-ninja is show a warning 
in 0.13+ will throw ConfigError

PR #84, opened 2021-02-24
Simplify router methods

jairhenrique:


vitalik:
Hi @jairhenrique 

Thank you for PR
Well... methods done explicitly so that editors can autocomplete all arguments...

but I think `VSCode + Pylance` now should be able to discover this - will give it a try later...
will keep this open for now...

jairhenrique:
Hi @vitalik, I didn't thought about that! 
With a simple test here, I saw the problem. 

I will close this PR.

PR #105, opened 2021-03-18
Add mypy integration

uriyyo:
What is done in the scope of this PR:
* Add mypy integration
* Add missed `__all__` to all modules
* Use `ABC` instead of `NotImplementedError` for abstract classes


vitalik:
@uriyyo 
Thank you I'll review it for the upcoming release

PR #149, opened 2021-06-10
Add support for streaming responses (FileResponse, StreamingHttpResponse)

ErwinJunge:
This PR adds support for streaming responses, specifically FileResponse and StreamingHttpResponse. Tests included. I didn't bother creating an issue first, please let me know if you need me to create an issue.

### DIFF
--- a/ninja/operation.py
+++ b/ninja/operation.py
@@ -16,6 +16,7 @@ from typing import (
 import django
 import pydantic
 from django.http import HttpRequest, HttpResponse, HttpResponseNotAllowed
+from django.http.response import HttpResponseBase
 
 from ninja.constants import NOT_SET
 from ninja.errors import ConfigError, ValidationError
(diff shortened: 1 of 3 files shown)


ninja/operation.py
```diff
@@ -144,7 +149,7 @@ def _result_to_response(self, request: HttpRequest, result: Any) -> HttpResponse
          - if tuple with 2 elements - means http_code + body
          - otherwise it's a body
         """
-        if isinstance(result, HttpResponse):
+        if isinstance(result, (HttpResponse, StreamingHttpResponse)):
```
vitalik, line 152: I think you can use here django.http.HttpResponseBase ?

ErwinJunge:
Thanks for your feedback. I've incorporated it in an updated push.

Additionally, I had to make some typing changes because the mypy test failed. I had to insert 2 `# type: ignore` statements though, because mypy was complaining about things that I didn't know how to tell it weren't actually a problem (missing `streaming_content` from `HttpResponse`, but it's inside a conditional).

Can you re-evaluate?

ErwinJunge:
@vitalik thanks for merging so quickly :)

Issue #147, opened 2021-06-08
Http codes descrptions


vitalik:
currently all that is >= 300 is just "Error":

![CleanShot 2021-06-08 at 18 38 48](https://user-images.githubusercontent.com/95222/121215572-d0419b00-c888-11eb-9d27-2f13c0d88976.png)

would be nice to reuse phrases from [http. HTTPStatus](https://docs.python.org/3/library/http.html)  and/or be able to pass code phrases by the user

PR #158, opened 2021-06-21
Grab http status code phrases from http library

GTBebbo:
This is related to Issue #147.

This PR replaces the status code description with the description in the `http.client` library. This is the same way Django finds HTTP status code descriptions ([Django source](https://github.com/django/django/blob/225d96533a8e05debd402a2bfe566487cc27d95f/django/http/response.py#L133)).

Looking into allowing custom status descriptions, there seems to be no way to override the description Django uses, making the ability to change the description in the OpenAPI schema useless and could encourage incorrect OpenAPI specifications. Feel free to correct me if I'm wrong though!

### DIFF
--- a/ninja/openapi/schema.py
+++ b/ninja/openapi/schema.py
@@ -9,6 +9,8 @@ from ninja.operation import Operation
 from ninja.types import DictStrAny
 from ninja.utils import normalize_path
 
+from http.client import responses
+
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


GTBebbo:
Tests should now be fixed, sorry I didn't see the CONTRIBUTING.md. I must be blind!

vitalik:
Thank you

PR #165, opened 2021-06-29
added paginate decorator

aleksandernovikov:
I really missed pagination in django-ninja.

my first pull request, please do not judge strictly

### DIFF
--- a/decorators/paginate/decorator.py
+++ b/decorators/paginate/decorator.py
@@ -6,6 +6,8 @@ from django.core.paginator import Page, Paginator
 from django.utils.encoding import force_str
 
 
+# based on https://github.com/vitalik/django-ninja/issues/104
+
 def replace_query_param(url, key, val):
     (scheme, netloc, path, query, fragment) = parse.urlsplit(force_str(url))
     query_dict = parse.parse_qs(query, keep_blank_values=True)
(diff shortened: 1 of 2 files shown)


GTBebbo:
This looks nice and clean.

A suggestion could be to also add the ability to override the `per_page` parameter with a GET variable. I know some APIs like to include this as an option. If this suggestion does get added, there will have to be the option to set minimum and maximum values. I'd like to see what @vitalik thinks.

vitalik:
hi

Well I'm keeping this open for now
basically the high-level idea is to provide first some decorator that would allow to add custom parameters to openapi schema 
(like some might want `page` argument, some like `limit` + `offset` argument)

(same for results - same want total results, some next/prev pages)

so my idea is to have some decorator that you can path input and output schemas and it will expose everything to schema and will do the logic

naprsa:
Hello. I want to make an endpoint with pagination. The idea with the decorator is amazing, but in my response there is no information about the current page(if I use the PageNumberPagination class), nor about the total number of pages. From the description of the new feature, I realized that it is possible to configure the outgoing response scheme, but I could not implement it. Can you tell me if this is possible and how to do it?

vitalik:
this is already implemented https://django-ninja.rest-framework.com/tutorial/pagination/

Issue #141, opened 2021-05-19
How to catch exception raised by auth module


icebearrrr:
I wrote an auth module, an raised an exception inside it , but this exception seems not be handled by exception handler,  just throwed directly. When i raised it from view, the handler just works well.

vitalik:
at this moment exceptions can be handled only globally (via api instance)

as an alternative you can make your global exception handler, and then check request.path for router path prefix and do your logic based on that

icebearrrr:
> at this moment exceptions can be handled only globally (via api instance)
> 
> as an alternative you can make your global exception handler, and then check request.path for router path prefix and do your logic based on that

Thanks a lot, i'll try it.

dozer133:
Wouldn't this work?

```python
def _run_authentication(self, request: HttpRequest) -> Optional[HttpResponse]:
        for callback in self.auth_callbacks:
            try:
                result = callback(request)
            except Exception as e:
                return self.api.on_exception(request, e)

            result = callback(request)

            if result:
                request.auth = result  # type: ignore
                return None
        return self.api.create_response(request, {"detail": "Unauthorized"}, status=401)
```

If no handler is defined for that exception, it would be raised by `on_exception`.
This way we can raise exceptions like `InvalidToken` or `RequiredRoleMissing` from within the authentication method which would be nice.

vitalik:
done in 0.14

PR #174, opened 2021-07-16
Handle custom exceptions in _run_authentication

dozer133:
This small patch catches all exceptions raised by the authentication callback and lets `on_exception` handle them.
This allows us to raise custom exceptions and return different responses depending on what went wrong during authentication.

Ref: #141 

I also added response body checking to `test_auth` since we need to check that the custom response comes through.


TaeHyoungKwon:
@vitalik Hello, I'm waiting for this PR to be released. It hasn't been released since June 6th(0.13.2). Could I know when it will be released?

PR #185, opened 2021-07-28
Resolve #178, mixing query param.

### LINKED ISSUE #178: mixing query param bug
CleanShot 2021-07-21 at 15 23 44@2x

this works

but this

CleanShot 2021-07-21 at 15 24 51@2x

not:

CleanShot 2021-07-21 at 15 24 17@2x

stephenrauch:


vitalik:
@stephenrauch  Thank you
Well I'll take a look into this closer - I think there is a chance to use multiple pydantic models for query params

vitalik:
Awesome

PR #187, opened 2021-07-29
Resolve #103, Path params with Django converters

### LINKED ISSUE #103: Path param with converter(s)
Need to support Djaggo path converters
https://docs.djangoproject.com/en/3.1/topics/http/urls/#path-converters

basically this should work:
```Python
@api.get('/test-path/{path:page}')
def test_path(request, page):
    return page
```

stephenrauch:
A basic resolution for Django path converters.  It currently adds [Custom Path Converters](https://docs.djangoproject.com/en/3.2/topics/http/urls/#registering-custom-path-converters) if they are annotated.


ninja/signature/utils.py
```diff
@@ -45,9 +55,36 @@ def make_forwardref(annotation: str, globalns: DictStrAny) -> Any:
...
+def get_path_param_names_types(path: str) -> Dict[str, Union[type, None]]:
+    """turns path string like /foo/{var}/path/{int:another}/end to dict {'var': None, 'another': int}"""
+    names_types = (
+        ([None] + item.strip("{}").split(":"))[-1:-3:-1]
+        for item in re.findall("{[^}]*}", path)
+    )
+    return {name: _path_converter_type(type_) for name, type_ in names_types}
+
+
+def _path_converter_type(converter_name: str) -> Union[type, None]:
```
igoose1, line 67: Hey again, Stephen! I enjoy reading your code lines.

In these cases, is there a special explanation of `Union[T, None]` usage instead of `Optional[T]`?

stephenrauch: I am very new to typing, so my ignorance is the only explanation.

vitalik:
So i've talk with a few companies who uses django-ninja about this PR

and the conclusion is that changing the type of the argument to the type in the path is somewhat unexpected for the users

I think I will first merge this partly - only the path types (and untayped argument will still be processed as string as it is now for the rest of the framework)

stephenrauch:
@vitalik I will gladly make changes for whatever spec is desired.  I however am a little lost.

If I am reading the above correctly, the desire is NOT to support [Django Path Conveters](https://docs.djangoproject.com/en/3.1/topics/http/urls/#path-converters), as requested in the ticket, the ask is to ONLY support the "path" variant of Django Path convertors?  Is the correct reading of the above?

The other variants are redundant with the functionality already provided via typing in the handler signature, except for "feature" inherent in the potential resolution to #203, which is the resolver not matching paths which cannot be coerced into the requested type.

In addition I was interested in supporting Django Path Converters, because my team has a use for the Custom Converters.

vitalik:
@stephenrauch 

Sure.. let me make this a bit more clear


so let's say you have view with argument where you did not annotate the type
```Python
@api.get("/foo/{int:arg}")
def view1(request, arg):
     return {'arg_type', str(type(arg))}  # returns "int" !!!
```

this is the part that conflicts with other parts of the framework  where if you did not annotate type you should expect a `str` type

So I guess the least breaking change would be only to pass path-converted to django url, but do not modify the annotation of the argument

stephenrauch:
So, how about leaving the changes I have now, but adding a check to verify that the view signature type matches the path type?

vitalik:
hm.. still not sure why you need to care about what user annotated ?

I can imagine a pretty usable case like this:
```Python
@api.get("/foo/{uuid:id_arg}")
def my_view(request, id_arg: str):
     return id_arg
```

so here I specifically want that django-url-resolver handles this url only if UUID is passed, but inside my function I'm fine working with id as string

stephenrauch:
@vitalik One minor kink, the data returned by the resolver when using the Django converters is of the type indicated.  EG: UUID has been already converted to a UUID.  Pydantic was not happy to convert that to a str (hmm..) Wasn't sure how best to deal with that, so added a conversion in `PathModel::get_request_data()`

Thoughts?

PR #188, opened 2021-07-30
Fix add_router() to router already attached to API

stephenrauch:
After a router is attached to an api, attaching further routers to the attached router fails.


stephenrauch:
Hi, @vitalik 

Is there something this PR is missing, or other changes that should be made?

Thanks.

vitalik:
I'm preparing/testing new release (just waiting feedback from few companies)
after that will start processing next PRs

PR #197, opened 2021-08-10
Authentication AND/OR operators

### LINKED ISSUE #196: Authentications with logical operators
Hi, ninja is really cool project, thank you for that!

The issue: swagger allows us to combine different authentications using logical operators, but ninja allows us to use them only with OR operator. (using the first suitable authentication). But sometimes we need to use all authentications, for example, I need to use secret authentication for all queries AND Bearer authorization for most endpoints. Could you please add this functionality. 

Now I am using middlewares for these purposes, but they don't appear in swagger :(

https://github.com/vitalik/django-ninja/blob/4e7fc53e57f98c21bf63da7d6b00688dee6185c0/ninja/operation.py#L137
the reference in source code, where the auth checking comes in

igoose1:
That closes #196.

After merging this will be possible:

    @api.get(
        "/",
        auth=(Auth1() & Auth2()) | Auth3()),
    )
    def f(request):
        pass

What I don't like at this moment (marked are solved):

- [ ] `AuthBase.__call__` now is not abstractmethod, I used it for recursive descent. The previous method is called "`callable`" and it's not the best name.

- [ ] Swagger docs is not built for operators. Must be something like this: https://swagger.io/docs/specification/authentication/#multiple

- [x] Django-ninja docs aren't yet written.

- [ ] No clear error messages when trying to pair with "just callable".

The PR is in progress. Comments are welcome!

### DIFF
--- a/docs/src/tutorial/authentication/multiple01.py
+++ b/docs/src/tutorial/authentication/multiple01.py
@@ -1,4 +1,4 @@
-from ninja.security import APIKeyQuery, APIKeyHeader
+from ninja.security import APIKeyQuery, APIKeyHeader, HttpBearer
 
 
 class AuthCheck:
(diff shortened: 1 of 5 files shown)


igoose1:
Note: it's impossible to unite `typing.Callable` and `AuthBase`. That's, and in order of compatibility, why the "old" way is kept as it was before. I've just added some tests, there were none of them testing the sequence in `auth=...`.

Issue #146, opened 2021-06-04
v1 changes


vitalik:
Outlining here potential backward incompatible changes that might be introduced in v1 release

 - auth will be class interface instead of callable (to support async authenticators)
 - responses to support codes/headers/content (like general Response class)
 - [routers paths joined with "/"](https://github.com/vitalik/django-ninja/blob/56ad4d614c6743197f3b1babefe23e0059ef9f74/ninja/router.py#L345) - which might not needed on some cases where router prefix will act like a prefix and not subfolder

ahmadaidin:
is there any plan to add custom decorator to the api module?

vitalik:
> is there any plan to add custom decorator support to the api module?

what do you mean ? example ?

ahmadaidin:
> > is there any plan to add custom decorator support to the api module?
> 
> what do you mean ? example ?

I tried to implement an async decorator to check whether user exists or not on database before an API function is called. At first I got an error and I think it's impossible to add any async decorator for now. But then I found a way to create one. So, I think it's already solved.

MathiasAG:
When i add List[ModelSchema] in another schema, he generate an error "AttributeError: __fields__". I don't understand why

WarKaa:
Hey @vitalik thank you for your great work

Will it be possible in v1 to add custom example to our open api schemas like in https://fastapi.tiangolo.com/tutorial/schema-extra-example/?h=example#pydantic-schema_extra ?

rngallen:
> When i add List[ModelSchema] in another schema, he generate an error "AttributeError: __fields__". I don't understand why
> 

Kindly share your schema and models code. It will be easy to solve having some hints

A-Asror:
will there be permission, for example, like [django rest framework ](https://www.django-rest-framework.org/api-guide/permissions/)

SaluteGF:
a error “TypeError: __modify_schema__() missing 1 required positional argument: 'field'
[28/Oct/2022 10:11:47] "GET /api/openapi.json HTTP/1.1" 500 254519
”，when i use “file: UploadedFile = File(...)”

PR #199, opened 2021-08-14
Add Authorization with Permissions

### LINKED ISSUE #156: [proposal] Authorization ( permissions ) handling
In the docs I can see how to handle Authentication but not how to handle Authorization ( permissions )
After I have identified a logged user I want to return 403 if that user hasn't enough permission to do it
DRF has a great way to handle it, using [BasePermission.has_permission and BasePermission.has_object_permission](https://www.django-rest-framework.org/api-guide/permissions/)

Maybe this library could handle permissions with something like this:

```python

from ninja import Schema
from ninja.permissions import BasePermission

class CanSeeFruitsPerm(BasePermission):
    def has_permission(self, request, view):
        return request.user.has_perm("can_see_fruits")

class AppleOut(Schema):
    color: str
    quality: str
    

@api.get("/apples", auth=BasicAuth(), perm=CanSeeFruitsPerm(), response=List[AppleOut])
def list_apples(request):
    return Apple.objects.all()
```

What do you think about it?

Thank you

mawassk:
This is a straightforward implementation of #156. 

After merging it will be possible to create an own permission class:
```python
from ninja.security import BasePermission

class MyPermission(BasePermission):
    def has_permission(self, request, permission):
        # return falsy type for authorization failure
```
Now it's possible to use this permission on any operation (specify in api, router or method scope):
```python
# ...

@api.get("/ping", perm=MyPermission("permission_string"))
def ping(request):
    return "Pong!"
```
It's also possible to use this in combination with authentication (single and multiple). For a minimal example take a look in the docs :smiley: 

## ToDo's
- [x] Add tests
- [x] Add documentation

Closes #156


vitalik:
Hi @mawassk

Thank you for collaboration, I really appreciate it !

there is one issue with this approach that may be broken at v1

https://github.com/vitalik/django-ninja/issues/146 


Baiscially authentication currently has a bummer - not allowing to use it in async context

so in v1 this will be most likely changed (Most likely authentication will be a subclassisng some base class and implementing some method in both sync and async modes)

mawassk:
> Hi @mawassk
> 
> Thank you for collaboration, I really appreciate it !
> 
> there is one issue with this approach that may be broken at v1
> 
> #146
> 
> Baiscially authentication currently has a bummer - not allowing to use it in async context
> 
> so in v1 this will be most likely changed (Most likely authentication will be a subclassisng some base class and implementing some method in both sync and async modes)

@vitalik thank you for the feedback!

I experimented a bit with `async_to_sync`function and found a simple way to add async support to my feature by adding an extra async base class:
```python
# ...

class BaseAsyncPermission(BasePermission, ABC):
    def __call__(self, request: HttpRequest) -> Optional[Any]:
        sync_has_permission = async_to_sync(self.has_permission)  # type: ignore
        return sync_has_permission(request, self.permission)

    @abstractmethod
    async def has_permission(
        self, request: HttpRequest, permission: str
    ) -> Optional[Any]:
        pass  # pragma: no cover
```
This works really well, but I don't know if it's satisfying enough.

denizdogan:
@mawassk Hi! Was there any particular reason why you chose to withdraw this PR? We are considering using Django Ninja for a project, but the lack of permission handling might be an issue.

Issue #44, opened 2020-12-04
Authentication is never awaited in AsyncOperation


dozer133:
## Example

```python
from asgiref.sync import sync_to_async
from ninja import NinjaAPI, Schema
from ninja.security import APIKeyQuery
from .models import Client

api = NinjaAPI()


class ApiKey(APIKeyQuery):
    param_name = "key"

    async def authenticate(self, request, key):
        try:
            return await sync_to_async(Client.objects.get)(key=key)
        except Client.DoesNotExist:
            pass


class Test(Schema):
    msg: str


@api.get("", response=Test, auth=ApiKey())
async def test(request):
    return 200, {"msg": "hello!"}

```

## Scenario

The API is called with a non-existing API key.
What should happen: Ninja should return a 401 response.
What actually happens: Returns 200.

## Possible solution

Implement _run_checks and _run_authentication as async methods in AsyncOperation.

```python
class AsyncOperation(Operation):
    def __init__(self, *args, **kwargs):
        if django.VERSION < (3, 1):  # pragma: no cover
            raise Exception("Async operations are supported only with Django 3.1+")
        super().__init__(*args, **kwargs)
        self.is_async = True

    async def _run_checks(self, request):
        "Runs security checks for each operation"
        # auth:
        if self.auth_callbacks:
            error = await self._run_authentication(request)
            if error:
                return error

        # csrf:
        if self.api.csrf:
            error = check_csrf(request, self.view_func)
            if error:
                return error

    async def _run_authentication(self, request):
        for callback in self.auth_callbacks:
            result = await callback(request)
            if result is not None:
                request.auth = result
                return
        return Response({"detail": "Unauthorized"}, status=401)

    async def run(self, request, **kw):
        error = await self._run_checks(request)
        if error:
            return error

        values, errors = self._get_values(request, kw)
        if errors:
            return Response({"detail": errors}, status=422)
        result = await self.view_func(request, **values)
        return self._create_response(result)
```

Above example will then return 401 on an invalid key as expected.

vitalik:
Yes, in nutshell this is the code I'm targeting for... 
but there are few things to deal with - like if you have async auth, but **not-a**sync operation or vice-versa

stinovlas:
[...]

PR #202, opened 2021-08-16
added methods for AsyncOperation

ETretyakov:
I have read this issue 44 (https://github.com/vitalik/django-ninja/issues/44), please explain what is needed to solve the issue, this fix is really necessary for our development.

### DIFF
--- a/ninja/operation.py
+++ b/ninja/operation.py
@@ -233,8 +233,29 @@ class AsyncOperation(Operation):
         super().__init__(*args, **kwargs)
         self.is_async = True
 
-    async def run(self, request: HttpRequest, **kw: Any) -> HttpResponseBase:  # type: ignore
-        error = self._run_checks(request)
+    async def _run_checks(self, request):
+        """Runs security checks for each operation"""
+        if self.auth_callbacks:
+            error = await self._run_authentication(request)
+            if error:
+                return error
+
+        if self.api.csrf:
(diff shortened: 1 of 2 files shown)


vitalik:
The main problem is that  you can combine both sync and async callbacks...

so you cannot just await all in a loop and need to determine if callback is async - then await it, otherwise run normally (and if in sync mode you got async authentication - need to run async_to_sync adapter)

ETretyakov:
Do recent commits fix the problem?

ETretyakov:
@vitalik sorry for pushing you, is there any hope to solve the issue?

al-stefanitsky-mozdor:
Agreed with @vitalik, I think it should be separated logic for async and sync views. I mean that you should NOT use sync_to_async at all, and handle such cases outside from the regular Operation class, only in the AsyncOperation.

stinovlas:
I just stumbled upon this when I was trying to create async auth for my API. It seems that right now, there is no way to have an async auth callback for the `NinjaAPI` or `Router` objects. Even the synchronous `django_auth` suggested in documentation does not work for my async views, because it accesses the database. That could be solved by decorating `django_auth` by `sync_to_async` if async auth callbacks were possible. Obviously, router can have both sync and async views and it needs to handle the authentication properly in both cases.

I'm willing to chip in and help this change to happen, but I'd need to know what's wrong with this PR and why it hasn't been moved forward.

Issue #240, opened 2021-10-09
auth not working on router


jhintr:
Sorry, I still have the same problem as #20 , only diff is

```python
api = NinjaAPI(version="1.0.0", auth=django_auth)
```

what returned is:

```python
Exception in thread django-main-thread:
Traceback (most recent call last):
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/zhi/Code/cliche/venv/lib/python3.8/site-packages/django/utils/autoreload.py", line 64, in wrapper
    fn(*args, **kwargs)
  File "/Users/zhi/Code/cliche/venv/lib/python3.8/site-packages/django/core/management/commands/runserver.py", line 118, in inner_run
    self.check(display_num_errors=True)
  File "/Users/zhi/Code/cliche/venv/lib/python3.8/site-packages/django/core/management/base.py", line 419, in check
    all_issues = checks.run_checks(
  File "/Users/zhi/Code/cliche/venv/lib/python3.8/site-packages/django/core/checks/registry.py", line 76, in run_checks
    new_errors = check(app_configs=app_configs, databases=databases)
  File "/Users/zhi/Code/cliche/venv/lib/python3.8/site-packages/django/core/checks/urls.py", line 13, in check_url_config
    return check_resolver(resolver)
  File "/Users/zhi/Code/cliche/venv/lib/python3.8/site-packages/django/core/checks/urls.py", line 23, in check_resolver
    return check_method()
  File "/Users/zhi/Code/cliche/venv/lib/python3.8/site-packages/django/urls/resolvers.py", line 412, in check
    for pattern in self.url_patterns:
  File "/Users/zhi/Code/cliche/venv/lib/python3.8/site-packages/django/utils/functional.py", line 48, in __get__
    res = instance.__dict__[self.name] = self.func(instance)
  File "/Users/zhi/Code/cliche/venv/lib/python3.8/site-packages/django/urls/resolvers.py", line 598, in url_patterns
    patterns = getattr(self.urlconf_module, "urlpatterns", self.urlconf_module)
[...]

PR #242, opened 2021-10-09
Resolve #229 & #240 reporting dev errors under django dev-server

### LINKED ISSUE #229: Looks like you created multiple NinjaAPIs
Hi there, I was playing around a bit with the basic tutorial (it worked perfectly in the beginning) and got this error at some point:

```
  File "/django/django_aia_backend/django_aia_backend/urls.py", line 31, in <module>
    path('api/', include(api.urls)),
  File "/djangovenv/lib/python3.9/site-packages/ninja/main.py", line 314, in urls
    self._validate()
  File "/djangovenv/lib/python3.9/site-packages/ninja/main.py", line 398, in _validate
    raise ConfigError("\n".join(msg))
ninja.errors.ConfigError: Looks like you created multiple NinjaAPIs
To let ninja distinguish them you need to set either unique version or url_namespace
 - NinjaAPI(..., version='2.0.0')
 - NinjaAPI(..., urls_namespace='otherapi')
Already registered: ['api-1.0.0']
```

not sure what is the problem here. I only have one instantiation of NinjaAPI. I tried giving it a specific version name but that does not change anything. 

To be fair, I played around a bit before, among other things creating a django app called 'api' for using standard DRF. I deleted this and every reference to it, but even after starting a brand new django project, the error persists. Does the api instantiation get persisted somewhere in venv? I cannot explain otherwise why this happens.

stephenrauch:
Resolves #229 & #240

When Django loads urls it uses: `django.urls.resolvers.urlconf_module()`

```python
@cached_property
def urlconf_module(self):
    if isinstance(self.urlconf_name, str):
        return import_module(self.urlconf_name)
    else:
        return self.urlconf_name
```

This uses the @cached_property to generally only import once.  But if the import
throws an error when using the development server, the following code in
`django.utils.autoreload.BaseReloader.run()` is used:

```python
# Prevent a race condition where URL modules aren't loaded when the
# reloader starts by accessing the urlconf_module property.
try:
    get_resolver().urlconf_module
except Exception:
    # Loading the urlconf can result in errors during development.
    # If this occurs then swallow the error and continue.
    pass
```

This means the (likely) developer error that caused the Exception is initially ignored. This is
not generally a problem since the error will usually be exercised again, and reported at that
time.  But Ninja has various code which guards against errors where items that cannot be reused,
are attempted to be reused.  This results in Ninja throwing a false error, and hiding the
true error from the developer when running under the development server.

### DIFF
--- a/ninja/main.py
+++ b/ninja/main.py
@@ -24,7 +24,7 @@ from ninja.parser import Parser
 from ninja.renderers import BaseRenderer, JSONRenderer
 from ninja.router import Router
 from ninja.types import TCallable
-from ninja.utils import normalize_path
+from ninja.utils import is_debug_server, normalize_path
 
 if TYPE_CHECKING:
     from .operation import Operation  # pragma: no cover
@@ -387,7 +387,11 @@ class NinjaAPI:
 
         # 1) urls namespacing validation
         skip_registry = os.environ.get("NINJA_SKIP_REGISTRY", False)
-        if not skip_registry and self.urls_namespace in NinjaAPI._registry:
+        if (
+            not skip_registry
+            and self.urls_namespace in NinjaAPI._registry
+            and not debug_server_url_reimport()
+        ):
             msg = [
                 "Looks like you created multiple NinjaAPIs",
                 "To let ninja distinguish them you need to set either unique version or url_namespace",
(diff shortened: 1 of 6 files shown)


pierrevanrensburg:
So what is the resolution for this?  Please provide clear guidance for Django developer @vitalik

PR #266, opened 2021-10-27
Make request object optional on view functions

stephenrauch:
A proposal (in the form of a Pull Request) for allowing request view params to be optional.

### Unchanged Behavior
- params in the first position, named "request" are not in the Schema Docs and are passed the request.

### Current Behavior (to be changed from):
- params in the first position, not named "request", are in the Schema Docs, but are actually passed the request, and if a param with a matching name is actually passed, causes an error.
- params not in the first position, named "request" are not in the Schema Docs, but are passed the param value named request.

### Proposed Behavior (to be changed to):
- params in the first position, not named "request", but annotated with HttpRequest, will be passed the request.
- params in the first position, not named "request" and not annotated with HttpRequest, are in the Schema Docs, and will be passed the matching param value.
- params not in the first position, named "request" are in the Schema Docs, and are passed the param value named request.

### DIFF
--- a/docs/docs/tutorial/index.md
+++ b/docs/docs/tutorial/index.md
@@ -43,7 +43,7 @@ api = NinjaAPI()
 
 
 @api.get("/hello")
-def hello(request):
+def hello():
     return "Hello world"
 
 ```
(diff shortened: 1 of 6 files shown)


vitalik:
Hi @stephenrauch 

I've been thinking about this for some time...
I like it - but I would add more flexibility by allowing also (`request=2` or `request = Body(...)` type of cases)

I merged your request into separate branch and adapted it to cover these as well see #282
Please let me know what you think

vitalik:
Hi @stephenrauch 

Thank you for contribution - but this cannot be introduced - there are lot of existing django ecosystem decorators that people use along with django-ninja - and all those decorators expect 1st param to be always a `request` object

PR #282, opened 2021-11-12
Request optional

vitalik:
Making `request` argument not required

automatically detecting http request argument if:
 - it's a **first** argument AND it's **not annotated** AND **does not have default** value
 - it's annotated with HttpRequest class
 
in all other cases `request` argument will act as any other parameter

### DIFF
--- a/docs/docs/tutorial/index.md
+++ b/docs/docs/tutorial/index.md
@@ -43,7 +43,7 @@ api = NinjaAPI()
 
 
 @api.get("/hello")
-def hello(request):
+def hello():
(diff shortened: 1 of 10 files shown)


al-stefanitsky-mozdor:
Good idea!

SmileyChris:
This looks good 👍🏻

vitalik:
This cannot be introduced - there are lot of existing django ecosystem decorators that people use along with django-ninja - and all those decorators expect 1st param to be always a `request` object

PR #285, opened 2021-11-20
Improve pagination

mom1:
A proposal for allowing use view without `**kwargs`, but with pagination.

- name_param for pagination data
- no need `**kwargs` for use pagination
- by default pagination argument not be passed to view function
- ability to pass a parameter into the view

P.S. I also wanted to express my deep gratitude to you for this library.

### DIFF
--- a/ninja/pagination.py
+++ b/ninja/pagination.py
@@ -1,33 +1,31 @@
 import inspect
 from abc import ABC, abstractmethod
 from functools import wraps
-from typing import Any, Callable, Type
+from typing import Any, Callable, Optional, Tuple, Type
 
 from django.db.models import QuerySet
-from django.http import HttpRequest
 from django.utils.module_loading import import_string
 
 from ninja import Field, Query, Schema
 from ninja.conf import settings
 from ninja.constants import NOT_SET
-from ninja.errors import ConfigError
-from ninja.signature import has_kwargs
 from ninja.types import DictStrAny
 
 
 class PaginationBase(ABC):
+    name_param: str = "pagination"
+    pass_parameter: Optional[str] = None
+
     class Input(Schema):
         pass
 
     InputSource = Query(...)
 
-    def __init__(self, **kwargs: DictStrAny) -> None:
-        pass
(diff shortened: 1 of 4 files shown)


mom1:
@vitalik maybe need something like this?
```python
class PaginationBase(ABC):
    name_param: str = "pagination"
    pass_to_parameters = False
    ...
    def __init__(self, pass_to_parameters: bool = None, **kwargs: DictStrAny) -> None:
        if pass_to_parameters is not None:
            self.pass_to_parameters = pass_to_parameters
```
```python
def view_with_pagination(*args: Tuple[Any], **kwargs: DictStrAny) -> Any:
    paginate_param = kwargs.get(paginator.name_param)
    if not paginator.pass_to_parameters:
        paginate_param = kwargs.pop(paginator.name_param)
    items = func(*args, **kwargs)
        return paginator.paginate_queryset(
            items, **{paginator.name_param: paginate_param}, **kwargs
    )
```

vitalik:
Hi @mom1 
Yeah, I'm still thinking for the best approach here...

I like what you did with `**kwrags` being not required

I think the final can be something like this

 - by default pagination argument maby not be passed to view function
 - and let user a way to pass it with something like:

```Python
@paginate(PageNumberPagination, per_page=50, pass_parameter='pagination')
def my_view(request, pagination):
       return queryset
     
```

mom1:
@vitalik I changed code according to your comment 😉

SmileyChris:
I wonder if it'd be better that the view requests the parameter, rather than the paginate decorator deciding on it. Something like:

```python
@paginate(PageNumberPagination, per_page=50)
def my_view(request, pagination: PageNumberPagination.Input):
       return queryset
```

vitalik:
Yeah, in my practise I never actually used to access kwargs for pagination, and forcing to add it is annoying
@mom1 - this is merged, I just adapted few things and a bit simplified, thank you

----

@SmileyChris
> I wonder if it'd be better that the view requests the parameter, rather than the paginate decorator deciding on it. Something like:
> 
> def my_view(request, pagination: PageNumberPagination.Input):

 yeah, that would be nice, but seems it will involve too much code reshaping

PR #288, opened 2021-11-24
Export OpenAPI schema management cmd

stefanitsky:
Hi!

I want to add a basic management command to export OpenAPI schema.

### DIFF
--- a/docs/docs/tutorial/management-commands.md
+++ b/docs/docs/tutorial/management-commands.md
@@ -17,5 +17,5 @@ Run `python manage.py export_openapi_schema --api project.urls.api`
 ## Options
 
 - `--help` show command help
-- `--api` specify api instance module (required)
+- `--api` specify api instance module (optional)
 - `--output` specify output file path (optional). You can omit this option to output in stdout.
--- a/docs/docs/tutorial/management-commands.md
+++ b/docs/docs/tutorial/management-commands.md
@@ -1,0 +1,21 @@
+# Management Commands
+
+## Export OpenAPI scheme
+
+Add ninja to `INSTALLED_APPS`
+
+```python
+INSTALLED_APPS = [
+    ...
+    'ninja',
+    ...
+]
+```
+
+Run `python manage.py export_openapi_schema --api project.urls.api`
+
+## Options
+
+- `--help` show command help
+- `--api` specify api instance module (required)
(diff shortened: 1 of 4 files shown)


redlovv:
Yes, for me it would also be relevant.

vitalik:
Hi @stefanitsky 

Looks good

just few  notes:

1) I think all this you can replace with [import_string](https://docs.djangoproject.com/en/3.2/ref/utils/#django.utils.module_loading.import_string):
```Python

        path_list = options["api"].split(".")

        if len(path_list) < 2:
            raise CommandError(
                "Module should contains at least package and api instance name attribute!"
            )

        module_path, api_attr = ".".join(path_list[:-1]), path_list[-1]
        try:
            api_module = importlib.import_module(module_path)
        except ModuleNotFoundError as e:
            raise CommandError(f"Module {module_path} not found!")

        api = getattr(api_module, api_attr, None)
```

2) 99% of the cases - people have just one api instance

in that case you can just url-reverse api root module to make `--api` not required (if multiple is defined - you can just throw an error to force user to pick)

stefanitsky:
> Hi @stefanitsky
> 
> Looks good
> 
> just few notes:
> 
> 1. I think all this you can replace with [import_string](https://docs.djangoproject.com/en/3.2/ref/utils/#django.utils.module_loading.import_string):
> 
> ```python
>         path_list = options["api"].split(".")
> 
>         if len(path_list) < 2:
>             raise CommandError(
>                 "Module should contains at least package and api instance name attribute!"
>             )
> 
>         module_path, api_attr = ".".join(path_list[:-1]), path_list[-1]
>         try:
>             api_module = importlib.import_module(module_path)
>         except ModuleNotFoundError as e:
>             raise CommandError(f"Module {module_path} not found!")
> 
>         api = getattr(api_module, api_attr, None)
> ```
> 
> 2. 99% of the cases - people have just one api instance
> 
> in that case you can just url-reverse api root module to make `--api` not required (if multiple is defined - you can just throw an error to force user to pick)

Good notes! I'll fix it, thanks!

stefanitsky:
@vitalik i cant figure out how to fix a problem with a default router, seems to be a problem with multiple router register on each import, so, i think, it can be fixed by specifying some function / class inside `demo_project.demo.urls` and call it just one time (like singleton). What do you think? Any ideas?

image

And also, what do you think now about that implementation with your notes, is it right?

PR #301, opened 2021-12-14
Add support for union type in payload

AkeemMcLennon:
@vitalik This MR adds support for defining a request body as a union of two schemas. 

Prior to this change, a payload with the union type would be incorrectly classified as a query parameter, and fail to appear within the list of components defined in the OpenAPI schema

### DIFF
--- a/ninja/signature/details.py
+++ b/ninja/signature/details.py
@@ -1,12 +1,12 @@
 import inspect
 import warnings
 from collections import defaultdict, namedtuple
-from typing import Any, Callable, Dict, Generator, List, Tuple
+from typing import Any, Callable, Dict, Generator, List, Tuple, Union
 
 import pydantic
 
 from ninja import UploadedFile, params
-from ninja.compatibility.util import get_origin as get_collection_origin
+from ninja.compatibility.util import get_args, get_origin as get_collection_origin
 from ninja.errors import ConfigError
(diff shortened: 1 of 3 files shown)


ninja/signature/details.py
```diff
@@ -241,6 +242,8 @@ def _get_param_type(self, name: str, arg: inspect.Parameter) -> FuncParam:
 
 def is_pydantic_model(cls: Any) -> bool:
     try:
+        if get_origin(cls) == Union:
+            return all([issubclass(arg, pydantic.BaseModel) for arg in get_args(cls)])
```
TaeHyoungKwon, line 246: Hello 😄 

You can delete useless `[]` for list comprehension on using all()

```suggestion
            return all(issubclass(arg, pydantic.BaseModel) for arg in get_args(cls))
```


```python
In [1]: l = [1, 2, 3]

In [2]: all(ele for ele in l)
Out[2]: True

In [3]: all([ele for ele in l])
Out[3]: True
```

vitalik:
Hi @AkeemMcLennon 

you need to to use get_origin from ninja/compatibility/util.py 
(to support python 3.6, 3.7 etc)

AkeemMcLennon:
@vitalik Ah, thanks. Updated

AkeemMcLennon:
Just fixed codestyle linting errors as well

PR #310, opened 2022-01-02
Pydantic 1.9.0 support

antonrh:
Add pydantic 1.9.0 version support
https://pydantic-docs.helpmanual.io/changelog/


MartinNowak:
Hello, could we get a release for this? Python 3.10 support only came with pydantic 1.9, so somewhat stuck on this upper version limit.

MartinNowak:
Hello, could we get a release for this? Python 3.10 support only came with pydantic 1.9, so somewhat stuck on this upper version limit @vitalik.

PR #317, opened 2022-01-09
Smarter schema that handles dotted aliases and resolver methods

### LINKED ISSUE #250: Using a source attribute in schemas?
I'm in the middle of migrating a Django Rest Framework API to Django ninja.

What is the workaround to the `source` keyword used in DRF, which does not seem to exist with Django ninja?

For example for this DRF code:
```python
class RecipeSerializer(serializers.ModelSerializer):
    name = serializers.CharField(source="item.name")
    cooking_tools = CookingToolSerializer(many=True)
```

```python
class RecipeSchema(ModelSchema):
    name: str # Obviously, it can NOT work as the `name` attribute belongs to `Item` model, not a `Recipe` model.
    cooking_tools: List[CookingToolSchema]
```

Any idea? Thanks.

### LINKED ISSUE #291: Foreign key field value instead of id without nesting schemas ?
Hi, is it possible to send a foreign key value without nesting the schemas ? Here's my scenario: 

```python
class Department(models.Model):
    id = models.IntegerField(primarty_key=True, editable=False)
    dep = models.CharField(max_length=200, null=False,unique=True)

 class Student(models.Model):
    id = models.IntegerField(primarty_key=True, editable=False)
    name = models.CharField(max_length=200, null=False)
    dep = models.ForeignKey(Department, on_delete=models.CASCADE)


 class DepartmentOut(Schema):
     dep: str
 

 class StudentOut(Schema):
    id: int
    name: str
    dep: DepartmentOut

```
this will result in an output of: 
```python
{
    id: 1000,
    name: 'Some name',
    dep: {
        dep: 'Some Text'
        }
        
}
```

How can I show the dep string without a nested object called dep? Like the below shape of JSON:

```python
{
    id: 1000,
    name: 'Some name',
    dep: 'Some Text'
    
}
```
I tried searching without any luck, thanks in advance.

SmileyChris:
Fixes #291 (even though it was closed) and fixes #250 and adds in new functionality to allow calculated fields via `resolve_` static methods on the response schema class.

Includes tests and docs.

### DIFF
--- a/docs/docs/tutorial/response-schema.md
+++ b/docs/docs/tutorial/response-schema.md
@@ -90,7 +90,7 @@ class TaskSchema(Schema):
 
 @api.get("/tasks", response=List[TaskSchema])
 def tasks(request):
-    queryset = Task.objects.all()
+    queryset = Task.objects.select_related("owner")
     return list(queryset)
 ```
 
@@ -116,6 +116,49 @@ If you execute this operation, you should get a response like this:
     },
 ]
 ```
+
+
+## Aliases
+
+Instead of a nested response, you may want to just flatten the response output.
+The Ninja `Schema` object extends Pydantic's `Field(..., alias="")` format to
+work with dotted responses.
+
+Using the models from above, let's make a schema that just includes the task
+owner's first name inline, and also uses `completed` rather than `is_completed`:
+
+```Python hl_lines="1 7-9"
+from ninja import Field, Schema
+
+
+class TaskSchema(Schema):
+    id: int
+    title: str
+    # The first Field param is the default, use ... for required fields.
+    completed: bool = Field(..., alias="is_completed)
+    owner_first_name: str = Field(None, alias="owner.first_name")
+```
+
+
+## Resolvers
+
+You can also create calculated fields via resolve methods based on the field
+name.
+
+```Python hl_lines="5 7-11" 
+class TaskSchema(Schema):
+    id: int
+    title: str
+    is_completed: bool
+    owner: Optional[str]
+
+    @staticmethod
+    def resolve_owner(obj):
+        if not obj.owner:
+            return
+        return f"{obj.owner.first_name} {obj.owner.last_name}"
+```
+
 
 ## Returning querysets
 
--- a/docs/docs/tutorial/response-schema.md
+++ b/docs/docs/tutorial/response-schema.md
@@ -145,18 +145,28 @@ class TaskSchema(Schema):
 You can also create calculated fields via resolve methods based on the field
 name.
 
-```Python hl_lines="5 7-11" 
+The method must accept a single argument, which will be the object the schema
+is resolving against.
+
+When creating a resolver as a standard method, `self` gives you access to other
+validated and formatted attributes in the schema.
+
+```Python hl_lines="5 7-11"
 class TaskSchema(Schema):
     id: int
     title: str
     is_completed: bool
     owner: Optional[str]
+    lower_title: str
 
     @staticmethod
     def resolve_owner(obj):
         if not obj.owner:
             return
         return f"{obj.owner.first_name} {obj.owner.last_name}"
+
+    def resolve_lower_title(self, obj):
+        return self.title.lower()
 ```
 
 
--- a/ninja/orm/metaclass.py
+++ b/ninja/orm/metaclass.py
@@ -1,16 +1,15 @@
 from typing import no_type_check
 
 from django.db.models import Model as DjangoModel
-from pydantic.main import ModelMetaclass
 
 from ninja.errors import ConfigError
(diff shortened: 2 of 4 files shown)


SmileyChris:
Fixes #250 too

SmileyChris:
I bumped the pydantic version to 1.9 to support the fancier `DjangoGetter` method being used. The alternative is to backport the pydantic fix to a method on the ninja `Schema` class: 
```
    @classmethod
    def _decompose_class(cls, obj: Any) -> GetterDict:
        if isinstance(obj, GetterDict):
            return obj
        return cls._decompose_class(obj)
```

vitalik:
> I bumped the pydantic version to 1.9
well this cannot be merged - there ALOT projects that depends on it 
I would keep minimum 1.6 pydantic while it still not breakable

SmileyChris:
> well this cannot be merged - there ALOT projects that depends on it
> I would keep minimum 1.6 pydantic while it still not breakable

Sure, I backported that little bit of logic from Pydantic into the `Schema._decompose_class` method and reverted the changes. Then I squashed everything to keep things tidier :)

vitalik:
Hi @SmileyChris 

I have one concern about this PR

Basically it's the performance penalty here:
```Python
    def __getitem__(self, key: str) -> Any:
        resolve_func = getattr(self._cls, f"resolve_{key}", None) if self._cls else None
        if resolve_func and callable(resolve_func):
            item = resolve_func(self._obj)
        else:
            try:
                item = attrgetter(key)(self._obj)
            except AttributeError as e:
                raise KeyError(key) from e
        return self.format_result(item)
```

Every attribute will call for getattr, callable check, attrgetter, etc

but in 99% of the cases schemas are just generally pulling existing atrs...

maybe solution here would be first try to get attr default way and if attr is not there try to resolve it with attrgetter/resolve_func:


```Python
try:
    return getattr(self._obj)
except AttributeError:
    if resolve_func and callable(resolve_func):
           return resolve_func(self._obj) 
    else:
        return attrgetter(key)(self._obj)
```

SmileyChris:
At first glance, that looked like a fine solution. But it'll get in the way allowing for `resolve_<attr>` functions if it's shadowing an existing attribute.

Doing some basic `%timeit` tests, it looks like it only adds about 400ns to each attr call, so 0.0000004s. 

```
# Old class
%timeit old_django_getter['name']
455 ns ± 2.44 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

# New class
%timeit new_django_getter['name']
875 ns ± 15.7 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

SmileyChris:
Ok, now the only overhead for the 99% case is a getattr for a resolve attr first. Python logic means it won't check if it's callable unless the resolve attr is found.

vitalik:
@SmileyChris great, looks promising

I guess the only small design issue is this part:
```Python
    @staticmethod
    def resolve_owner(obj):
```
that you have to to mark method as static... I guess generally people will forget it and will use `self` as first argument (maybe need some validation on top to force it never forgotten)


or maybe just pass `self` somehow  anyway ?

```Python
def resolve_owner(self, obj):
```


also would be nice to have access to `self` dict to cover long awaited [feature #935](https://github.com/samuelcolvin/pydantic/issues/935) from pydantic: 

```Python
class Rectangle(BaseModel):
    width: int
    length: int
    area: int

    def resolve_area(self) -> int:
        return self.width * self.length
```

SmileyChris:
Technically it doesn't need to be a static method, that's just me being a
bit pedantic in the documentation. Since it's not being called as a bound
method on an instance of the class, the first argument will be passed
in, it just won't be an instance of the class itself, which is slightly
confusing...

But the method is being called from the class object, and expects one
parameter, so a normal method like `def resolve_area(self) -> int:` already
also works fine.

SmileyChris:
Ok, now resolvers can be standard methods, giving you access to `self` like you want (it has to be a bit magic, because the real schema instance is only created _after_ all the fields are resolved fully).

vitalik:
Hi @SmileyChris 

I've been running some test - so far looks good except the perfomance for `from_orm`

example - 50k executions for from_orm takes 10 seconds on my machine (while it's only 1 second before)

So I basically moved the function to find `resolve_*` attributes to one time call and cache on class level

see - https://github.com/vitalik/django-ninja/pull/339/commits/da28fa1fb65902c48ffdc0e750aa7754aa9b8d1c

It still in progress - there are few coverage missing lines, but so far looks promising

SmileyChris:
That's nice caching speedup. How about this metaclass approach to keep resolver data on their own classes rather than one mega dictionary?

vitalik:
@SmileyChris 
Thank you! Works perfect

PR #336, opened 2022-01-27
Provide a temporal HttpResponse typed argument to views

### LINKED ISSUE #106: Provide a way to specify response headers
Seems like currently [Operation._result_to_response](https://github.com/vitalik/django-ninja/blob/5a19d230cb451f1f3ab0e167460bc05f49a6554e/ninja/operation.py#L106-L135) do not expect response headers from view function result (if not already part of `HttpResponse` instance).

This makes passing response headers tedious task.

Maybe `django-ninja` should allow view handler to setup response headers,

1. Via [temporal Response instance](https://fastapi.tiangolo.com/advanced/response-headers/) as done in FastAPI 
2. Or via [last tuple item](https://flask.palletsprojects.com/en/1.1.x/quickstart/#about-responses) as done in Flask

What do you think?

### LINKED ISSUE #117: Cookies and responses
Just needed to set cookie when returning json from an api-function.

SmileyChris:
Provide a temporal response object to views allowing the altering of headers / cookies.
(needs docs, just pushing for initial review)

Fixes #106 and fixes #117

e.g.

```Python
@api.post("/boop")
def boop(request, response: HttpResponse):
    response.set_cookie("beep", "boop")
    return True
```


SmileyChris:
Have you had a chance to look at the idea here, @vitalik ? I can write some docs up if it's something you want to proceed with.

vitalik:
Hi @SmileyChris 
Yes, this will likely land in the next release

SmileyChris:
I'll write some docs up then. It'd be nice to avoid conflicts by getting my docs branch in first ;)

SmileyChris:
The only additional thing I've been thinking about in regards to this is having the status code set in the temporal response be used, rather than overwritten. We could deprecate the 2-part tuple behaviour of passing `(status_code, body)` and just use this new response object as a way of providing an alternative status code.

This doesn't have to happen as part of this initial pull, just wanted to bring this up as a related idea.

stephane (approved):
LGTM

Issue #345, opened 2022-01-31
Cache: Improvement proposal


mom1:
Hello.

It would be great to be able to use the cache out of the box in a similar way to working with authorization.

```python
    @router.get('/ping', response=..., cache=Cache(timeout=..., cache_key=..., cache_alias=...))
    def pong(request):
        ...
```

I have read the method from the task (#286) and there are several problems.

1. Doesn't work, if returned queryset from view
2. Double serialization of response

The proposed approach can help solve these problems.

PR #346, opened 2022-01-31
Support caching.

mom1:
Hello. 

This is an implementation of caching by suggestion #345.

I did it only for `get`, but I think it will not be difficult to expand.

I hope it will be useful :)


SmileyChris:
Why not just implement this as a decorator than an operator argument?

mom1:
A lot of response options from the view.  You need to understand what is expected from the view for serialization and caching.  An obvious example is when a queryset is returned from a view, it cannot be sent to the cache directly.

SmileyChris:
Reading your reasons on the other ticket. I'm not sure I agree that it means you couldn't have it as an argument still. Just make the decorator return a wrapping cache class, which then the operator function can check for and still use to determine the cache status.

I guess I'm just not a fan of adding more and more arguments to the operators...

mom1:
> Reading your reasons on the other ticket. I'm not sure I agree that it means you couldn't have it as a decorator still. Just make the decorator return a wrapping cache class, which then the operator function can check for and still use to determine the cache status.
> 
> I guess I'm just not a fan of adding more and more arguments to the operators...

I don't quite understand how it can be done differently.  Can you provide some kind of outline or meta code?

mom1:
@vitalik I fully agree with this. 
I tried to implement this, but I ran into a problem: for serialization, you need to know the response, which is in the Operation class, and the serialization logic too. When processing the request, the router does not participate and it is impossible to get the desired operation.

To avoid code duplication
```python
def cache_page(f):
    @wraps(f)
    def wrapper(request, *args, **kwargs):
        key  = `your key from request`
        result = django_cache.get(key)
        if result:
             return result
        result = f(request, *args, **kwargs)
        # here you need to magically serialize result using MyModel
        django_cache.set(key, result, timeout=30)
        return result

    return wrapper

@api.get('/foo', response={200: MyModel})
@cache_page(response={200: MyModel})
def foo(request, a: int, b: str):
     return Model.objects.all()
```
I came to the suggested solution

the idea of `decorate` will require more changes and I don't fully understand it.

mom1:
Hmm. I understand. `_ninja_contribute_to_operation`

vitalik:
@mom1 

yeah, I guess you can use `_ninja_contribute_to_operation` callback to change decorate `Operation.run`

just keep in mind that run can be both sync and async and it should be checked before decorating

mom1:
@vitalik You can look at my [commit](https://github.com/vitalik/django-ninja/pull/346/commits/e3d90d5c289e94e8cf0d1ff6adcbe8901a411dc5), would such a check be enough?

vitalik:
@mom1 
I think it will not work - basically to support async views you need to implement two wrappers (sync and async)

you need to add there tests as well (where you've put `# pragma: no cover`)

see how async tests are done here - https://github.com/vitalik/django-ninja/blob/master/tests/test_async.py

baseplate-admin:
Hi is this stale ?

Can we get some progress on this? I am in a situation where i need caching support badly

jaredleishman:
It looks like this is stale. There are alternatives to using caching as mentioned in the docs [here](https://django-ninja.dev/whatsnew_v1/?h=cache#decorators) and in Vitalik's comment [here](https://github.com/vitalik/django-ninja/issues/775#issuecomment-1646654821)

PR #351, opened 2022-02-08
[feat] add superuser session authentication only

areski:
I have some use cases where some APIs are only for some privileged root user and it quite helpful to have an helper to add this extra security.

Let me know

### DIFF
--- a/ninja/security/__init__.py
+++ b/ninja/security/__init__.py
@@ -1,6 +1,6 @@
 from ninja.security.apikey import APIKeyCookie, APIKeyHeader, APIKeyQuery
 from ninja.security.http import HttpBasicAuth, HttpBearer
-from ninja.security.session import SessionAuth
+from ninja.security.session import SessionAuth, SessionAuthSuperUser
 
 __all__ = [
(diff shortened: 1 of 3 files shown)


areski:
There is still an error with codestyle: https://github.com/areski/django-ninja/actions/runs/1813501165

Related issue: https://github.com/typeddjango/django-stubs/issues/402
* https://github.com/typeddjango/django-stubs#how-can-i-create-a-httprequest-thats-guaranteed-to-have-an-authenticated-user

Any hints how to fix this, I'm not super familiar with mypy?

SmileyChris:
Well, mypy is correct -- `AbstractBaseUser` doesn't provide an `is_superuser` attribute. You're making the assumption that all user backends need to provide is_superuser (inferred from the base class, they don't).

I'd just do:
```python
if request.user.is_authenticated and getattr(request.user, 'is_superuser', None):
```

areski:
Thanks for the pointer @SmileyChris 

All tests are now passing: https://github.com/areski/django-ninja/actions/runs/1847443859

PR #361, opened 2022-02-14
Implement implicit reverse url name generation

### LINKED ISSUE #355: [BUG] Reverse url names are not auto generated
**Describe the bug**
Reverse resolution of urls described on https://django-ninja.rest-framework.com/tutorial/urls/ does not work for me. By inspecting the generated resolver, I discovered, that views that do not explicitly specify `url_name` do not have a name generated at all. View function name is not used.

 - Python version: 3.8
 - Django version: 3.2
 - Django-Ninja version: 0.17.0

SmileyChris:
Fixes #355

Huh, I had assumed this is how it worked anyway.

### DIFF
--- a/docs/docs/tutorial/urls.md
+++ b/docs/docs/tutorial/urls.md
@@ -18,6 +18,8 @@ def index(request):
 index_url = reverse_lazy("api-1.0.0:index")
 ```
 
+This implicit URL name will only be set for the first operation for each API path.
+
 ### Changing the URL name
 
 Rather than using the default URL name, you can specficy it explicitly as a property on the method decorator.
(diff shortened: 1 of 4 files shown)


vitalik:
Hi @SmileyChris 

Just wondering what would be the best approach when url is the same for get/post methods 

```Python

@api.post('/tasks')
def create_task(request):
     pass

@api.get('/tasks')
def list_tasks(request):
     pass

```

As I understand the last added name ("list_tasks") will be the name for that path (while some users might expect it to be create_task)

also there might be conflicts - if users have the same function name (like `list_items`) in multiple modules/routers...

maybe the best approach is to generarate url name (and open api operation_id) based on tags and function name

f.e. 
```
@router.get('/')
def list_tasks(request):
    return ...


....

api.add_router('/tasks/', router, tags=['tasks'])
```

the resulting url name should be `api-1.0.0:tasks:list_tasks`  (the "`tasks`" taken from the first tag assigned to router)

PR #369, opened 2022-02-20
Allow path parameters to be specified at router level

kaschnit:
This PR proposes and implements the following.

## Current behavior

Adding path params when including the router (e.g. `app.add_router("/item/{item_id}", router)` does not work as expected. The test client and the actual application routing logic cannot find the endpoint.

## New (proposed) behavior

Using `app.add_router("/item/{item_id}", router)` will allow the user to treat `item_id` as a path parameter if they so choose.

This change should not break any existing behavior.

## Example

With FastAPI, you can do something like this

```python
api = FastAPI()
router = Router(prefix="item/{item_id}/")

@router.get("/metadata")
def get_item_metadata(item_id: str):
  return item_id

api.include_router(router)
```

Then you could do the following and expect output:

```shell
# command
curl localhost:8000/item/my_item/metadata

# output
my_item
```

This change will allow a similar pattern in django-ninja path parameters when including the router in a parent router, e.g.:

```python
app = Ninja()
router = Router()

@router.get("/metadata")
def get_item_metadata(request, item_id: str = Path(None)):
  return item_id

app.add_router("/item/{item_id}/", router)
```

This PR only allows this to work when using the `Path()` object to indicate that it's a path parameter. The change to make it auto detect these is less trivial because the router's "prefix" is not actually a property of the `Router,` but a property of its parent router, so reading the prefix path is difficult when resolving path parameters.

### DIFF
--- a/ninja/main.py
+++ b/ninja/main.py
@@ -324,8 +324,7 @@ class NinjaAPI:
         result = get_openapi_urls(self)
 
         for prefix, router in self._routers:
-            for path in router.urls_paths(prefix):
-                result.append(path)
+            result.extend(router.urls_paths(prefix))
 
         result.append(get_root_url(self))
         return result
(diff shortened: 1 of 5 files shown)


vitalik:
thank you!

PR #397, opened 2022-03-18
Fix Django-Ninja PUT/PATCH request with UploadedFile

deniswvieira:
Request PUT or PATCH raises an error for endpoints that receives UploadedFile. This happens because:
1. Since we receive a File within the request, the content-type can not be application/json anymore, forcing to be multipart/form-data (and the `ParamModel` to be `_MultiPartBodyModel`);
2. Django does not handle the request like he does for POST requests, not loading the multipart/form-data from request body into request.POST and request.FILES. 

This crashes in Django Ninja while processing the request (`Operation.run()`) for `ParamModel` `_MultiPartBodyModel` and `FileModel` since it tries to retrieve data from request.POST and request.FILES.

Possible solutions I reached:
1. Add some logic into `_MultiPartBodyModel`  and `FileModel` to load the request body data like the Django does for POST request (essencially, replicate Django logic)
2. Ask "nicely" to Django load the data for us, making he think he is loading a POST request.

I've choose the second approach.

### DIFF
--- a/ninja/operation.py
+++ b/ninja/operation.py
@@ -31,6 +31,27 @@ if TYPE_CHECKING:
 __all__ = ["Operation", "PathView", "ResponseObject"]
 
 
+def load_request(request: HttpRequest):
+    """
+    Loads a non-json PUT/PATCH request as a POST, reading data from request body
+    and placing them in request.POST and request.FILES.
+    This is required to fix the behaviour when we have a PUT/PATCH endpoint that also
+    receives a File as parameter, forcing the data to be sent as multipart/form-data and
+    making the ParamModel look up at request.POST and request.FILES
+    """
+    if request.method in ("PUT", "PATCH") and request.content_type != "application/json":
+        if hasattr(request, '_post'):
+            del request._post
+            del request._files
+        try:
+            request.method = "POST"
+            request.META['REQUEST_METHOD'] = 'POST'
+            request._load_post_and_files()
+            request.META['REQUEST_METHOD'] = 'PUT'
+            request.method = "PUT"
+        except Exception:
+            pass
+
 class Operation:
     def __init__(
         self,
@@ -90,6 +111,7 @@ class Operation:
             view_func._ninja_contribute_to_operation(self)  # type: ignore
 
     def run(self, request: HttpRequest, **kw: Any) -> HttpResponseBase:
+        load_request(request)
         error = self._run_checks(request)
         if error:
             return error
@@ -240,6 +262,7 @@ class AsyncOperation(Operation):
         self.is_async = True
 
     async def run(self, request: HttpRequest, **kw: Any) -> HttpResponseBase:  # type: ignore
+        load_request(request)
         error = self._run_checks(request)
         if error:
             return error
--- a/ninja/operation.py
+++ b/ninja/operation.py
@@ -40,6 +40,7 @@ def load_request(request: HttpRequest):
     making the ParamModel look up at request.POST and request.FILES
     """
     if request.method in ("PUT", "PATCH") and request.content_type != "application/json":
+        original_method = request.method
         if hasattr(request, '_post'):
             del request._post
             del request._files
@@ -47,8 +48,8 @@ def load_request(request: HttpRequest):
             request.method = "POST"
             request.META['REQUEST_METHOD'] = 'POST'
             request._load_post_and_files()
-            request.META['REQUEST_METHOD'] = 'PUT'
-            request.method = "PUT"
+            request.META['REQUEST_METHOD'] = original_method
+            request.method = original_method
         except Exception:
             pass
 


guicarvalho:
@vitalik What do you think of this proposal?

ninja/operation.py
```diff
@@ -31,6 +31,27 @@
...
+    if request.method in ("PUT", "PATCH") and request.content_type != "application/json":
+        if hasattr(request, '_post'):
+            del request._post
+            del request._files
+        try:
+            request.method = "POST"
+            request.META['REQUEST_METHOD'] = 'POST'
+            request._load_post_and_files()
+            request.META['REQUEST_METHOD'] = 'PUT'
+            request.method = "PUT"
```
OtherBarry, line 51: Probably best not to convert a PATCH into a PUT needlessly.

```suggestion
    if request.method in ("PUT", "PATCH") and request.content_type != "application/json":
        original_method = request.method
        if hasattr(request, '_post'):
            del request._post
            del request._files
        try:
            request.method = "POST"
            request.META['REQUEST_METHOD'] = 'POST'
            request._load_post_and_files()
            request.META['REQUEST_METHOD'] = original_method
            request.method = original_method
```

danihuge: I got same issue today using form data and PUT method... :(

baseplate-admin:
Hi @vitalik, 

Do you think we can merge this?

This is a requirement for some users ( eg : https://github.com/vitalik/django-ninja/discussions/613?notification_referrer_id=NT_kwDOA69C67M0ODgyNDE5MDc4OjYxODE3NTc5#discussioncomment-4269198 )

OtherBarry:
@deniswvieira you'll need to write some tests and (likely) do some codestyle fixes before this is able to be merged. I'm happy to help out if needed.

baseplate-admin:
@vitalik,  this pull request is better than middleware approach. Please merge this. ( i can port my tests if you want )

ninja/operation.py
```diff
@@ -31,6 +31,28 @@
 __all__ = ["Operation", "PathView", "ResponseObject"]
 
 
+def load_request(request: HttpRequest):
```
baseplate-admin, line 34: def load_request(request: HttpRequest) -> None:

ninja/operation.py
```diff
@@ -31,6 +31,28 @@
...
+            del request._post
+            del request._files
+        try:
+            request.method = "POST"
+            request.META['REQUEST_METHOD'] = 'POST'
+            request._load_post_and_files()
+            request.META['REQUEST_METHOD'] = original_method
+            request.method = original_method
+        except Exception:
+            pass
```
baseplate-admin, line 54: If we do this.. why not use `contextlib.supress` ?

baseplate-admin (changes requested):

baseplate-admin:
@vitalik do you want anything added to this PR?

ninja/operation.py
```diff
@@ -31,6 +31,28 @@
...
+    """
+    if request.method in ("PUT", "PATCH") and request.content_type != "application/json":
+        original_method = request.method
+        if hasattr(request, '_post'):
+            del request._post
+            del request._files
+        try:
+            request.method = "POST"
+            request.META['REQUEST_METHOD'] = 'POST'
+            request._load_post_and_files()
```
SmileyChris, line 50: I think this should just hit `request.FILES` rather than manually accessing an undocumented underscored method?
```suggestion
            request.FILES
```

ninja/operation.py
SmileyChris: This is more than just a style fix too. `request.META['REQUEST_METHOD']` and `request.method` should be returned to their original method even if the file loading method fails (silently).

ninja/operation.py
baseplate-admin: Correct me if i am wrong but this is probably due to how django handles `FILES`.

Quoting @deniswvieira, The alternative is to,

> Add some logic into _MultiPartBodyModel and FileModel to load the request body data like the Django does for POST request (essencially, replicate Django logic)

Which in turn adds churn

ninja/operation.py
baseplate-admin:
```python
with contextlib.supress(Exception):
    request.method = "POST"
    request.META['REQUEST_METHOD'] = 'POST'
    request._load_post_and_files()

request.META['REQUEST_METHOD'] = original_method
request.method = original_method
```

I was thinking something like this

ninja/operation.py
SmileyChris: @baseplate-admin `request.FILES` is a property of WSGIRequest in django that triggers _load_post_and_files if `._files` doesn't exist. Which it doesn't, because it was deleted a few lines earlier in this PR ;)
https://github.com/django/django/blob/main/django/core/handlers/wsgi.py#L104-L108

ninja/operation.py
SmileyChris: Yes, your code suggestion is what I had in mind.
I'm not quite sure if it can actually fail, and maybe we shouldn't be catching and silently swallowing an exception anyway... Yeah, this just seems like a bad idea altogether unless @deniswvieira can say why we need to catch it (and with such a broad catch at that)

baseplate-admin:
@vitalik, What's your take on this PR?

If i rebase this PR with some tests would you be willing to merge this?

bigusef:
Hi @vitalik
First. thanks for creating this project! I recently merging my project from DRF to ninja. hope to you all the best.
regarding to this issue I facing it in my project and there are many issues about it. what are the plan here to fix this issue?

yokotoka:
Hello! When you accept that patch? That problem with post/patch is annoying (

pmdevita:
It looks like the user who originally wrote this PR is inactive now and has been for two years. I'd be willing to pick this up and add the suggested changes along with tests if that is alright. @vitalik do you have any input for this?

baseplate-admin:
Due to in-activeness of this pr, i have decided to create [`ninja_put_patch_file_upload_middleware`](https://github.com/baseplate-admin/ninja_put_patch_file_upload_middleware)

jleclanche:
@vitalik Any chance you can take a look at this? It's blocking #1059.

adambirds:
Would be good to know what is blocking this still @vitalik ?

demiurg:
why is this blocked? I spent a few days on this

baseplate-admin:
@vitalik, please close this issue

vitalik:
Thank your @deniswvieira 

but I think #1423 this makes it the best apprach so far:

 - no performance penalties if user do not use files with patch/put
 - error notice if user does use it and forgot to include "fix" middleware

JoaoPedroMDP:
+1. But ok, i will download the library.

PR #400, opened 2022-03-21
UploadedFile inherit from Django's UploadedFile

OtherBarry:
Having uploaded file inherit from bytes causes inaccurate typing.

This PR fixes the typing issues by:
- having UploadedFile inherit from Django's UploadedFile instead of bytes
- adding a __modify_schema__ class method so that it appears the same way as a `bytes` type would in schemas.

### DIFF
--- a/ninja/files.py
+++ b/ninja/files.py
@@ -18,5 +18,5 @@ class UploadedFile(DjangoUploadedFile):
         return v
 
     @classmethod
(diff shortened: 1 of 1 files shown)


OtherBarry:
Missing type hint fixed.

Running mypy was not in `CONTRIBUTING.md`, so I'll add that in a separate PR.

PR #402, opened 2022-03-24
Access to test response attributes

stephane:
Give an access to the attributes of the Django response wrapped inside the test response of Ninja by providing a `__getattr__` method that it called when an attribute lookup has not found the attribute of the test response.


stephane:
It was ⚡ 
Thank you.

PR #407, opened 2022-03-28
Allow return of status code only with many responses

stephane:
It's possible to return a single int value as status code in the request when the `response` variable define several responses but not when only one response is defined.

This PR fixes this difference of behavior.

### DIFF
--- a/ninja/operation.py
+++ b/ninja/operation.py
@@ -162,8 +162,11 @@ class Operation:
         if isinstance(result, HttpResponseBase):
             return result
 
-        status: int = 200
+        status: int = result if isinstance(result, int) else 200
+
         if len(self.response_models) == 1:
+            # Fast way to extract the first key of the response models dict,
+            # tuple or list to get the status.
             status = next(iter(self.response_models))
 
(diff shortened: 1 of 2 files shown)


stephane:
I don't know why `mypy` check failed, I didn't touch the concerned file and it doesn't fail on my env:

```
❯ mypy ninja
Success: no issues found in 44 source files
```

SmileyChris:
So what's the inconsistent behaviour that works currently? According to the docstring (and my interpretation of the code), a single int should _always_ be just the body:

> The protocol for results
>  - if HttpResponse - returns as is
>  - if tuple with 2 elements - means http_code + body
>  - otherwise it's a body

vitalik:
@stephane 
I think this overcomplicates the logic 

and I think it will also break when you want to return just integer number

```Python
@api.get('/hello', result=int)
def number(request):
    return 42
```

stephane:
You're right!
I misinterpreted the API by using the return of the view as a status code instead of body (I was thinking scalar value should be interpreted as status code).

A unit test with a unique int in body could help me to detect my mistake (PR to follow...) ;)

PR #410, opened 2022-03-29
Fix TypeError in is_collection_type

baijifeilong:
`is_collection_type(typing.Any)` will throw exception: **TypeError: issubclass() arg 1 must be a class**


flaeppe:
Changes here are probably superseded by #434

PR #414, opened 2022-04-03
Provide a way to inject pagination to multiple response schema.

hslim8888:
Hi @vitalik 
I found some issues in pagination feature when using multiple response schemas.

### Motivation
- The type of multiple response schemas is different with the single schema. 
    - Input of response schema
        - single : type == Schema or List of Schema
       `@api.get("/tasks", response=List[TaskSchema])`
        - multiple : type == dict 
       `@api.post('/login', response={200: Token, 401: Message, 402: Message})`
    
    - Output of response schema
        - single : type == queryset or list
        `return Task.objects.all()`
        - multiple : type == tuple
        `return 200, Task.objects.all()`

- The difference in type of Input reponse occurs an error when using a builtin router class(`RouterPaginated`)
- The difference in type of Output reponse occurs an error when injecting pagination to result(response).
     
### Modifications & Checkpoints
- Add method check & response type check in `RouterPaginated` class.
    - I assume only GET method needs to be injected to pagination.
    - Also, status code 200 should be required if someone want to use multiple response schemas.
- Add result(response) type check in `_inject_pagination` function.
    - `paginator.paginate_queryset` can not handle items which type is tuple.
    - No need to force queryset evaluation to result['items'] since `from_orm` will take care of the rest of the process.

I must say django ninja is amazing. It makes my coding experience happy. 
Thank you very much.

### DIFF
--- a/ninja/pagination.py
+++ b/ninja/pagination.py
@@ -145,9 +145,7 @@ def _inject_pagination(
         result = paginator.paginate_queryset(
             items, pagination=pagination_params, **kwargs
         )
-        if paginator.Output:
-            result["items"] = list(result["items"])
-        # ^ forcing queryset evaluation #TODO: check why pydantic did not do it here
+
         return result
 
     view_with_pagination._ninja_contribute_args = [  # type: ignore
--- a/ninja/pagination.py
+++ b/ninja/pagination.py
@@ -139,6 +139,8 @@ def _inject_pagination(
             kwargs[paginator.pass_parameter] = pagination_params
 
         items = func(*args, **kwargs)
+        if isinstance(items, tuple) and len(items) == 2:
+            items = items[1]
 
         result = paginator.paginate_queryset(
             items, pagination=pagination_params, **kwargs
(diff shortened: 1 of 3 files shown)


vitalik:
Hi @hslim8888 
Thank you for PR
Would you be able to make a test case that covers this functionality ?

ninja/pagination.py
```diff
@@ -176,9 +176,13 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
     def add_api_operation(
         self, path: str, methods: List[str], view_func: Callable, **kwargs: Any
     ) -> None:
```
vitalik, line 178: I'm not sure about this one - only GET method
Technically you can return paginated response on any method GET/POST/etc...

hslim8888: I just assume people usually use pagination on GET method and thought it's more efficient not to inject other methods to pagination.
But you're right. No need to limit possibilities.
I will rollback that thing.

vitalik:
this one gets conflicts, and I'm not sure yet for the status codes...

PR #437, opened 2022-05-04
Make `NinjaAPI.add_router` idempotent

### LINKED ISSUE #436: Make `NinjaAPI().add_router` idempotent
**Problem**

I have a project where we attach/register routes/urls during `apps.ready`. Which becomes problematic to keep doing when using `django-ninja` while also having test cases calling

```python
with django.test.utils.modify_settings(INSTALLED_APPS={"append": ["django-app"]}):
    ...
```

As that'll trigger a runthrough of `apps.ready` for each installed app and `NinjaAPI().add_router` isn't idempotent, but instead raises a `ConfigError` from the code below:

https://github.com/vitalik/django-ninja/blob/22e97cdab9faabc84a048eeac688192f3f1f19d7/ninja/router.py#L352-L359

I'd like `NinjaAPI().add_router` to be idempotent. Which I, from my brief investigation, think is possible. Since a `NinjaAPI` instance stores a list of 2-tuples with each tuple containing a prefix and a router. So what if `NinjaAPI.add_router` starts out by checking

```python
if router.api == self and (prefix, router) in self._routers:
    # Idempotent on identical registration
    return
elif router.api is not None:
    # Router already attached to an api
    raise ConfigError(...)
```

I think that should also allow removal of the `debug_server_url_reimport` in `Router.build_routers`:

https://github.com/vitalik/django-ninja/blob/22e97cdab9faabc84a048eeac688192f3f1f19d7/ninja/router.py#L354-L356

flaeppe:
- Makes attachment of identical routes a noop call
- Fixes #436


flaeppe:
Since Django versions < 3.2 are not officially supported by Django any longer, perhaps we should drop all of those? (https://www.djangoproject.com/download/)

flaeppe:
@vitalik is this something you would consider merging?

linspw:
Do you have any predictions on when this will be merged?
@vitalik

vitalik:
> Do you have any predictions on when this will be merged? @vitalik

well I'm still thinking about it - as some times I see people have this error, but then endsup it's because some internal import loop... trying to collect more stats and feedbacks on this one

karel-caplena:
+1 on merging this one, would help us big time!

is there a downside? if the fear is silently adding something twice, would a key in the settings do it? i.e. if one specifies `idempotent=True`, then it would let add duplicates and not complain

vitalik:
Hi @flaeppe 

I'm still considering adding this... but too much conflict at the momment

would you be able to create a new PR from v1

flaeppe:
Hi @vitalik I've branched out from this in my fork and rebased it a little while ago. I'll try to rebase it again and push it here. I think I resolved most conflicts the last time.

flaeppe:
@vitalik I've rebased and resolved conflicts just now. Think the mypy error is related to the newly released 1.7.0 incompatibility with pydantic or something like that.

Object905:
This also seems to fix issue with django 5.1.
runserver errors due to this, unless --skip-checks is specified

hassaanalansary:
any news when will this get merged, it looks very helpful

Beki95:
when will it be added?

anryangelov:
I am very much waiting too

vitalik:
done in 1.6

PR #464, opened 2022-06-07
Allow arbitrary Pagination Output

rafonseca:
Currently, the schema of a paginator output MUST contain the field "items".
The scope of this PR is to eliminate this restriction.

I ran into this problem while trying to replicate a DRF [paginator](https://www.django-rest-framework.org/api-guide/pagination/#limitoffsetpagination). There we need a field named "results". Besides that, we need access to the request object during pagination to compose the fields "next" and "previous". I would like to add this functionality as well in the next PR. 
Please let me know if it is more appropriate to include it in this PR instead.

BTW, thanks @vitalik . This project rocks.

### DIFF
--- a/ninja/pagination.py
+++ b/ninja/pagination.py
@@ -208,10 +208,7 @@ def make_response_paginated(paginator: PaginationBase, op: Operation) -> None:
     new_schema = type(
         new_name,
         (paginator.Output,),
-        {
-            "__annotations__": {"items": List[item_schema]},  # type: ignore
-            "items": [],
-        },
+        {},
     )  # typing: ignore
 
     response = op._create_response_model(new_schema)
--- a/ninja/pagination.py
+++ b/ninja/pagination.py
@@ -63,8 +63,11 @@ class LimitOffsetPagination(PaginationBase):
     ) -> Any:
         offset = pagination.offset
         limit: int = pagination.limit
+        _items = queryset[offset : offset + limit]
+        items = list(_items)
+        # ^ forcing queryset evaluation #TODO: check why pydantic did not do it here
         return {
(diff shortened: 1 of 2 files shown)


vitalik:
Well, it does not really work for arbitrary responses:


```Python
api = NinjaAPI()


class PictureSchema(Schema):
    class Meta:
        model = Picture
        fields = ("id", "title", "image")


class ResultsPaginator(PaginationBase):
    class Input(Schema):
        skip: int

    class Output(Schema):
        results: list
        count: int
        skip: int

    def paginate_queryset(self, items, pagination: Input, **params):
        skip = pagination.skip
        return {
            "results": list(items[skip : skip + 5]),
            "count": self._items_count(items),
            "skip": skip,
        }


@api.get("/pics", response=list[PictureSchema])
@paginate(ResultsPaginator)
def pics(request):
    return Picture.objects.all()
```
throws a 

```
  File "json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type Picture is not JSON serializable

```

this is actually an important part:

CleanShot 2022-06-25 at 17 15 13@2x


that let's pydantic know full schema and convert it during evaluation

vitalik:
@rafonseca 
I think this one should do the expected

https://github.com/vitalik/django-ninja/pull/483/files#diff-feb8cae3c400ac0b90e421ea784871ac389894df1e7db50cfa253ccb8ee89ff2R58

rafonseca:
Hi @vitalik . 
Indeed, my solution was working for me only because I was returning list of pydantic objects. However, we do want to deal with a queryset here. Sorry for the mistake.

PR #485, opened 2022-06-27
Add exception `MethodNotAllowedError` and handler for it

redlovv:
Instead of a direct answer from the b"Method not allowed", I propose to process the resulting exception, like other exceptions.
Added a `MethodNotAllowedError` error, as well as a handler for this error.

### DIFF
--- a/ninja/errors.py
+++ b/ninja/errors.py
@@ -1,7 +1,7 @@
 import logging
 import traceback
 from functools import partial
-from typing import TYPE_CHECKING, List
+from typing import TYPE_CHECKING, List, Set
 
 from django.conf import settings
 from django.http import Http404, HttpRequest, HttpResponse
@@ -49,6 +49,12 @@ class HttpError(Exception):
         super().__init__(message)
 
 
+class MethodNotAllowed(Exception):
+    def __init__(self, allowed_methods: Set[str]) -> None:
+        self.allowed_methods = allowed_methods
+        super().__init__()
+
+
 def set_default_exc_handlers(api: "NinjaAPI") -> None:
     api.add_exception_handler(
         Exception,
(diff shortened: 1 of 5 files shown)


ninja/errors.py
```diff
@@ -43,6 +43,12 @@ def __init__(self, status_code: int, message: str) -> None:
         super().__init__(message)
 
 
+class MethodNotAllowed(Exception):
+    def __init__(self, allowed_methods: Set[str]) -> None:
+        self.allowed_methods = allowed_methods
+        super().__init__()
```
carlosdorneles-mb, line 49: Hello, I noticed that this new class was not declared in the `__all__` of this file. It must be declared so that it is available to be imported into other modules.

ninja/errors.py
```diff
@@ -43,6 +43,12 @@ def __init__(self, status_code: int, message: str) -> None:
         super().__init__(message)
 
 
+class MethodNotAllowed(Exception):
```
duducp, line 46: Hello, I noticed that this new class was not declared in the `__all__` of this file. It must be declared so that it is available to be imported into other modules.

redlovv: Hello! Thanks for the review. I fixed it.

ninja/errors.py
```diff
@@ -66,10 +73,18 @@ def set_default_exc_handlers(api: "NinjaAPI") -> None:
...
+        partial(_default_method_not_allowed, api=api),
+    )
     api.add_exception_handler(
         AuthenticationError,
         partial(_default_authentication_error, api=api),
     )
+    api.add_exception_handler(
+        MethodNotAllowedError,
+        partial(_default_method_not_allowed, api=api),
+    )
```
carlosdorneles-mb, line 87: After your last push the `add_exception_handler` for a `MethodNotAllowedError` exception got duplicated.

duducp:
It would be nice in your pull request to have a description of the changes that have taken place.

ninja/errors.py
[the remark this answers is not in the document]
redlovv: Yes, sorry, removed it. Also added a description to the PR.

dan83g:
This feature is required. I am waiting too.

PR #486, opened 2022-06-27
Also pass the router to `get_operation_url_name`

SmileyChris:
Seems like having access to the router could also be useful when you fall back to configuring the url_name (building the url name by tags, or whatever)

### DIFF
--- a/ninja/router.py
+++ b/ninja/router.py
@@ -314,10 +314,11 @@ class Router:
             route = normalize_path(route)
             route = route.lstrip("/")
 
-            if not path_view.url_name:
-                url_name = self.api.get_operation_url_name(path_view.operations[0])  # type: ignore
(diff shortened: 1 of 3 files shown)


ninja/router.py
```diff
@@ -314,10 +314,11 @@ def urls_paths(self, prefix: str) -> Iterator[URLPattern]:
             route = normalize_path(route)
             route = route.lstrip("/")
 
-            if not path_view.url_name:
-                url_name = self.api.get_operation_url_name(path_view.operations[0])  # type: ignore
-            else:
-                url_name = path_view.url_name
+            url_name = path_view.url_name or ""
+            if not url_name and self.api:
```
SmileyChris, line 318: Note: this is also a small logical improvement.

PR #530, opened 2022-08-16
Attribute `description` declared has type `str` but is used as type None

luca-digrazia:
**"filename"**: "docs/src/tutorial/body/code01.py"
**"warning_type"**: "Incompatible attribute type [8]"
**"warning_message"**: " Attribute `description` declared in class `Item` has type `str` but is used as type `None`.",
**"warning_line"**: 6
**"fix"**: "" instead of None

### DIFF
--- a/docs/src/tutorial/body/code01.py
+++ b/docs/src/tutorial/body/code01.py
@@ -3,7 +3,7 @@ from ninja import Schema
 
 class Item(Schema):
     name: str
-    description: str = None
+    description: str = ""
     price: float
     quantity: int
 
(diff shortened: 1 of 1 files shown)


luca-digrazia:
Up :)

vitalik:
@luca-digrazia thank you for PR - but tests are failing...
https://github.com/vitalik/django-ninja/runs/7858252406?check_suite_focus=true

and if you read the docs for that page - there is also a reference that `None` is expected there


so what you probably mean is to mark it as Optional:


```Python
description: Optional[str] = None
```

luca-digrazia:
> @luca-digrazia thank you for PR - but tests are failing... https://github.com/vitalik/django-ninja/runs/7858252406?check_suite_focus=true
> 
> and if you read the docs for that page - there is also a reference that `None` is expected there
> 
> so what you probably mean is to mark it as Optional:
> 
> ```python
> description: Optional[str] = None
> ```

Yes, updated with optional.

PR #539, opened 2022-08-26
Add YAML output format option

denizdogan:
Let me know what you think.

### DIFF
--- a/ninja/management/commands/export_openapi_schema.py
+++ b/ninja/management/commands/export_openapi_schema.py
@@ -1,11 +1,6 @@
 import json
 from typing import Any, Optional
 
-try:
-    from django.core.serializers.pyyaml import yaml, DjangoSafeDumper
-except ImportError:
-    yaml = None
-
 from django.core.management.base import BaseCommand, CommandError, CommandParser
 from django.urls.base import resolve
 from django.utils.module_loading import import_string
@@ -80,36 +75,38 @@ class Command(BaseCommand):
             dest="sort_keys",
             default=False,
             action="store_true",
-            help="Sort Json keys",
+            help="Sort schema keys",
         )
 
     def handle(self, *args: Any, **options: Any) -> None:
         api = self._get_api_instance(options["api"])
         schema = api.get_openapi_schema()
-        result = json.dumps(
-            schema,
-            cls=NinjaJSONEncoder,
-            indent=options["indent"],
-            sort_keys=options["sort_keys"],
-        )
 
-        if options["format"] == "yaml":
-            data = json.loads(result)
-            result = yaml.dump(data, indent=options["indent"])
+        fmt = options["format"]
+        if fmt == "json":
+            result = json.dumps(
+                schema,
+                cls=NinjaJSONEncoder,
+                indent=options["indent"],
+                sort_keys=options["sort_keys"],
+            )
+        elif fmt == "yaml":
+            try:
+                import yaml
 
-        format = options["format"]
-        if format == "json":
-            result = json.dumps(schema, cls=NinjaJSONEncoder, indent=options["indent"])
-        elif format == "yaml":
-            if not yaml:
-                self.stderr.write("Install PyYAML or django-ninja[yaml].")
-                self.exitcode = 1
-                return
-            result = yaml.dump(dict(schema), Dumper=DjangoSafeDumper)
+                from ninja.yaml import NinjaSafeDumper
+            except ImportError:
+                raise CommandError("Install PyYAML or django-ninja[yaml].")
+            result = yaml.dump(
+                schema,
+                Dumper=NinjaSafeDumper,
+                default_flow_style=False,
+                indent=options["indent"],
+                sort_keys=options["sort_keys"],
+            )
         else:
-            self.stderr.write(f"Unknown format: {format}")
-            self.exitcode = 2
-            return
+            # this shouldn't happen, argparse would already complain
+            raise CommandError(f"Unknown schema format: '{fmt}'")
 
         if options["output"]:
             with open(options["output"], "wb") as f:
--- a/ninja/management/commands/export_openapi_schema.py
+++ b/ninja/management/commands/export_openapi_schema.py
@@ -100,7 +100,6 @@ class Command(BaseCommand):
             result = yaml.dump(
                 schema,
                 Dumper=NinjaSafeDumper,
-                default_flow_style=False,
                 indent=options["indent"],
                 sort_keys=options["sort_keys"],
             )
--- a/ninja/management/commands/export_openapi_schema.py
+++ b/ninja/management/commands/export_openapi_schema.py
@@ -1,6 +1,7 @@
 import json
 from typing import Any, Optional
 
+import yaml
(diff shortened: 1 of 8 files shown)


denizdogan:
@vitalik Could I bother you by asking for a review of this? :)

baseplate-admin:
Hey there,

I dont think having `PyYaml` as a hard dependency is good for this project. ( because `django-ninja` should be minimal by default  ) 

( maybe add yaml as an extra dependency? )

ninja/management/commands/export_openapi_schema.py
```diff
@@ -55,14 +65,18 @@ def add_arguments(self, parser: CommandParser) -> None:
...
         )
 
     def handle(self, *args: Any, **options: Any) -> None:
         api = self._get_api_instance(options["api"])
         schema = api.get_openapi_schema()
         result = json.dumps(schema, cls=NinjaJSONEncoder, indent=options["indent"])
 
+        if options["format"] == "yaml":
+            data = json.loads(result)
+            result = yaml.dump(data, indent=options["indent"])
```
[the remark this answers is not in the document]
denizdogan: @baseplate-admin Good question! I don't remember the exact details, but it has to do with the order of the YAML output. When loaded via `json` first, the order of the underlying data structure is preserved somehow, so that when PyYAML does its thing, the order of the YAML looks better. Obviously, the order technically doesn't matter, but I personally think it's worth it. Do you think I should add a comment about it?

denizdogan:
I've made some changes to this, making the PyYAML dependency optional/extra, but I haven't been able to test it yet due to some issues related to my Python environment setup. As soon as I've fixed that, I'll update the PR.

ninja/management/commands/export_openapi_schema.py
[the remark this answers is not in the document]
baseplate-admin: If you need sorting [maybe sort the dictionary](https://realpython.com/sort-python-dictionary/)?

Doing JSON → YAML takes a lot of unnecessary processing.

:)

baseplate-admin:
I also have another suggestion. In the case `PyYaml` is not installed and user tries to dump. Show a message telling the user to install `PyYaml`

denizdogan:
@baseplate-admin OK, the changes that I had planned make it so that the `--format` argument only exists if `yaml` is importable, otherwise it's just silently always JSON. But I can follow your suggestion, no problem 👍

baseplate-admin:
In my opinion it improves DX a whole lot to show the developer why things are not working instead of silently failing :)

We might get a ton of PR on why YAML dump doesn’t work in the future if we take the silent fail route.

Just my 0.02$

baseplate-admin:
Try modifying the code im [here](https://github.com/vitalik/django-ninja/blob/3e9c10d5c8915a2ef76b08de8cdff98b45ca55f8/.github/workflows/test_full.yml#L22)

denizdogan:
@baseplate-admin Thanks, I will check it out tomorrow. Getting late now :) I have some more work to do `ninja.yaml.NinjaSafeDumper` still.

denizdogan:
@baseplate-admin Alright, so I think it's ready for a new review. I had to stop myself from doing too much with the unit tests. I think the test suites could really benefit from using [Hypothesis](https://hypothesis.readthedocs.io/en/latest/) and [Faker](https://faker.readthedocs.io/en/master/).

Anyways, let me know what you think!

ninja/management/commands/export_openapi_schema.py
```diff
@@ -55,25 +64,48 @@ def add_arguments(self, parser: CommandParser) -> None:
...
+                schema,
+                cls=NinjaJSONEncoder,
+                indent=options["indent"],
+                sort_keys=options["sort_keys"],
+            )
+        elif fmt == "yaml":
+            try:
+                import yaml
+
+                from ninja.yaml import NinjaSafeDumper
```
baseplate-admin, line 97: We dont need this

The import yaml will raise the error.

ninja/management/commands/export_openapi_schema.py
```diff
@@ -55,25 +64,48 @@ def add_arguments(self, parser: CommandParser) -> None:
...
+            except ImportError:
+                raise CommandError("Install PyYAML or django-ninja[yaml].")
+            result = yaml.dump(
+                schema,
+                Dumper=NinjaSafeDumper,
+                indent=options["indent"],
+                sort_keys=options["sort_keys"],
+            )
+        else:
```
baseplate-admin, line 106: Do we need this else statement if argparse already raises an error?🤔

baseplate-admin (commented):
I have some reviews. Please check :)

ninja/yaml.py
```diff
@@ -0,0 +1,69 @@
+from collections import OrderedDict
+from decimal import Decimal
+from typing import IO, Mapping, Optional, Tuple
+
+import yaml
```
baseplate-admin, line 5: Wrap this in a try except

ninja/management/commands/export_openapi_schema.py
denizdogan: We need `NinjaSafeDumper` for the `yaml.dump` call after

ninja/management/commands/export_openapi_schema.py
denizdogan: Yeah, I wasn't sure what to do with this part. I can remove the else and keep the comment. 👍

ninja/yaml.py
denizdogan: I tried with try/except before, but realized that the rest of the module makes no sense if `yaml` cannot be imported. What should we do in the `except` clause?

baseplate-admin: Lets assume someone wants to access this internal module. But he doesnt have `PyYAML` installed. 

So i think we should remove try except from the parser and move this logic into here. With same error message.

ninja/management/commands/export_openapi_schema.py
baseplate-admin: Lets move this try...except into NinjaSafeDumper.

denizdogan:
@baseplate-admin Phew, okay, I hope I didn't misunderstand your latest feedback. When you have time, please take another look and let me know if there's any other issues!

baseplate-admin (approved):
LGTM


Awesome work @denizdogan. 

This PR is ready for merging :)


Thanks for listening to my feedbacks

denizdogan:
@baseplate-admin Thanks. Are you waiting for me for something regarding this PR?

baseplate-admin:
Umm sadly i dont have merge access :}

So i will have to ping @vitalik for merging

baseplate-admin:
Hey @vitalik,

Do you think having `openapi.yaml` is a bad idea?  

[Swagger editor](https://editor.swagger.io/)  by default supports `yaml`

vitalik:
I do not like to add extra dependency for functionality that 99.99% users do not use

overall adding yaml is as simple as adding one extra method to NinjaAPI:

```Python

class NinjaWithYaml(NinjaAPI):

    def get_sechema_yaml(self, prefix=None):
          return yaml.dumps(self.get_openapi_schema(prefix))
         
```

PR #558, opened 2022-09-07
ModelSchema: allow model_fields = []

hiaselhans:
This PR tests model_fields and model_exclude against None, so the following two ModelSchemes are valid:

``` python
class SampleSchema3(ModelSchema):
    class Config:
        model = User
        model_fields  = []


class SampleSchema4(ModelSchema):
    class Config:
        model = User
        model_exclude  = [] # equivalent to model_fields = "__all__"
```

### DIFF
--- a/ninja/orm/factory.py
+++ b/ninja/orm/factory.py
@@ -44,7 +44,7 @@ class SchemaFactory:
     ) -> Type[Schema]:
         name = name or model.__name__
 
-        if fields and exclude:
+        if fields is not None and exclude is not None:
             raise ConfigError("Only one of 'fields' or 'exclude' should be set.")
 
         key = self.get_key(model, name, depth, fields, exclude, custom_fields)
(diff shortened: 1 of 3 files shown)


SmileyChris:
What's the reasoning behind this versus what already works?

Especially `model_fields = []` doesn't naturally infer that you mean all fields. It reads as the opposite!

hiaselhans:
its more idiomatic and it allows for a modelschema without any model_fields.

I think your deduction is wrong, please read the test included:

model_fields = [] => no fields
model_exclude = [] => all fields

I think thats exactly what one would expect and its much easier to remember than __all__

hiaselhans:
No worries, and sorry for the short answer :)

After all it might turn out to be a really just a matter of preference.

In my opinion, if i can exclude one field, I should also be able to exclude zero fields. If i can include one, i should also be able to include 0.

vitalik:
sorry, closing this - as of django ninja v1 - we are following same guides as django model forms - so now recomended way is

class Foo(ModelSchema):
    class Meta:
          model = Foo
          fields = ['some'] # NOT !! model_fields

Issue #573, opened 2022-09-24
Add "servers" inside openapi.json schema


Vaiders:
**Is your feature request related to a problem? Please describe.**
According to openapi specification there is an optional "servers" field that contain a list of servers. Reference (https://spec.openapis.org/oas/v3.0.2#fixed-fields)

**Describe the solution you'd like**
add a new argument to NinjaAPI class that would expect list of objects with url and description fields

Example implementation:
api =.NinjaAPI(servers=['url': 'https://api-example.com', 'description': 'Development server'])

Vaiders:
Working on this during DjangoCon Europe

PR #574, opened 2022-09-24
#573 Added a support for openapi ServerObject

Vaiders:
Implementation for #573


Vaiders:
Missing:
Unit tests
Documentation/Example

Issue #586, opened 2022-10-11
Add support for adding multiple examples in request bodies and path, query, cookie, and header params


xncbf:
Does django-ninja support or plan to support the following features?

ref: https://github.com/tiangolo/fastapi/pull/1267

xncbf:
This is my commit.
https://github.com/xncbf/django-ninja/commit/bae69821dd3968b2e9d6fec16ed4ad080a3ba95a
However, it is difficult because the process of creating fastapi and openapi schema is slightly different. Can someone implement a test to put examples in requestBody and Response ?
A get_openapi_schema generated by a fixture that does not go through an endpoint is not sufficient.

xncbf:
I just uploaded my implementation to PR.

PR #588, opened 2022-10-12
Add support for adding multiple examples in request bodies and path, query, cookie, and header params

xncbf:
#586 

It was developed based on the fastapi method.
I used type ignore to pass pre-commit , but I need some help.

### DIFF
--- a/ninja/encoders.py
+++ b/ninja/encoders.py
@@ -1,0 +1,172 @@
+# type: ignore
+import dataclasses
+from collections import defaultdict
+from enum import Enum
+from pathlib import PurePath
+from types import GeneratorType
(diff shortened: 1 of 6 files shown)


vitalik:
already implemented

```python
  @api.get("/items")
  def list_items(
      request,
      search: str = Query(
          ...,
          max_length=5,
          examples={
              "normal": {
                  "summary": "A normal example",
                  "description": "A **normal** string works correctly.",
                  "value": "Foo",
              },
              "invalid": {
                  "summary": "Invalid data is rejected with an error",
                  "value": "MoreThan5Length",
              },
          },
      ),
  ):
  ```

PR #597, opened 2022-10-26
make enums show up in schema, fixes #596

### LINKED ISSUE #596: [BUG] Multiple enums not shown in OpenAPI docs
**Describe the bug**
Multiple enums not shown in OpenAPI docs. Might be related to #537 although I am not using nested classes.

I simply have multiple enums in a schema, and only one of them shows up in the OpenAPI docs. I have created a PR which fixes the issue in my end.

 - Python version: 3.9.10
 - Django version: 4.1.2
 - Django-Ninja version: 0.19.1
 - Pydantic version: 1.9.2

tykling:
this added elif catches the cases where an array of items needs a subschema

### DIFF
--- a/ninja/openapi/schema.py
+++ b/ninja/openapi/schema.py
@@ -308,6 +308,11 @@ def flatten_properties(
         definition = definitions[def_name]
         yield from flatten_properties(prop_name, definition, prop_required, definitions)
 
+    elif "items" in prop_details and "$ref" in prop_details["items"]:
+        def_name = prop_details["items"]["$ref"].split("/")[-1]
+        definition = definitions[def_name]
+        yield from flatten_properties(prop_name, definition, prop_required, definitions)
+
     elif "properties" in prop_details:
         required = set(prop_details.get("required", []))
         for k, v in prop_details["properties"].items():


tykling:
This should be linked to #596

vitalik:
Hi @tykling 
Thank you for PR
Would you be able to add a unit test for this case ?

tykling:
absolutely, will look at it this evening

tykling:
Welp, that was more of a faff than I suspected. Spent all evening yesterday and got nowhere. Writing a unittest for something this involved requires a deeper understanding of the django-ninja codebase than I currently have. 

I might give it another go in the weekend with fresh eyes.

tykling:
But for now I can say this: 

The issue was encounteres with these enums https://github.com/bornhack/bma/blob/develop/src/files/models.py#L11-L37 when used in this filter class https://github.com/bornhack/bma/blob/develop/src/files/schema.py#L136-L147 used in this api filters in https://github.com/bornhack/bma/blob/develop/src/files/api.py#L385

One of the enums is used in the model https://github.com/bornhack/bma/blob/develop/src/files/models.py#L91-L96 and the others are only used in the filter. The one used in the model shows up in the schema, the other two do not show up in the schema.

With this PR all three show up in the schema.

vitalik:
fixed in #679

PR #598, opened 2022-10-27
Set correct media type in docs for custom renderer

Svenito:
When specifying a renderer with a custom media type, the generated schema docs don't reflect the media type header.
It always returns application/json.

This patch uses the media_type from the custom renderer to render the response schema media type


vitalik:
Thank you

PR #604, opened 2022-11-07
Refactor pagination to not overwrite _ninja_contribute_args

OtherBarry:
I've written some of my own decorators for things like sorting and filtering that would take advantage of the `_ninja_contribute_args` functionality. However, the current implementation of the pagination decorator overwrites any values that may already be in `_ninja_contribute_args`, meaning that you can not use it in conjunction with other decorators.

This PR:

- Creates a function called `inject_contribute_args` that will add new args without overwriting existing ones
- Updates the `_inject_pagination` function to use the new function
- Adds tests for `inject_contribute_args` in a new file called `test_signature_utils`
- Moves some other tests relating to signature utils from `test_misc` to `test_signature_utils`


OtherBarry:
@vitalik is there anything else I need to add to get this merged? I was hoping to see it in the v0.20 release.

vitalik:
Hi @OtherBarry

yeah - there was an open question:

```
        else:
            raise ConfigError(f"{func} has an invalid _ninja_contribute_args value")
````
Why would this case ever happen if inject function always initialies a list ? I think it does not need this extra check...

OtherBarry:
The intent was so that you'd get a nicer error message than `object has no attribute 'append'` if the `_inject_contribution_args` variable had been incorrectly modified by some other function/decorator, but I guess that's a pretty unlikely scenario. I'll remove it shortly.

OtherBarry:
@vitalik removed, after a slight delay

OtherBarry:
...and now passing - serves me right for thinking I could get away with using the github web ui for a quick fix.

OtherBarry:
@vitalik is this also covered by the changes in v1.0 mentioned in #779 ?

vitalik:
@OtherBarry - it now always appends instead of overwrite

PR #608, opened 2022-11-11
Allow Renderer charset to be None

linssen:
We have an especially strict client for an API which demands the content-type be exactly what was entered, which is also strictly defined. When defining a renderer, you previously also had to specify a charset which was then appended to the content-type on response.

* Make charset optional when defining a renderer
* Only append the charset to the content type if it's not None
* Fix a type on the media_type prop which was incorrectly set as optional

Hope this isn't too controversial?

### DIFF
--- a/ninja/main.py
+++ b/ninja/main.py
@@ -411,7 +411,11 @@ class NinjaAPI:
         return HttpResponse("", content_type=self.get_content_type())
 
     def get_content_type(self) -> str:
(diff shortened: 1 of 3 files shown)


linssen:
On weekend reflection, I think the lib is wrong and having a charset is fine. The commit about the media type being a string I think still stands, but it's the most minor of minor things. Apologies for the timewasting!

Issue #623, opened 2022-12-08
Improve startup time


tybirk-accuranker:
**Is your feature request related to a problem? Please describe.**
We recently added Ninja to a Django project. This increased app reload times significantly. It seems that the culprit is the call to "is_debug_server" on startup which takes more than a second

See
![image](https://user-images.githubusercontent.com/95625870/206461698-861e038b-c4ee-4566-bddd-58c750bbdc04.png)

It would be lovely if there was some way to either disable this check, or if the check could run faster.

OtherBarry:
Setting the environment variable `NINJA_SKIP_REGISTRY` to True might halve the number of calls to is_debug_server.

I'll put a PR in to make it skip all calls.

vitalik:
Hi @tybirk-accuranker would you be able to test your case again latest master version of jango ninja (that inlucdes #624 merge) ?

flaeppe:
Wouldn't it be possible to just check `sys.argv` instead?

flaeppe:
Also, I think merging #437 would remove a call to `is_debug_server` (via removing a call to `debug_server_url_reimport`)

vitalik:
> Wouldn't it be possible to just check `sys.argv` instead?

well debug server can be started with custom commands or in tests so no way to catch correct word in the argv

> Also, I think merging https://github.com/vitalik/django-ninja/pull/437 would remove a call to is_debug_server (via removing a call to debug_server_url_reimport)

I'm still thinking about it... maybe the registry can actually go away as the only reason for it is to find correct urls in docs

flaeppe:
> I'm still thinking about it... maybe the registry can actually go away as the only reason for it is to find correct urls in docs

Maybe a question for a different issue, but am I assuming things correctly saying that the registry is a mitigation for Django silently swallowing multiple views registered under the same URL route?

If so, that might be able to become a system check or something similar.

tybirk-accuranker:
> Hi @tybirk-accuranker would you be able to test your case again latest master version of django ninja (that inlucdes #624 merge) ?

Hi @vitalik, 
Thanks for the quick response to the issue! The newest master has cut this time in half so its ~0.5 seconds instead of 1 second  :muscle: Still some time to gain, but it's much better.

PR #624, opened 2022-12-08
remove context from inspect.stack

hiaselhans:
This is a quick take on #623 

https://docs.python.org/3/library/inspect.html#inspect.stack

inspect.stack(context=1) adds context lines to the stackframe.
using context=0 file loads can be avoided and might bring an improved startup time already


vitalik:
Hi @hiaselhans can you merge it with #625 to drop

PR #625, opened 2022-12-08
remove python 3.6

hiaselhans:
python 3.6 isn't available anymore on gh-actions.

- removes version 3.6
- adds version 3.11


hiaselhans:
depends on #438 to be merged

hiaselhans:
Oh! Just saw that you fixed it already! Thanks!

that way #438 is obsolete :)

PR #635, opened 2022-12-18
Refactor serialization of pydantic response object data and move it to a NinjaAPI method for easier overriding

adamsussman:
This change takes the logic for converting a pydantic response object to a dict (pre-json conversion) and refactors it into a method on NinjaAPI.  The actual functionality and logic does not change, just the location.

The purpose of this is to make this logic easier to override by subclassing the NinjaAPI object.  If this is not the best way to organize this logic step, I am happy to do it some other way.  The main thing is to have the ability to override/replace this step with custom logic.

Why have this ability?  Sometimes you need more control over the conversion of the response object to a dict.  For example, pydantic's `dict` method supports `include` and `exclude` parameters with which you can implement output field filtering.  Alternatively, there are more complex and featureful methods of serializing a pydantic model that may be desirable.   In any case, this move allows one to modify the pydantic `dict` arguments as needed, or replace the `dict` call with something else entirely, if wanted.

### DIFF
--- a/ninja/main.py
+++ b/ninja/main.py
@@ -20,6 +20,7 @@ from ninja.errors import ConfigError, set_default_exc_handlers
 from ninja.openapi import get_schema
 from ninja.openapi.schema import OpenAPISchema
 from ninja.openapi.urls import get_openapi_urls, get_root_url
+from ninja.operation import ResponseObject
 from ninja.parser import Parser
 from ninja.renderers import BaseRenderer, JSONRenderer
 from ninja.router import Router
@@ -28,6 +29,7 @@ from ninja.utils import is_debug_server, normalize_path
 
 if TYPE_CHECKING:
     from .operation import Operation  # pragma: no cover
+    from .schema import Schema
 
 __all__ = ["NinjaAPI"]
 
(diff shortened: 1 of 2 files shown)


vitalik:
@adamsussman Can we name it simpler ? maybe just `serialize_response` ?

adamsussman:
hi @vitalik! is this mergable or is more work/thought needed?

vitalik:
@adamsussman sorry I'm still thinking/concerned about the name  `serialize_response`... because tehnically it's not serialisation, but sort of conversion to dict

any ideas how to better name it ?

adamsussman:
Yes, I am having a similar challenge naming a module I am writing which is a "value added" pydantic.dict conversion (something that covers expansions and easier field limits).

I have been thinking of using the "marshaling" term, which used to be in more common use in the IPC days before REST and JSON.  Marshaling is similar but not necessarily the same as serialization.  So in this case, we could take the following terminology:

- serialization: Conversion between an in memory data structure (say a dict) and a string (say JSON)
- marshaling: Conversion between one data structure and another (say dict <-> BaseModel)

Thus you would have "deserialize" + "unmarshal" on the request and "marshal" + "serialize" on the response.

I am being a bit loose here as often serialization and marshaling have been used as synonymously.  So, someone is sure to split hairs over it.

See: https://en.wikipedia.org/wiki/Marshalling_(computer_science)

Aside from "marshaling", other possible terms I have seen for magical data structure include:
- type "casting"
- type "conversion"
- broadcasting

vitalik:
@adamsussman maybe `prepare_response` ?

adamsussman:
> @adamsussman maybe `prepare_response` ?

hmm, well, looking at it naively, to me "prepare_response" covers the entire pydantic -> json process, so if we moved both steps into one method called "prepare_response", it would make sense to me.  If it only did one step or the other, I think it would be confusing.

adamsussman:
@vitalik How about this variation: I've moved the 'pydantic to dict' rendering step to the Renderer class with an explicit method name and put some of the response schema juggling back into Operation.  The renderer class is easy to override on the NinjaAPI object and no other changes to the main object are required.

adamsussman:
ping @vitalik What do you think of the above iteration (move code to Renderer class)

vitalik:
@adamsussman 

yes, that might be a better place... I will try to go through issues - maybe some extra stuff or common questions can land there as well

PR #639, opened 2022-12-28
RapiDoc added and tested.

webdastur:


webdastur:
@vitalik Hi Vitaliy. I added RapiDoc and tested it locally, production. Please check this. Thanks.

webdastur:
@vitalik Is there anything I should add to this PR for approval?

ninja/openapi/views.py
```diff
@@ -16,12 +16,16 @@
 __all__ = ["default_home", "openapi_json", "openapi_view", "openapi_view_cdn"]
 
 
-render_swagger = ninja_settings.DOCS_VIEW == "swagger"
-view_tpl = "ninja/swagger.html" if render_swagger else "ninja/redoc.html"
+render_doc = "swagger"
+
+if ninja_settings.DOCS_VIEW == "redoc":
```
baseplate-admin, line 21: Can we use a walrus operator here?

baseplate-admin (changes requested):

vitalik:
> @vitalik Is there anything I should add to this PR for approval?

well, I'm thinking about some more flexible solution so that anyone can add their favourite docs engine...

ninja/openapi/views.py
vitalik: > Can we use a walrus operator here?

no - since we support pytthon3.7 and pytrhon3.6 probably works as well

vitalik:
Hi @webdastur 


since django-ninja v1 allows to make plugable openapi viewers

```Python
api = NinjaAPI(docs=YourEngine())
```

see how redoc implemented here https://github.com/vitalik/django-ninja/blob/2f5a5766ea6e86784d5eb5cafc0ab3e0e5f60e96/ninja/openapi/docs.py#L58-L77

you can make a separate package for Elements - and we can promote it in the docs

Issue #381, opened 2022-03-01
AttributeError when running export_openapi_schema


lausek:
Running

```
./manage.py export_openapi_schema
```

Results in the following error:

```
Traceback (most recent call last):
  File "./manage.py", line 22, in <module>
    main()
  File "./manage.py", line 18, in main
    execute_from_command_line(sys.argv)
  File ".../venv/lib/python3.7/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File ".../venv/lib/python3.7/site-packages/django/core/management/__init__.py", line 413, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File ".../venv/lib/python3.7/site-packages/django/core/management/base.py", line 354, in run_from_argv
    self.execute(*args, **cmd_options)
  File ".../venv/lib/python3.7/site-packages/django/core/management/base.py", line 398, in execute
    output = self.handle(*args, **options)
  File ".../venv/lib/python3.7/site-packages/ninja/management/commands/export_openapi_schema.py", line 45, in handle
    api = self._get_api_instance(options["api"])
  File ".../venv/lib/python3.7/site-packages/ninja/management/commands/export_openapi_schema.py", line 17, in _get_api_instance
    return resolve("/api/").func.keywords["api"]  # type: ignore
AttributeError: 'function' object has no attribute 'keywords'
```

Debugging with pdb on the corresponding line:

```
-> return resolve("/api/").func.keywords["api"]  # type: ignore
(Pdb) m = resolve("/api/").func
(Pdb) m
<bound method PathView._sync_view of <ninja.operation.PathView object at 0x7f6b81f7a5c0>>
(Pdb) dir(m)
['__call__', '__class__', '__delattr__', '__dir__', '__doc__', '__eq__', '__format__', '__func__', '__ge__', '__get__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__self__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', 'csrf_exempt']
```

So it seems the original intention of this line was to determine the NinjaAPI instance. Yet, I fail to understand how a `keywords` attribute could be present on a bound method object. Wouldn't it be better to get the API instance like this?

```
return resolve("/api/").func.__self__.api 
```

**Versions:**
 - Python version: 3.7.3
[...]

PR #653, opened 2023-01-17
improve error message in export_openapi_schema

taobojlen:
Closes https://github.com/vitalik/django-ninja/issues/381

Currently, if you don't have a NinjaAPI instance at the URL `/api/`, then the `export_openapi_schema` command will fail with an opaque error message. This PR improves the error message in that scenario, telling the user to pass the path to the API using the `--api` flag.

### DIFF
--- a/ninja/management/commands/export_openapi_schema.py
+++ b/ninja/management/commands/export_openapi_schema.py
@@ -27,7 +27,12 @@ class Command(BaseCommand):
 
     def _get_api_instance(self, api_path: Optional[str] = None) -> NinjaAPI:
         if not api_path:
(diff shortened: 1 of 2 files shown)


taobojlen:
thanks for approving the CI run here!

there was a failing step that complained about missing code coverage. i fixed this with an additional test -- not a super elegant one, but it keeps the code coverage at 100%.

vitalik:
> i fixed this with an additional test -- not a super elegant one, but it keeps the code coverage at 100%.

Hi @taobojlen - I do not see this test in PR ? are you sure it's pushed ?

taobojlen:
@vitalik 🤦 sorry, my mistake, it's pushed now!

PR #663, opened 2023-01-25
Add a recipe for handling the file uploads

barseghyanartur:
Add a recipe for handling the file uploads.

### DIFF
--- a/docs/docs/tutorial/other/crud.md
+++ b/docs/docs/tutorial/other/crud.md
@@ -53,7 +53,7 @@ def create_employee(request, payload: EmployeeIn):
 
     You can pass it as `**kwargs` to the Django model's `create` method (or model `__init__`).
 
-See the recipe below for the file upload:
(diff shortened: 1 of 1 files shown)


vitalik:
Hi @barseghyanartur 

I think you can simplify saving:

```Python
from ninja import UploadedFile, File

@api.post("/employees")
def create_employee(request, payload: EmployeeIn, cv: UploadedFile = File(...)):
    payload_dict = payload.dict()
    employee = Employee(**payload_dict)
    employee.cv.save(cv.name, cv) # will save model instance as well
    return {"id": employee.id}
```

barseghyanartur:
@vitalik:

Will do.

vitalik:
> Shall I, perhaps, add a second recipe for when you don't work with models, but still want to upload files?

yes - makes sense

barseghyanartur:
@vitalik:

Ready.

vitalik:
Thank you

PR #667, opened 2023-01-28
Fix #666: Support auth inheritance for routers

### LINKED ISSUE #666: Router does not support auth inheritance
**Describe the bug**
When you try adding a new router to an existing router, the leaf router doesn't inherit the top-level auth.
Consider the below example:

```py
from ninja import NinjaAPI, Router
from ninja.security import APIKeyQuery

api = NinjaAPI()

r1 = Router()
r2 = Router()
r3 = Router()


class Auth(APIKeyQuery):
    def __init__(self, secret):
        self.secret = secret
        super().__init__()

    def authenticate(self, request, key):
        if key == self.secret:
            return key


api.add_router("/r1", r1, auth=Auth("r1_auth"))
r1.add_router("/r2", r2)
r2.add_router("/r3", r3)


@r1.get("/")
def op1(request):
    return request.auth


@r2.get("/")
def op2(request):
    return request.auth


@r3.get("/")
def op3(request):
    return request.auth

```

So the auth provided for router `r1` won't be present for any operations in routers `r2` and `r3` even though it comes under it.
This is only for routers though. If we add auth when we initialize `NinjaApi()`  it propagates down to all routers and endpoints even if we provide the auth when initializing router r1 as `r1 = Router(auth=Auth("r1_auth"))`. Screenshot of the above code is shown below.

Screen Shot 2023-01-28 at 9 44 00 AM

I think it's super helpful to include this in the documentation if we don't plan to support it as a dev can easily misinterpret it and in turn, pose a security threat to the app they are building.

 - Python version: 3.9.12
 - Django version: 4.1.5
 - Django-Ninja version: 0.20.0
 - Pydantic version: 1.10.4

aasiffaizal:
This PR fixes #666 and adds support to the inheritance of auth in routers.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,7 +12,7 @@ repos:
         additional_dependencies: ["django-stubs", "pydantic"]
         exclude: (tests|docs)/
   - repo: https://github.com/psf/black
-    rev: "22.3.0"
+    rev: "23.1.0"
     hooks:
       - id: black
         exclude: docs/src/
--- a/ninja/main.py
+++ b/ninja/main.py
@@ -353,7 +353,7 @@ class NinjaAPI:
             prefix = normalize_path("/".join((parent_prefix, prefix))).lstrip("/")
 
         self._routers.extend(router.build_routers(prefix))
-        router.set_api_instance(self)
+        router.set_api_instance(self, parent_router)
 
     @property
     def urls(self) -> Tuple[List[Union[URLResolver, URLPattern]], str, str]:
--- a/ninja/openapi/schema.py
+++ b/ninja/openapi/schema.py
@@ -179,7 +179,6 @@ class OpenAPISchema(dict):
         by_alias: bool = True,
         remove_level: bool = True,
     ) -> Tuple[DictStrAny, bool]:
-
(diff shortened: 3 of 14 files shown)


aasiffaizal:
Thanks for approving the CI run. Fixed the lint errors, it shouldn't be failing now.

aasiffaizal:
@vitalik Could you please rerun the CI checks and take a pass on this if it's not an issue? 😃

aasiffaizal:
@vitalik hmm is it a flake or something with the code style test?

Seems like black in the code style is failing for files which aren't touched in this PR. Tried `make lint` in my local and this was the result.

Screen Shot 2023-02-01 at 7 40 13 AM

aasiffaizal:
@vitalik, I checked why black in codestyle was breaking for untouched files, turns out black released a new version [23.1.0](https://pypi.org/project/black/23.1.0/) 10 hours back and that was used for my last CI run. So I have updated the PR by formatting those files with the new version and updated the version of black in pre-commit-config yaml file as well.

Confirmed the checks in my forked repo in a PR over [here](https://github.com/aasiffaizal/django-ninja/pull/2), the checks are passing.

vitalik:
Thank you

PR #673, opened 2023-01-30
Somewhat advanced filtering with FilterSchema

l1b3r:
The documentation section [explains](https://django-ninja.rest-framework.com/guides/input/query-params/#using-schema) how to add query parameters to an API handler encapsulated into a separate schema:
```python
class Filters(Schema):
    limit: int = 100
    offset: int = None
    query: str = None
    category__in: List[str] = Field(None, alias="categories")


@api.get("/filter")
def events(request, filters: Filters = Query(...)):
    return {"filters": filters.dict()}
```
 That is helpful already as it:
1. Takes care of documenting and validating the fields in the schema;
2. Provides an explicit separation of concerns to the parameters of the method.

However, it seems that when it comes to actually applying those filters in the API handler, the developer is left on their own and will probably have to do something like this all the time:
```python
@api.get("/filter")
def events(request, filters: Filters = Query(...)):
    events = Event.objects.all()
    if filters.query:
        events = events.filter(name__icontains=filters.query)
    if filters.category__in:
        events = events.filter(category__in=filters.category__in)
    # the more filters the more ifs
    return events
```

While this approach is obviously not very DRY and readable, it also exposes unexperienced Django developers to a risk of getting unexpected filtering results when multiple `.filter` clauses are used. There is a slight [difference](https://stackoverflow.com/questions/8164675/chaining-multiple-filter-in-django-is-this-a-bug) in how Django translates ORM queries with multiple `.filter`s into SQL queries.

In an attempt to bridge the gap between the API layer and the DB execution layer, this PR is introducing a `FilterSchema` class. It inherits from Ninja's (Pydantic's) `Schema` and provides extra functionality that would convert filter values into ready to use Django Q expressions that can be fed straight into the queryset.

Below I will include the documentation for the feature from this PR as is. Hopefully, it explains it sufficiently enough.

Please let me know what you think about the proposal and if it has a chance to make it.

- - -

# Filtering

If you want to allow the user to filter your querysets by a number of different attributes, it makes sense
to encapsulate your filters into a `FilterSchema` class. `FilterSchema` is a regular `Schema`, so it's using all the
necessary features of Pydantic, but it also adds some bells and whistles that ease the translation of the user-facing filtering
parameters into database queries. 

Start off with defining a subclass of `FilterSchema`:

```python hl_lines="6 7 8"
from ninja import FilterSchema, Field
from typing import Optional


    name: Optional[str] = Field(q='name__icontains')
    author: Optional[str] = Field(q='author__name__icontains')
    created_after: Optional[datetime] = Field(q='created__gte')
```

Pay attention to the field definition. `FilterSchema` requires that you provide a kwarg `q`, which should contain
a keyword argument name which will then under the hood be used to translate the filter values into a [Q](https://docs.djangoproject.com/en/3.1/topics/db/queries/#complex-lookups-with-q-objects) expression used for filtering the queryset.

Next, use this schema in conjunction with `Query` in your API handler:
```python hl_lines="2"
@api.get("/books")
def list_books(request, filters: BookFilterSchema = Query(...)):
    books = Book.objects.all()
    books = filters.filter(books)
    return books
```

Just like described in [defining query params using schema](./query-params.md#using-schema), Django Ninja converts the fields
defined in `BookFilterSchema` into query parameters.

You can use a shorthand one-liner `.filter()` to apply those filters to your queryset:
```python hl_lines="4"
@api.get("/books")
def list_books(request, filters: BookFilterSchema = Query(...)):
    books = Book.objects.all()
    books = filters.filter(books)
    return books
```

Alternatively, you can get the prepared `Q`-expression and perform the filtering yourself.
That can be useful, when you have some additional queryset filtering on top of what you expose to the API:
```python hl_lines="4 5 7 8"
@api.get("/books")
def list_books(request, filters: BookFilterSchema = Query(...)):

    # Never serve books from inactive publishers and authors
    q = Q(author__is_active=True) | Q(publisher__is_active=True)
    
    # But allow filtering the rest of the books
    q &= filters.get_filter_expression()
    return Book.objects.filter(q)
```

By default, the filters will behave the following way:

* `None` values will be ignored and not filtered against;
* Every non-`None` field will be converted into a `Q`-expression based on the `Field` definition of each field;
* All `Q`-expressions will be merged into one using `AND` logical operator;
* The resulting `Q`-expression is used to filter the queryset and return you a qeryset with a `.filter` clause applied.

## Filtering by Nones
You can make the `FilterSchema` treat `None` as a valid value that should be filtered against.

This can be done on a field level with a `ignore_none` kwarg:
```python hl_lines="3"
    name: Optional[str] = Field(q='name__icontains')
    tag: Optional[str] = Field(q='tag', ignore_none=False)
```

This way when no other value for `"tag"` is provided by the user, the filtering will always include a condition `tag=None`.

You can also specify this settings for all fields at the same time in the Config:
```python hl_lines="6"
    name: Optional[str] = Field(q='name__icontains')
    tag: Optional[str] = Field(q='tag', ignore_none=False)
    
    class Config:
        ignore_none = False
```

## Combining expressions
By default filters are joined together using `AND` operator. This can be changed in the schema Config:

```python
    name: Optional[str] = Field(q='name__icontains')
    author: Optional[str] = Field(q='author__name__icontains')
    created_after: Optional[datetime] = Field(q='created__gte')
    
    class Config:
        expression_connector = 'OR'     # can be 'AND', 'OR', 'XOR'
```

With such filtering configuration the endpoint...
```python
http://localhost:8000/api/books?name=harry&author=poe
```
...will return Harry Potter series as well as books from Edgar Allan Poe.

## Custom expressions
Sometimes you might want to have complex filtering scenarios that cannot be handled by individual Field annotations.
For such cases you can implement your own custom logic in a `custom_expression` method:

```python
    name: Optional[str]         # No need to supply "q" kwarg
    popular: Optional[bool]     # when custom expression is used

    def custom_expression(self) -> Q:
        q = Q()
        if self.name:
            q &= Q(name__icontains=self.name)
        if self.popular:
            q &= (
                Q(view_count__gt=1000) |
                Q(downloads__gt=100) |
                Q(tag='popular')
            )
        return q
```

### DIFF
--- a/docs/docs/guides/input/filtering.md
+++ b/docs/docs/guides/input/filtering.md
@@ -13,13 +13,10 @@ from typing import Optional
 
 
 class BookFilterSchema(FilterSchema):
-    name: Optional[str] = Field(q='name__icontains')
-    author: Optional[str] = Field(q='author__name__icontains')
-    created_after: Optional[datetime] = Field(q='created__gte')
+    name: Optional[str]
+    author: Optional[str]
+    created_after: Optional[datetime]
 ```
-
-Pay attention to the field definition. `FilterSchema` requires that you provide a kwarg `q`, which should contain
-a keyword argument name which will then under the hood be used to translate the filter values into a [Q](https://docs.djangoproject.com/en/3.1/topics/db/queries/#complex-lookups-with-q-objects) expression used for filtering the queryset.
 
 
 Next, use this schema in conjunction with `Query` in your API handler:
@@ -43,9 +40,12 @@ def list_books(request, filters: BookFilterSchema = Query(...)):
     return books
 ```
 
-Alternatively, you can get the prepared `Q`-expression and perform the filtering yourself.
-That can be useful, when you have some additional queryset filtering on top of what you expose to the API:
-```python hl_lines="4 5 7 8"
+Under the hood, `FilterSchema` converts its fields into [Q expressions](https://docs.djangoproject.com/en/3.1/topics/db/queries/#complex-lookups-with-q-objects) which it then combines and uses to filter your queryset.
+
+
+Alternatively to using the `.filter` method, you can get the prepared `Q`-expression and perform the filtering yourself.
+That can be useful, when you have some additional queryset filtering on top of what you expose to the user through the API:
+```python hl_lines="5 8"
 @api.get("/books")
 def list_books(request, filters: BookFilterSchema = Query(...)):
 
@@ -63,6 +63,65 @@ By default, the filters will behave the following way:
 * Every non-`None` field will be converted into a `Q`-expression based on the `Field` definition of each field;
 * All `Q`-expressions will be merged into one using `AND` logical operator;
 * The resulting `Q`-expression is used to filter the queryset and return you a qeryset with a `.filter` clause applied.
+
+## Customizing Fields
+By default, `FilterSet` will use the field names to generate Q expressions:
+```python
+class BookFilterSchema(FilterSchema):
+    name: Optional[str]
+```
+The `name` field will be converted into `Q(name=...)` expression.
+
+When your database lookups are more complicated than that, you can explicitly specify them in the field definition using a `"q"` kwarg:
+```python hl_lines="2"
+class BookFilterSchema(FilterSchema):
+    name: Optional[str] = Field(q='name__icontains') 
+```
+You can even specify multiple lookup keyword argument names as a list:
+```python hl_lines="2 3 4"
+class BookFilterSchema(FilterSchema):
+    search: Optional[str] = Field(q=['name__icontains',
+                                     'author__name__icontains',
+                                     'publisher__name__icontains']) 
+```
+By default, field-level expressions are combined using `"OR"` connector, so with the above setup, a query parameter `?search=foobar` will search for books that have "foobar" in either of their name, author or publisher.
+
+
+## Combining expressions
+By default,
+
+* Field-level expressions are joined together using `OR` operator.
+* The fields themselves are joined together using `AND` operator.
+
+So, with the following `FilterSchema`...
+```python
+class BookFilterSchema(FilterSchema):
+    search: Optional[str] = Field(q=['name__icontains', 'author__name__icontains'])
+    popular: Optional[bool]
+```
+...and the following query parameters from the user
+```
+http://localhost:8000/api/books?search=harry&popular=true
+```
+the `FilterSchema` instance will look for popular books that have `harry` in the book's _or_ author's name. 
+
+
+You can customize this behavior using an `expression_connector` argument in field-level and class-level definition:
+```python hl_lines="3 7"
+class BookFilterSchema(FilterSchema):
+    active: Optional[bool] = Field(q=['is_active', 'publisher__is_active'],
+                                   expression_connector='AND')
+    name: Optional[str] = Field(q='name__icontains')
+    
+    class Config:
+        expression_connector = 'OR'     # can be 'AND', 'OR', 'XOR'
+```
+
+Now, a request with these query parameters 
+```
+http://localhost:8000/api/books?name=harry&active=true
+```
+...shall search for books that have `harry` in their name _or_ are active themselves _and_ are published by active publishers.
 
 
 ## Filtering by Nones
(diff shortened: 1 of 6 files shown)


vitalik:
Hi @l1b3r 

Thank you for contribution
Looks promising

Could you check tests for earlier python versions (<3.9)

l1b3r:
@vitalik Fixed something in the tests, requesting CI re-run.

vitalik:
Hi @l1b3r 

I like the idea, just for now sitting on it thinking if the naming is fine (like here `filters.filter(books)` - mabye `filter_queryset` or `apply`... not sure..)

Also - looks like `q` argument is required:
```Python
name: Optional[str] = Field(q='name__icontains')
```

```Python
name: Optional[str] = Field(...)
```
maybe it's better to make it optional - in this case ^ it will run `.filter(name=value)`


Last - I see often case where I need to run search on multiple fields... maybe `q` should accept list of fields: 


```Python
    query: Optional[str] = Field(q=['name__icontains', 'author__name__icontains'])
```

l1b3r:
Hey @vitalik,

First of all, Thank you for having a look in this PR. Let me address your feedback one by one:

1. In regards to the naming, the current names were picked in an attempt to align with the conventions of Django:
    * `FilterSchema.filter` is named `.filter` since all it does essentially is invoke Django's `QuerySet.filter`. My assumption is that the developer used to Django should have an immediate expectation of what the method does and returns when it's called `.filter`. However I do acknowledge a slight inconsistency: `QuerySet.filter` is a method of the queryset that takes in filter arguments, while `FilterSet.filter` is the other way around a method of the filter arguments that takes a queryset. So, my assumption may not be entirely correct.
    * kwarg `q` is named `q` since the string value provided to it goes straight into Django's `Q` expression as a keyword argument;
    * setting `expression_connector` is called this way since Django's Q refers to the `"AND"`, `"OR"` and `"XOR"` as [connectors](https://github.com/django/django/blob/main/django/db/models/query_utils.py#L48). However, I am not so fond of this name myself as albeit its descriptiveness it's too long IMO, especially when provided in a Filed definition (see declaration of field `search` in point 3 below);

    Having that said, I am not holding on to any of those names and will not object you willing to change them into something else.

2. That makes sense, will do.

3. With the current state of the PR you'd be able to achieve your goal with a custom expression. Though I see that this may be too verbose for such not-so-complex cases and it should be more declarative. So I agree with your suggestion and propose some additions to it. Consider the following filter declaration:
    ```python
        tag: Optional[str]
        name: Optional[str] = Field(q='name__icontains')
        search: Optional[str] = Field(q=['name__icontains', 'author_name__icontains'], expression_connector='OR')
        popular: Optional[bool]

        def filter_popular(self, value) -> Q:
            return Q(view_count__gt=1000) if value else Q()
    ```
    * For `tag`, the query would be `Q(tag=...)` (where `...` is whatever value the user provided). This is what you are suggesting in point 2.
    * For `name` the query would be `Q(name__icontains=...)`;
    * For `search` the query would be `Q(name__icontains=...) | Q(author_name__icontains=...)`. This is what you are suggesting in point 3. In this example Qs are combined using `|` (or) due to `expression_connector` kwarg specified in the Field declaration. I have come hesitations on what the Field-level `expression_connector` default should be, but I'm slightly inclining towards `OR`.
    * For `popular` `FilterSchema` would search for `filter_<field_name>` method and invoke `filter_popular` to use the custom expression it returns. This should symantically align to what you do with `resolve_<field_name>` in `Schema`.

    Then, all Qs will be combined together using a class-level `expression_connector`.

    What do you think of it? If there are no objections, I will implement it shortly in the PR and fix the tests & docs accordingly.

vitalik:
Hi @l1b3r 
I'm planning to make release somewhere coming week - would you be able to make p2 ?

l1b3r:
Hi @vitalik,
I have implemented the following changes:

* When no `Field()` is specified for the field, the field's name is taken as a lookup
* "q" kwarg now supports not only a string, but a list of strings.
  * List items are combined using OR by default, but that can be overridden on a field basis.
* It is now possible to define a method `filter_<fieldname>` for a field, which must return a Q expression that will be used for that field.

Adjusted the tests and the docs accordingly. Could you launch the CI please?

l1b3r:
@vitalik, apologies, I have fixed the tests again. Didn't realize that Django added `XOR` connector for `Q` only in 4.1.


As a side note, I seem to be unable to run the matrix tests locally on a Windows + (WSL + Docker) setup due to a Docker build failure:
```
...\django-ninja\tests\env-matrix>docker-compose up --build

Step 14/31 : RUN /install_env.sh 3.6.10  2.1.15  env-36-21
 ---> Running in e49b1d95bb24
/bin/sh: 1: /install_env.sh: not found
The command '/bin/sh -c /install_env.sh 3.6.10  2.1.15  env-36-21' returned a non-zero code: 127
```
But I suppose that's out of scope of this PR.

vitalik:
@l1b3r  Thank you

l1b3r:
Thank you for giving this PR a look and thanks for the framework in general!

Having used both Django+DRF and FastAPI extensively and having struggled with both in one way or the other, I find Django-Ninja a very solid middle ground that takes best from both worlds. It's awesome how Ninja combines lightweight Pydantic-powered FastAPI-like handlers (instead of that bulky and hard-to-tweak DRF) with the rest of Django's goodies such as the ORM (something that FastAPI is missing completely), admin interface, ecosystem etc.

Looking forward to the new releases <3

Jragon:
I did something very similar for orderable fields by query param. Ends up looking like:

```python
class OrderableFields(OrderableFieldsEnum):
    amount = "amount", "order__amount_gross"
    date = "date", "start"

class FiltersQuery(Schema):
    ordering: Optional[OrderableFields | List[OrderableFields]]

@api.get(...)
def list_orders(request, filters: FiltersQuery = Query(...)):
    qs = ...
    qs = order_queryset(qs, filters.ordering)
    return qs
```

The `OrderableFieldsEnum` takes the items in the enum and adds a descending option for them too. So in the schema the accepted values are `amount, -amount, date, -date`. Could quite easily modify it to allow F() field descriptors for more complex orderings. 

Here's the enum def:
https://gist.github.com/Jragon/d7acabb05238f040fed86c44652c4658

Jragon:
I could tidy it up and add some docs if you thinks it's something worth adding? Seems relevant if we're adding a way to filter schemas.

vitalik:
@Jragon yeah, sure - if you see some issues with the docs feel free to add PRs

Issue #152, opened 2021-06-15
Swagger: Cannot use Enum as type in query params


Adamantish:
The following creates a functioning api just fine

```python
from enum import Enum

class Language(Enum):
    en = "en"
    de = "de"
    

@router.get("/availability", response={200: Availability, 400: ErrorMessage})
def availability(
    _: WSGIRequest,
    language_alpha_2: Language,
    other_param: str = None,
):
    ...
```
However, this is shown  at the top of the swagger docs
![Screenshot from 2021-06-15 15-42-31](https://user-images.githubusercontent.com/9261192/122064234-49913e80-cdf1-11eb-9a5f-940394a1e9ea.png)

Of course, this is no problem when the argument is instead listed in a pydantic schema. That would mean API breaking changes though.

Adamantish:
I just found that if I use this `Language` enum in one of the pydantic schemas it fixes the problem everywhere. So I guess the issue is that the Enums are turned into openapi schemas only when they're included in another schema. Is there a workaround here?

Adamantish:
Great. Tried it. Works. Thank you!

michaldyczko:
The same problem appears in the case of list of enums, like`List[Language]`. The same error is raised. It would be nice to have `<select multiple>` also.

vitalik:
@michaldyczko 

could you provide full example ? what is `Language` in your case ?

vvoody:
Everything works, the query parameter `protocol` is documented and shown as selectable values.
```python
class Proto(Enum):
    SSL2 = "SSL2"
    SSL3 = "SSL3"
    TLS1 = "TLS1"
    TLS1_1 = "TLS1_1"
    TLS1_2 = "TLS1_2"
    TLS1_3 = "TLS1_3"


class VersionsQueryParams(Schema):
    ipv4: IPv4Address
    port: conint(ge=0, le=2**16 - 1) = 443
    protocol: Optional[Proto] = None


@router.get(
    "/versions",
    summary="Get Supported SSL Versions on IP",
    operation_id="get_versions",
    response=list[SSLVersionSupported],
)
def get_versions(request, filters: VersionsQueryParams = Query(...)):
    logger.debug("protos: %s", filters.protocol)
    return {}
```
<img width="564" alt="image" src="https://user-images.githubusercontent.com/65963/167373199-3d3d32d3-91e5-4f6f-b2b0-304124447ea9.png">

---

Have some issues if uses `List[Enum]`:
```python
[...]

PR #679, opened 2023-02-10
fix List[enum] in query params

Jragon:
For #152 

At the moment the schema-to-openai converter doesn't correctly resolve a List[enum]. It assumes it's in the definitions - but is never added to the definitions unless it is used in a response schema.

I've fixed this by resolving the 'items' using the definitions in the schema: https://json-schema.org/understanding-json-schema/reference/array.html#items

Though I think a better way would be to just add the enum to the schema definitions but it didn't look like that's how you had done it before with the other enums so I emulated your method.


vitalik:
Merged, thank you

PR #687, opened 2023-02-23
Support both new and old style Union/Optional

### LINKED ISSUE #685: [BUG] Python 3.10: union compatibility (for payload)
**Describe the bug**
The behaviour introduced in #301 to handle `Union`-types, is no longer working on the new style union-type syntax (`|`).

```pycon
>>> from typing import Optional, Union, get_origin
>>> get_origin(Union[int, str])
typing.Union
>>> get_origin(int | str)
<class 'types.UnionType'>
>>> get_origin(Optional[int])
typing.Union
>>> get_origin(int | None)
<class 'types.UnionType'>

```

My proposed solution would be to loosen the test in `is_pydantic_model` to include the new type.
https://github.com/vitalik/django-ninja/blob/432ab5313bef53941441a44bf7882ba9656abd43/ninja/signature/details.py#L248-L254

 - Python version: 3.10.9
 - Django version: 4.1.7
 - Django-Ninja version: 0.20.0
 - Pydantic version: 1.10.5

gvangool:
The Python 3.10+ syntax of creating a union or optional type (with `str | int`), results in a different type than using the pre-Python 3.10 syntax (`Union[str, int]`).

```pycon
>>> from typing import Optional, Union, get_origin
>>> get_origin(Union[int, str])
typing.Union
>>> get_origin(int | str)
<class 'types.UnionType'>
>>> get_origin(Optional[int])
typing.Union
>>> get_origin(int | None)
<class 'types.UnionType'>
```

I'm curious about how to update the test so that it works on both Python 3.9 and Python 3.10.
If you can point me in the right direction, I'll gladly update it.

This would fix #685.

### DIFF
--- a/ninja/signature/details.py
+++ b/ninja/signature/details.py
@@ -24,6 +24,13 @@ FuncParam = namedtuple(
     "FuncParam", ["name", "alias", "source", "annotation", "is_collection"]
 )
 
+try:
+    from types import UnionType
+
+    UNION_TYPES = (Union, UnionType)
+except ImportError:
+    UNION_TYPES = (Union,)
+
 
 class ViewSignature:
     FLATTEN_PATH_SEP = (
@@ -247,7 +254,7 @@ class ViewSignature:
 
 def is_pydantic_model(cls: Any) -> bool:
     try:
-        if get_collection_origin(cls) == Union:
+        if get_collection_origin(cls) in UNION_TYPES:
             return any(issubclass(arg, pydantic.BaseModel) for arg in get_args(cls))
         return issubclass(cls, pydantic.BaseModel)
     except TypeError:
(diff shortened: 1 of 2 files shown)


OtherBarry:
@gvangool you should be able to only run the test if the version is 3.10+ using pytest skipif [(docs)](https://docs.pytest.org/en/7.1.x/how-to/skipping.html#id1).

You can then get a union value using eval:
```python
example_union = eval("str | int")
```

Or define a function with union types using exec:
```python
exec("""def example_func(something: int | str) -> int | str:\n    return something""")
```

These should both stop earlier python versions from complaining about the | operator. There might be a cleaner way to do it, but it's a difficult problem to google.

vitalik:
Hi @gvangool 
thank you for contribution, would you be able to add tests for this ?

gvangool:
@OtherBarry thanks for that pointer!
I realized after pushing that version that we can also use strings there.
So this version works on Python 3.10 and throws errors on 3.9:
```
TypeError: unsupported operand type(s) for |: 'ResolverMetaclass' and 'ResolverMetaclass'
```

OtherBarry:
@gvangool @vitalik might need to look into using some `pragma: no cover` comments in this PR.

I tried bumping the coverage CI job python version to 3.10, but naturally it just fails on the other side of the try-except.

Issue #693, opened 2023-03-06
[BUG] Content Security Policy: The page’s settings blocked the loading of a resource at inline (“script-src”).


andZibralabs:
**Describe the bug**
At /docs url, there is a CSP error:
Content Security Policy: The page’s settings blocked the loading of a resource at inline (“script-src”)

Looks like there is some inline script in templates/ninja/swagger.html

**Versions:**
 - Python version: 3.10.8
 - Django version: 4.1.7
 - Django-Ninja version: 0.21.0
 - Pydantic version: 1.10.4

vitalik:
@andZibralabs how to reproduce this ? which browser (browser version) do you use ?

andZibralabs:
@vitalik It is not about the browser, but about the Content-Security-Policy that has been set on the web server.
But the browser I was using is:
Mozilla Firefox 110.0.1
How to reproduce:
Set the Content-Security-Policy: "default-src 'none'; script-src 'self'; img-src 'self' data:; style-src 'self';"
go to url /docs

vitalik:
> Set the Content-Security-Policy

so this is not default browser behaviour? you just enforce it manually ?

I do not see any error messages:
<img width="1413" alt="SCR-20230309-opa" src="https://user-images.githubusercontent.com/95222/224077457-e4fa0e9d-4e2c-41ed-875f-051bae9f78ab.png">

tykling:
hello @vitalik :)

You can read about the `Content-Security-Policy` [here](https://content-security-policy.com/ )

It is an HTTP header you add on the webserver to instruct the webbrowser where to accept different types of content from. For example you can make a CSP header telling the browser to only accept images from `cdn.example.com` and js from `scripts.example.com` - making XSS attacks impossible because the browser now knows not to load javascript from `evilhacker.example.org` even if an XSS vuln should exist.

The problem with inlining scripts like the current state before #695 is that to the browser it looks exactly like a script injection. The only way to make it work with CSP is to effectively water down the CSP header so much that you might as well not have it.

TL;DR: Never use inline <script> tags, always load JS from external files. The same for images, don't use data: in <img> tags, load external images instead. It helps the browser keep the users safe.

tykling:
Hello again :)

Is there anything I can to to help get #695 merged? At the moment we have to set an unsafe CSP header on the webservers where we use django-ninja API docs. This is a shame since the fix is simple.

Thanks!

fretchen:
[...]

PR #695, opened 2023-03-06
Moved inline script from swagger.html to a js file.

andZibralabs:


baseplate-admin:
hi there ,, 

do you have any particular reasons for this change?

andZibralabs:
Yes I created an issue:
https://github.com/vitalik/django-ninja/issues/693
It is for CSP.

tykling:
This would be good to get in to deal with strict Content-Security-Policy headers. :+1:

PR #699, opened 2023-03-14
Cache

webdastur:
**"Django's cache framework"** added.

**"caching"** can be enabled by global NinjaAPI's **"cache"** parameter. Each HTTP method(GET, PUT, POST... or routes) has its own **"cache"** and **"cache_timeout"** parameters.

**Constants:**

**CACHE_TIMEOUT**: 1 second


webdastur:
@vitalik What do you think about this PR?

PR #700, opened 2023-03-14
"Django's cache framework" added.

webdastur:
"caching" can be enabled by global NinjaAPI's "cache" parameter. Each HTTP method(GET, PUT, POST... or routes) has its own "cache" and "cache_timeout" parameters.

Constants:

CACHE_TIMEOUT: 1 second

### DIFF
--- a/ninja/constants.py
+++ b/ninja/constants.py
@@ -15,3 +15,6 @@ class NOT_SET_TYPE:
 
 
 NOT_SET = NOT_SET_TYPE()
+
+# 1 second
+CACHE_TIMEOUT = 1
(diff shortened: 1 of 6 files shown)


webdastur:
@vitalik What do you this about this PR?

vitalik:
Hi @webdastur 

Thank you for your PR
But - have doubts that it should belong as part of the atributes passed to operation - as there can be more "candidates" (throttling, logging, etc) to be there and it will lead to hundreds of arguments

Personally I'm looking into this feature from the perspective of custom decorator (OR decorator that can be reused from django it self  like `cache_page`)

baseplate-admin:
> Personally I'm looking into this feature from the perspective of custom decorator (OR decorator that can be reused from django it self like cache_page)

I second this opinion. The main decorator can easily get incomprehensible after adding so many custom variables. A decorator that comes in after main api call seems more ergonomic.

PR #711, opened 2023-03-23
Add support for favicon and a custom redoc x-logo

chenatlas:
I find it hard to add the x-logo support for redoc/easily add a favicon url

Solution:
adding `favicon_url` and `x_logo_url` into `__init__` in `main.py`

and allow more attributes in the schema info section


vitalik:
Hi @chenatlas 

Thank you for PR, but unfortunately it cannot pass

 - You can achieve icon by adding 'ninja` to INSTALLED_APPS and override a template
 - There are plans to make docs engines pluggable (as there is already 4+ libraries that requested) 

so extending NinjaAPI arguments for docs does not make sense anymore

PR #735, opened 2023-04-07
Implement rudimentary support for async auth

kjagiello:
This PR introduces initial support for asynchronous authentication by allowing users to pass in async callables through the `auth` kwarg. The built-in authentication classes provided by this project remain synchronous-only.

Doesn't fully solve https://github.com/vitalik/django-ninja/issues/44, but it might be a good first step towards full async auth support.


kjagiello:
I've added some missing tests for CSRF and authentication errors. Note that I did not add any tests for CSRF exemption for async views, since that's [not yet supported](https://code.djangoproject.com/ticket/31949) in Django.

Issue #646, opened 2023-01-06
why use field.verbose_name.title() instead of field.verbose_name in orm/fields.py


adi1:
As the title says.... When we have lazy texts as verbose_names for django model fields, the schema generation will not use these in the correct language. Calling `field.verbose_name.title()` will evaluate it when the model is loaded and not when openapi.json is generated.

Changing `field.verbose_name.title()` to `field.verbose_name` resolves this issue, but I would like to know if this can cause other unexpected issues.

vitalik:
@adi1 

how do you generate openapi.json ? (or should I ask differently - how do I reproduce the issue ? )

adi1:
Sorry I'm not at the pc anymore so i will try my best to describe it on my phone. To reproduce, you need to import ugettext_lazy as _, then set your field verbose_name=_('some string'). I don't think i can describe the entire django i18n workflow.

This will use a proxy for the string, rather than the string itself. This proxy uses some funny thread local stuff (I think) which checks the current language, set at the beginning of every request.

To generate openapi.json, i simply go to /api/openapi.json . If my session (or if i have the right language cookie) is in German, the texts will be in German.

Currently, help_text is translated to german, verbose_name is english.

When you can .title() on that proxy, you're converting it to a string, without why language context, so it uses the original language of the text.

Let me know if you need more information, i can try to help better when I'm actually at the computer.

adi1:
I had some time to look at the code and tests, and the simplest solution would be something like this:

```diff
--- a/ninja/orm/fields.py
+++ b/ninja/orm/fields.py
@@ -15,6 +15,7 @@ from uuid import UUID
 
 from django.db.models import ManyToManyField
 from django.db.models.fields import Field
+from django.utils.functional import Promise
 from pydantic import IPvAnyAddress
 from pydantic.fields import FieldInfo, Undefined
 
@@ -141,7 +142,7 @@ def get_schema_field(field: Field, *, depth: int = 0) -> Tuple:
         default = Undefined
 
     description = field.help_text
-    title = field.verbose_name.title()
+    title = field.verbose_name if isinstance(field.verbose_name, Promise) else field.verbose_name.title()
 
     return (
         python_type,
@@ -173,6 +174,6 @@ def get_related_field_schema(field: Field, *, depth: int) -> Tuple[OpenAPISchema
         FieldInfo(
             default=default,
             description=field.help_text,
-            title=field.verbose_name.title(),
[...]

PR #748, opened 2023-04-21
Only title-case fields in schema if all lowercase verbose name

SmileyChris:
The more important purpose of this fix though, is that it changes the verbose_name to not be translated too early (see #646).

If the actual logical change isn't required, it's even simpler (but I think the smarter way  is more sensible).


OtherBarry (approved):

vitalik:
Great, Thank you

Issue #702, opened 2023-03-14
[BUG] Get operations to not respect Pydantic config options


scott-8:
**Describe the bug**
When using a Schema for GET parameters (as documented [here](https://django-ninja.rest-framework.com/guides/input/query-params/#using-schema), django-ninja ignores the `extra=Extra.forbid` declaration on the schema. The intended purpose of this configuration option is to cause validation to fail if extra parameters are provided. See [here](https://docs.pydantic.dev/usage/model_config/). This works when using POST requests, but not GET.

An example would be something like this:

```python

class TestSchema(Schema, extra=Extra.forbid):
    a: str

@router.get("/")
def get_endpoint(request, params: TestSchema= Query(...)):
    # succeeds when passing extra params
    return params

@router.post("/")
def post_endpoint(request, body: TestSchema):
    # fails when passing extra params
    return body
```

**Versions (please complete the following information):**
 - Python version: [e.g. 3.10]
 - Django version: [e.g. 4.1.7]
 - Django-Ninja version: [e.g. 0.19.1]
 - Pydantic version: [e.g. 1.10.5]

jkeyes:
This is due to `_map_data_paths` only including the query parameters that are included in `_flatten_map` for the param.

https://github.com/vitalik/django-ninja/blob/6ed724e24323647200f63593864c6074fc0827d5/ninja/params_models.py#L60-L72

I don't know enough about the internals to fix this properly (or even if it is desirable to do so), but I got the example above to work by modifying the method to add any `data` items that are not present in the `mapped_data` if the mapped data type is a `dict`.

```Python
    @classmethod
    def _map_data_paths(cls, data: DictStrAny) -> DictStrAny:
        flatten_map = getattr(cls, "_flatten_map", None)
        if not flatten_map:
            return data

        mapped_data: DictStrAny = NestedDict()
        for k in flatten_map:
            if k in data:
                cls._map_data_path(mapped_data, data[k], flatten_map[k])
            else:
                cls._map_data_path(mapped_data, None, flatten_map[k])

        # start of bodge #############
        for k in mapped_data:
            if isinstance(mapped_data[k], dict):
                for d in data:
                    if d not in mapped_data[k]:
                        mapped_data[k][d] = data[d]
        # end of bodge #############

        return mapped_data
```

For whatever it's worth, the addition of this [bodge](https://en.wiktionary.org/wiki/bodge) didn't break any existing tests.

jkeyes:
[...]

PR #749, opened 2023-04-21
fix: support Pydantic config options for GET requests

jkeyes:


baseplate-admin:
Hey nice PR,

Would you mind adding the options ( or anything ) to the documentation to make sure `django-ninja` supports `pydantic` config options?

jkeyes:
Thanks @baseplate-admin. The thing is, I'm not even sure this is a **good** fix. I don't yet know enough about Django Ninja internals to suggest it is. I included the code for this change in #702 but @vitalik didn't comment on it and then added the "help wanted" label, so I wasn't sure if the change I'd suggested was bad or that the lack of a test meant the code in the comment wasn't enough. I created the PR to ensure there was a test case in place, and then was hoping to get more feedback on the code itself to determine if it's good enough or not.

I can add a documentation piece to the PR for sure, but I was waiting to see if the code was good first.

The [Overriding Pydantic Config](https://django-ninja.rest-framework.com/guides/response/config-pydantic/) page doesn't refer to using keyword's in the class definition, so this feels like an entire section of documentation is needed rather than a paragraph explaining how to do use `extra`.

Issue #627, opened 2022-12-09
Add support for defining a custom parser or/and renderer on an endpoint instead of the whole API


kgeorgy:
**Is your feature request related to a problem? Please describe.**

It could be nice to be able to define custom renderer and parser on an endpoint basis: in my use case, in the same API, some endpoints (which semantically should be in the same API) needs to return the default API render format (json) and some needs to return a custom render format (msgpack). As far as I know, for now it is only possible to set custom renderer and parser on the NinjaAPI object, not on Routers or endpoints.

**Describe the solution you'd like**
Add the custom optional parameters "renderer" and "parser" on endpoints definition and also on routers. Endpoint custom renderer/parser override the router one, and the router one overrride the API one.

kgeorgy:
Hi @vitalik, thank's for the quick response. I've already splitted my API as a workaround. This is definitly not a "key blocker" issue, but could be a nice addition as parser and renderer are already supported at the API level. The question is "Do you think it could be a nice addation to Django Ninja?". If you or other people think that it could be interesting, I could potentialy work on a PR for that.

hossfard:
+1 for this feature request.

The choice of having parser, response content-type, and renderers all defined at the project level relies on the premise that the input/outputs are homogeneous throughout the project. That is not the case for my use-cases and so this *does* become a major issue for me.

Here are the possible workarounds and the issues with each:

  - **Option 1**: Set renderer based on `request.route` (proposed in #109). This violates separation of concerns and significantly reduces maintainability. Also the returned content-type would still not be adjustable in the generated docs, and this will not be able to handle custom input parser.
  - **Option 2**: Not relying on any renderers and instead returning a `django.http.HttpResponse` object. The issue with this is that it won't be able to handle custom parser on an endpoint, and the auto-generated doc pages would not reflect the response schema, response content-type, and the returned data will not be validated.
  - **Option 3**: Returning custom data-type and handling the rendering with a custom renderer (what is proposed in this in the second comment in this post). This is again a maintenance nightmare, will not allow for data validation, the documented return schema would be incorrect, content-type will not be correct, and there will be no data validation.
  - **Option 4**: Render based on request `Accept` headers. This violates separation of concerns for the renderer, and would work if only a single return type is available throughout. Also `Accept` header should not become a requirement on the client if the response data type is known (e.g. binary). Also, this approach fails if two different endpoints have different return types, but different renderers. Also the returned content-type would not be adjustable.
[...]

PR #754, opened 2023-04-27
Fix #674: Allow overriding renderer per endpoint

### LINKED ISSUE #674: Renderer should be configurable on a per-router or per-endpoint basis
**Is your feature request related to a problem? Please describe.**
I am trying to export metrics of my application via `prometheus_client`. Prometheus will consume plain-text data from any HTTP endpoint and parse it.
All my endpoints are JSON, except the metrics endpoint, which should be plaintext.

I'd like to be able to, ideally, set my renderer on a per-endpoint basis:

```python
@router.get("/", renderer=MyRenderer)
def endpoint(request):
    ...
```

Alternatively, I would be OK with having the router itself take a renderer:

```python
router = Router(renderer=MyRenderer)
```

What I have done now works, but seems like a hack, I bypass the renderer selection when bypassing `create_response` by returning an `HttpResponse`

```python
@router.get("/", response=str)
def endpoint(request):
    return HttpResponse("some plain text", status=200, content_type="text/plain")
```

DavidVentura:
This PR aims to address #674 and #627.
This is a minimum implementation that allowed me to validate the mechanism; before implementing further I'd like your opinion (@vitalik ?) on whether this feature would be OK to merge and whether this is the correct way to do so.

If you confirm, I'll add support for this feature on the other paths (patch/put/post/..) and will add some tests.

This PR allows for the following snippet to execute as expected:

```python
class PlainTextRenderer:
    media_type: str = "text/plain"
    charset: str = "utf-8"

    def render(self, request: HttpRequest, data: Any, *, response_status: int) -> Any:
        return str(data)

@api.get("/sub", renderer=PlainTextRenderer())
def sub(request, a: int, b: int):
    return f"some cool text; {a - b}\n"


@api.get("/add")
def add(request, a: int, b: int):
    return {"result": a + b}
```

### DIFF
--- a/ninja/main.py
+++ b/ninja/main.py
@@ -117,6 +117,7 @@ class NinjaAPI:
         url_name: Optional[str] = None,
         include_in_schema: bool = True,
         openapi_extra: Optional[Dict[str, Any]] = None,
+        renderer: Optional[BaseRenderer] = None,
     ) -> Callable[[TCallable], TCallable]:
         """
         `GET` operation. See <a href="../operations-parameters">operations
@@ -138,6 +139,7 @@ class NinjaAPI:
             url_name=url_name,
             include_in_schema=include_in_schema,
             openapi_extra=openapi_extra,
+            renderer=renderer,
         )
 
     def post(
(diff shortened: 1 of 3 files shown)


baseplate-admin:
I think we can just return `HttpResponse` from endpoints with correct headers :)

DavidVentura:
That is the workaround I have right now in my service. However, having to deal with the "lower level" features directly kinda defeats the point of using a framework

PR #755, opened 2023-04-27
Implement PlainTextRenderer

DavidVentura:
Having a `PlainTextRenderer` is useful; mostly to hook up a [prometheus_client](https://github.com/prometheus/client_python).

This PR is not particularly useful by itself, but it will make writing useful tests for #754 easier/better.


baseplate-admin:
Hi, There. Thanks for this PR. 

But i must ask : Why not just return `HttpResponse` ( with correct headers ) from `django` ?

PR #766, opened 2023-05-22
Make optional field of UploadedFile optional

karlosss:
`UploadedFile.__modify_schema__` has a `field` parameter that is not used in the method body, is typed as `Optional`, but still requires a value. Schema generation fails because of the value for this parameter not being provided.


karlosss:
thanks for the fast review and merge!

Issue #777, opened 2023-06-21
@paginate overrides `_ninja_contribute_to_operation`


asiffer:
First of all, I definitely love the `_ninja_contribute_to_operation` entrypoint. It allows to enhance the API operations (for example I could define permissions with this). First it seems a bit hacky but finally the pattern in very interesting. Maybe, in the future, providing some kind of helpers to use it would be great (I can help in that way).

I noticed that the `@paginate` decorator uses this hook. However, it does not take into account whether this hook is already defined, basically it overrides it:

```python
# ninja.pagination._inject_pagination
# ...
if paginator.Output:
    view_with_pagination._ninja_contribute_to_operation = partial(  # type: ignore
        make_response_paginated, paginator
    )
```

I think, it would be nice to define a function that "appends" these hooks.

OtherBarry:
I've fixed this over in #604

asiffer:
Oh I didn't look at the PRs! I noticed that you have worked on `_ninja_contribute_args` but not `_ninja_contribute_to_operation`, is that right?

OtherBarry:
Ahhh right, that one is a little more complex as it's not in a list, but it shouldn't be hard to implement.

Something like:
```python
if hasattr(view_func, "_ninja_contribute_to_operation"):
    # Allow 3rd party code to contribute to the operation behaviour
    for func in view_func._ninja_contribute_to_operation:
        func(self)
```

and then a similar function to my `inject_contribute_args` called `inject_contribution_to_operation` or something.

Happy to add to my todo list, but given my previous history it might be a few months before I get around to it, so if you want it sooner best to implement it yourself.

asiffer:
Ah I did not think about changing the type of `_ninja_contribute_to_operation`. To make my decorator work, either I define the method or I insert my function in the end:

```python
from functools import partial, wraps
from typing import Any, Callable

from django.http import HttpRequest, HttpResponse
from django.http.response import HttpResponseBase
from ninja.operation import Operation

Contribution = Callable[[Operation], None]

def post_hook(
    function: Contribution,
    post_function: Contribution,
):
    """This function adds a function call (post_function) at the end
    of the initial function ('function' parameter)
    """

    @wraps(function)
    def fun(op: Operation) -> None:
        function(op)
        post_function(op)

    return fun


def append_contribution(view_func: Callable, contribution: Contribution) -> Callable:
    method = "_ninja_contribute_to_operation"
    if hasattr(view_func, method):
[...]

PR #779, opened 2023-06-22
Refactor pagination to not overwrite _ninja_contribute_to_operation

OtherBarry:
Solves #777, using a similar solution to my other PR #604 

- Creates a function called inject_contribute_args that will add new args without overwriting existing ones
- Updates the _inject_pagination function to use the new function

No tests written yet, as #604 reorganises the relevant test structure. Leaving as draft until #604 is merged.

### DIFF
--- a/ninja/operation.py
+++ b/ninja/operation.py
@@ -92,7 +92,8 @@ class Operation:
 
         if hasattr(view_func, "_ninja_contribute_to_operation"):
             # Allow 3rd party code to contribute to the operation behaviour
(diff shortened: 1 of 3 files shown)


vitalik:
Thank you

this is now implemented in separate patch in upcoming release along with operation callback

https://github.com/vitalik/django-ninja/blob/e4c10ade3c59e162626e9291f4ca74a60ef9c553/ninja/utils.py#L53-L64

PR #785, opened 2023-07-08
authenticate a superuser

vpoulailleau:
from ninja.security import django_auth_superuser


vpoulailleau:
This is only a documentation update… (no checks should fail here)

vitalik:
Thank you

PR #788, opened 2023-07-12
Fix OpenAPI spec example in api-docs.md

pavel-beaufort:
Removes redundant curly bracket which makes it invalid.


vitalik:
thank you

Issue #782, opened 2023-06-27
A way to pass arguments to SwaggerUI (like `persistAuthorization`)


pawelngei:
**Is your feature request related to a problem? Please describe.**
Currently Django-Ninja's Swagger doesn't remember tokens between refresh, which can be pretty annoying when developing permission-heavy endpoints requiring refreshing the docs. As far as I'm aware, the `openapi_extra` doesn't pass any parameters (like https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/ ) to SwaggerUI itself.

**Describe the solution you'd like**
We could introduce a new parameter, like `swaggerui_params` (and maybe a matching `redoc_params`), which will be a dictionary of parameters passed to each UI (for example via [`swagger.html`](https://github.com/vitalik/django-ninja/blob/master/ninja/templates/ninja/swagger.html) or [`swagger-ui-init.js`](https://github.com/vitalik/django-ninja/blob/master/ninja/static/ninja/swagger-ui-init.js))

Alternatively, we could set `persistAuthorization` as a separate boolean option to `NinjaAPI`, as just `persist_doc_authorization=true`. This will be a simpler case, but might be limiting to the users in the long run.

Should we implement either of these, it would be good to put an example in the docs, especially https://django-ninja.rest-framework.com/guides/api-docs/ .

vitalik:
Hi @pawelngei 

yeah, some extensions would be nice... but would not make it a separate parameter like `NinjaAPI(swagger_params={})` - the idea for future is add ability to add way more OpenAPI doc viewers as plugins (swagger, redocs, rapidocs, etc)

I think most flexible approach would be to just add a context that will be passed to a html template that renders that doc engine.. like:

```Python

api = NinjaAPI(docs_context={'swagger': {'persistAuthorization': True}}}

```

pawelngei:
If I was to write the code (can't promise I'll have time right now), would it be enough to pass and test `docs_context` for `swagger` only, or should I test `redoc` as well? Should I update the docs, or leave it to you / other maintainers?

pawelngei:
Created a PR implementing both Swagger, Redoc and tests, hopefully it's good enough to get merged ;)

PR #806, opened 2023-07-30
Add `docs_context` to pass additional parameters to Swagger and Redoc

pawelngei:
Created a PR implementing https://github.com/vitalik/django-ninja/issues/782 .

- Added `docs_context` optional parameter to NinjaAPI, which passes additional parameters straight to Swagger or Redoc templates
- Implemented and tested (by hand) successful parameter consumption by Swagger, Swagger CDN, Redoc and Redoc CDN versions.
  - Also tested passing wrong configuration (Swagger to Redoc, Redoc to Swagger) and empty configuration.
  - In order to pass additional `options` to Redoc, we cannot use the simple `<redoc />` HTML shorthand, as per [this document](https://redocly.com/docs/redoc/deployment/html/).
  - Used Django's [`json_script`](https://docs.djangoproject.com/en/4.2/ref/templates/builtins/#json-script) to safely pass multiple options to the template.
- Created simple tests for coverage.
- More tests could be useful, but I'm not sure how to properly instantiate NinjaAPI and Django, needing to modify both the `api` object and override `NINJA_DOCS_VIEW`.
- Added documentation for the feature in `api-docs.md`.
- Ran all the linters and formatters.
- More renderers / engines / plugins can be easily added, as `extended_settings` now exists in the OpenAPI View Template.

If you'd like to test the changes locally, the best set of settings should be:
```Python
api = NinjaAPI(docs_context={"swagger": {"filter": "SthToFilter"}})
api = NinjaAPI(docs_context={"redoc": {"disableSearch": True}})
```
This way the changes will be visible immediately after loading the page, no need to debug it in the console.

### DIFF
--- a/docs/docs/guides/api-docs.md
+++ b/docs/docs/guides/api-docs.md
@@ -41,6 +41,15 @@ from django.contrib.admin.views.decorators import staff_member_required
 api = NinjaAPI(docs_decorator=staff_member_required)
 ```
 
+## Extending your docs engine configuration
+
+To provide some advanced configuration for your documentation engine, like setting [`persistAuthorization`](https://swagger.io/docs/open-source-tools/swagger-ui/usage/configuration/#persistAuthorization) in Swagger, use the `docs_context` argument with `swagger` and / or `redoc` key matching your `NINJA_DOCS_VIEW`:
+
+```Python
+api = NinjaAPI(docs_context={'swagger': {'persistAuthorization': True}}}
+```
+
+
 ## Extending OpenAPI Spec with custom attributes
 
 You can extend OpenAPI spec with custom attributes, for example to add `termsOfService`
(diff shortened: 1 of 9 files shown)


pawelngei:
@vitalik not meaning to push you to merge this immediately, are there any more tests / updates you'd like me to do for this to be viable?

I could really really use this in my daily work ;)

vitalik:
Hi @pawelngei 

sure thank you for PR - I'm still thinking/researching this - as docs viewer questions are popping up pretty often and I want to give pretty pluggable solution that should be east extendable


most probably the solution would be something like passing a Docs engine object to NinjaAPI class

```Python

api = NinjaAPI(docs=Swagger())

...

api = NinjaAPI(docs=Swagger(settings={'persistAuthorization': True}))

....

api = NinjaAPI(docs=Redocs(settings={'something': 'else'}))


....

api = NinjaAPI(docs=ThirdPartyPlugin())

```

pawelngei:
So, do you take PRs for that, or would you like to write such a system yourself? 

I could really use this for my daily workflow ;)

vitalik:
@pawelngei the latest implemntation is in master

you can change settings like this:

```Python

swagger = Swagger(settings={"persistAuthorization": True})
api = NinjaAPI(docs=swagger)

```

pawelngei:
Well, it looks like you implemented everything I wanted to do here! I tested it locally and it works well :) We can close the PR, I just can't wait until it gets released publicly!

Thank you once again!

PR #832, opened 2023-08-24
Include Enum to NinjaJSONEncoder

Kristinus:
This would allow proper usage of enums in path parameters
Currently if you want to have an enum in the path, you would do this.
```
enum_value: SomeEnumClass = Path(...),
```
Which works fine until the request path includes an enum value that does not exist.
ie.
```
class SomeEnumClass(enum.Enum):
    first: "first"
    second: "second"

requests.get("some_path/third/other")
```

This would return a 500 instead of an expected 422 due to crashing in the JSONEncoder.
Returning and error
```
TypeError: Object of type SomeEnumClass is not JSON serializable"
```

### DIFF
--- a/ninja/responses.py
+++ b/ninja/responses.py
@@ -1,5 +1,6 @@
 from ipaddress import IPv4Address, IPv6Address
 from typing import Any, FrozenSet
+from enum import Enum
 
 from django.core.serializers.json import DjangoJSONEncoder
 from django.http import JsonResponse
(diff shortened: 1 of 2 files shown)


vitalik:
Hi @Kristinus 
Thank you
could you also add a test case for this ?

Kristinus:
I have now added a test @vitalik

vitalik:
Thank you

Issue #385, opened 2022-03-08
Cursor pagination


asaff1:
**Is your feature request related to a problem? Please describe.**
The current way to implement pagination is based on limit offset. I prefer the cursor style pagination.

**Describe the solution you'd like**
Standard way to implement cursor style pagination, similar to DRF one:
https://www.django-rest-framework.org/api-guide/pagination/#cursorpagination

benjaoming:
This would be really nice. FastAPI Pagination only has cursor support for its sqlalchemy and casandra backend, so here's a good chance to have better features :)

benjaoming:
Also relevant: https://github.com/photocrowd/django-cursor-pagination

PR #836, opened 2023-08-28
Add cursor pagination implementation

zachmullen:
This is a direct port of DRF's [CursorPagination](https://www.django-rest-framework.org/api-guide/pagination/#cursorpagination) to django-ninja. It is designed to be completely API- and behaviorally-compatible with the DRF version; even opaque cursor URLs generated with DRF should still work under this implementation, which was a design goal as the downstream project I initially created this on was switching from DRF to ninja.

If the ninja maintainers approve of the general concept, I can add tests and whatever else is required to get this merged.


vitalik:
Hi @zachmullen 

that's a good start, but would be nice to add some test coverage and fix failing tests on old python versions

ninja/pagination.py
```diff
@@ -295,15 +297,19 @@ def decode_cursor(cls, encoded_cursor: str | None) -> Cursor:
 
                 position = tokens.get("p", [None])[0]
             except (TypeError, ValueError):
-                raise ValueError(_("Invalid cursor."))
+                raise ValueError(_("Invalid cursor.")) from None
```
baseplate-admin, line 300: is it necessary to raise from None?

baseplate-admin: shouldn't we raise from TypeError or ValueError?

baseplate-admin:
```python
 except (TypeError, ValueError) as e:
                raise ValueError(_("Invalid cursor.")) from e
```
                
```

zachmullen: I suppose `__cause__` is semantically correct, though the underlying exception will be in `__context__` regardless. Let me know which is preferred.

zachmullen:
I've got the tests passing and added some of my own, but `mypy` is unhappy, and the coverage isn't at 100% and I don't feel comfortable inline-disabling it. This is all the work I will be able to do for a while, so let me know if you want to take over this branch, or else I can release this paginator as its own third party package.

pranto157:
Any idea when this will merge ?

jamesrkiger:
@vitalik I would like to help get this MR up to your standards so it can be merged. From some local testing the basic functionality seems to work fine. So it seems like the main issue will be getting more testing coverage. Before I proceed, however, I was wondering if you could tell me whether the current testing for cursor pagination in this branch is up to your standards. It adds endpoints to the `demo_project ` and runs tests through there. Looking at other testing in the main repo, however, I got the impression you might want something a bit more abstracted. Very few tests for other parts of django-ninja use the demo_project. Should I build on the current cursor pagination testing here or try to overhaul it before building out more coverage? Any input would be appreciated.

bufke:
I noticed that the existing pagination works with both querysets and lists. See [existing pagination](https://github.com/vitalik/django-ninja/blob/master/ninja/pagination.py#L53). This one does not support lists. It looks like it would be have be more careful around using .count and perhaps avoid attempting to order lists.

vitalik:
Hi @zachmullen 

sorry for delay, we just got async pagination support so this PR got a bit spoiled with conflicts

zachmullen:
Thanks for the update @vitalk . What needs to be done to get this merged?

zachmullen:
Hi @vitalik , is this something you intend to get merged into django-ninja, or would it make more sense for me to publish a separate package for this feature?

danlamanna:
Cross linking #385

bufke:
Just FYI - I've been [using cursor pagination](https://gitlab.com/glitchtip/glitchtip-backend/-/blob/master/glitchtip/api/cursor_pagination.py?ref_type=heads) for 6 months now based on this PR.

zachmullen:
I have published this capability as a supplementary package: https://pypi.org/project/django-ninja-cursor-pagination/

I'll close this issue for now, but the maintainers should feel free to let me know if they'd like to upstream this into django-ninja and I'd be happy to help.

(FYI @bufke )

zachmullen:
I'm reopening this in case the maintainers want to keep this record. Feel free to close it.

vitalik:
done in 1.6

PR #838, opened 2023-08-30
#822 "ValidationError" and "HttpError" are not unpicklable

### LINKED ISSUE #822: [BUG] "ValidationError" and "HttpError" are not unpickable
**Describe the bug**

Hi.

It seems the `ValidationError` can be pickled but not un-pickled:

```python
import pickle
from ninja.errors import ValidationError

error = ValidationError(errors=[])

serialized = pickle.dumps(error)  # OK
deserialized = pickle.loads(serialized)  # NOK
```

It causes:

```
  File "/home/delgan/test.py", line 6, in <module>
    deserialized = pickle.loads(serialized)
                   ^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: ValidationError.__init__() missing 1 required positional argument: 'errors'
```

This is actually a reccurent problem while inheriting Python `Exception`:
- [Cannot unpickle Exception subclass](https://stackoverflow.com/questions/41808912/cannot-unpickle-exception-subclass)
- [How to pickle inherited exceptions?](https://stackoverflow.com/questions/49715881/how-to-pickle-inherited-exceptions)
-  [How to make a custom exception class with multiple init args pickleable](https://stackoverflow.com/questions/16244923/how-to-make-a-custom-exception-class-with-multiple-init-args-pickleable)

The fix requires to implement a `__reduce__` method for these classes.

The `multiprocessing` module makes extensive use of `pickle` protocol, and having objects that can't be de-serialized is source of unexpected errors (in my case, because the `Exception` is sent to a background processing thread for logging).

 - Python version: 3.11.3
 - Django version: 4.2.4
 - Django-Ninja version: 0.22.2
 - Pydantic version: 0.4.4

AlTosterino:
I'm using[ `loguru` ](https://github.com/Delgan/loguru) for logging and had hard time debugging `TypeError: ValidationError.__init__() missing 1 required positional argument: 'errors'` while calling `logger.exception` in Ninja `api.on_exception`. This is addressing the issue.

Fixes #822

### DIFF
--- a/ninja/errors.py
+++ b/ninja/errors.py
@@ -39,14 +39,18 @@ class ValidationError(Exception):
     """
 
     def __init__(self, errors: List[DictStrAny]) -> None:
-        super().__init__()
         self.errors = errors
+        super().__init__(errors)
 
 
 class HttpError(Exception):
     def __init__(self, status_code: int, message: str) -> None:
         self.status_code = status_code
-        super().__init__(message)
+        self.message = message
+        super().__init__(status_code, message)
+
+    def __str__(self) -> str:
+        return self.message
 
 
 def set_default_exc_handlers(api: "NinjaAPI") -> None:
--- a/tests/test_errors.py
+++ b/tests/test_errors.py
@@ -1,0 +1,26 @@
+import pickle
+
+from ninja.errors import HttpError, ValidationError
+
+
+def test_validation_error_is_picklable_and_unpicklable():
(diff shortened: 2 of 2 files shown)


OtherBarry:
Minor nitpick, but it should be `picklable` and `unpicklable`, with an L, according to the python documentation.

AlTosterino:
@OtherBarry Oh my I didn't even notice... Thanks!

I also deleted `__reduce__` method as `super().__init__(...)` did the trick 👌

Issue #835, opened 2023-08-28
[BUG] Django Ninja does not respect Schema config extra options


nofalx:
**Describe the bug**

**Versions (please complete the following information):**
 - Python version: 3.10
 - Django version: 4.24
 - Django-Ninja version: v1.0a3
 - Pydantic version: 2.3.0

I have the below structure and when trying to use the option `extra` it doesnt work. This is unlike the pre v1 versions
```

class UserTokenType(Schema):
    linking_id: LinkingId
    queue_reference_id: UUID | None = None
    meta_data: MetaData
    cp_transaction_entry: UUID

    class Config:
        extra = "forbid"
        
---------------


@router.post(
    "token/",
    response={201: dict, 400: dict, 403: dict, 404: ErrorResponse},
)
def survey_user_token_add(request: HttpRequest, user_token: UserTokenType):
   ....
```

vitalik:
@nofalx 

does it work if you pass it [like in v2 docs ](https://docs.pydantic.dev/latest/usage/model_config/#extra-attributes)?

```Python
class UserTokenType(Schema):
    ...
    model_config = ConfigDict(extra='allow')
```

nofalx:
@vitalik  I tried that approach as well and it did *not* work as well

 if keep both like below I get an error that i cant use both at same time, so im guessing both approach are supported and they both dont work even when used alone

```
class UserTokenType(Schema):
    model_config = ConfigDict(extra='forbid')
    
    class Config:
        extra = "forbid"
```

PR #845, opened 2023-09-07
Fix issue for extra fields not appearing in the errors

nofalx:
Solves Issue #835

Reference for impact of this change:
https://docs.pydantic.dev/latest/usage/validators/#model-validators

### DIFF
--- a/ninja/signature/details.py
+++ b/ninja/signature/details.py
@@ -159,6 +159,7 @@ class ViewSignature:
             )
 
             base_cls = param_cls._model
+            base_cls.model_config["from_attributes"] = True
             model_cls = type(cls_name, (base_cls,), attrs)
             # TODO: https://pydantic-docs.helpmanual.io/usage/models/#dynamic-model-creation - check if anything special in create_model method that I did not use
             result.append(model_cls)
--- a/ninja/signature/details.py
+++ b/ninja/signature/details.py
@@ -159,7 +159,6 @@ class ViewSignature:
             )
 
             base_cls = param_cls._model
-            base_cls.model_config["from_attributes"] = True
             model_cls = type(cls_name, (base_cls,), attrs)
             # TODO: https://pydantic-docs.helpmanual.io/usage/models/#dynamic-model-creation - check if anything special in create_model method that I did not use
             result.append(model_cls)
--- a/ninja/operation.py
+++ b/ninja/operation.py
@@ -1,16 +1,5 @@
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Callable,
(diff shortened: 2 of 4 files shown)


nofalx:
This PR actually fixes another issue as well:

when u have Schema as attribute for another schema:
```
class UserTokenType(Schema):
    meta_data: MetaDataSchema | None = None
```
and you send this json
```
{
    "meta_data": "xxx",
}
```
Django ninja reports that schema is correct, however after the change in this pr we get the correct error that we missed before
```
{
    "error": "validation_errors",
    "details": [
        {
            "field_name": "meta_data",
            "msg": "Input should be a valid dictionary or object to extract fields from",
            "type": "model_attributes_type"
        }
    ]
```

vitalik:
well it's not that easy (as you see from failed tests) 

I think there must be some trick so that pydantic thinks DjangoGetter is a dict

vitalik:
Hi @nofalx 

regarding - `from_orm` ? did you find this somewhere on pydantic code ? I think they deprecated this keyword in favour to "from_attributes"...

baseplate-admin:
Hi, Please use black on your code

nofalx:
@vitalik  turns out the logic does not depend on the `from_attributes`  but more on if the source of validate if from from `model_validate` or somewhere else. 

I dont understand the inners of pydantic very well but I have tested the update code in a django projects and all scenarios seem to be working fine for root of the json and nested json shema we using for the API. The projects tests are all passing as well

nofalx:
Added new test cases to cover the issues this pr fixed.  Code now passes lint and tests

ninja/signature/details.py
```diff
@@ -159,6 +159,7 @@ def _create_models(self) -> TModels:
             )
 
             base_cls = param_cls._model
+            base_cls.model_config["from_attributes"] = True
```
vitalik, line 162: what's the logic for "from_attributes" here ? the signature module kinda works only on anayling function signatures...

nofalx: You are right, it was needed when I based my logic on `from_attributes` but its not valid anymore. Reverting

vitalik:
There might be also development on pydantic side for this: https://github.com/pydantic/pydantic/issues/5434 will check if that covers this case as well

vitalik:
Hi @nofalx 

I simplified a bit your approach (used class.model_config to find out if default validaion wrapper is needed) - see here for more details - #863 - let me know if that works ?

nofalx:
Hi @vitalik 

Looks like a better approach, hope this makes it to next beta

PR #852, opened 2023-09-13
Add a test for list query params

marius-mather:
I had a bit of trouble figuring out list query parameters - possibly they could be a bit better documented in terms of needing to set them up with `Query(...)`?

While working this out I implemented a quick test for django-ninja's handling of list params, hopefully this is useful.


vitalik:
Thank you

Issue #764, opened 2023-05-12
Can `ModelSchema` use django `IntegerRangeField`?


ohyunkyo:
*This is just simplified source code.

I try to use django `IntegerRangeField`. but when I get queryset with `ModelSchema`, `KeyError: 'IntegerRangeField'` has occurred. 

```
model.py

class MotorVehicle(BaseModel):
    ...
    price = IntegerRangeField(null=True, blank=True)
    ...
```

```
schema.py

class MotorVehicleOutSchema(ModelSchema):
    class Config:
        model = MotorVehicle
        model_exclude = ["id"]
```

```
api.py

@router.get("/motor-vehicles", response={200: List[MotorVehicleOutSchema]})
def get_motor_vehicles(request):

    motor_vehicles = MotorVehicle.objects.all()

    return motor_vehicles
```

```
error

Traceback (most recent call last):
  File "/Users/ohyunkyo/.pyenv/versions/3.11.1/lib/python3.11/threading.py", line 1038, in _bootstrap_inner
    self.run()
  File "/Users/ohyunkyo/.pyenv/versions/3.11.1/lib/python3.11/threading.py", line 975, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/ohyunkyo/.local/share/virtualenvs/oh-backend-v1-Vo_OQdRI/lib/python3.11/site-packages/django/utils/autoreload.py", line 64, in wrapper
    fn(*args, **kwargs)
  File "/Users/ohyunkyo/.local/share/virtualenvs/oh-backend-v1-Vo_OQdRI/lib/python3.11/site-packages/django/core/management/commands/runserver.py", line 133, in inner_run
    self.check(display_num_errors=True)
  File "/Users/ohyunkyo/.local/share/virtualenvs/oh-backend-v1-Vo_OQdRI/lib/python3.11/site-packages/django/core/management/base.py", line 485, in check
    all_issues = checks.run_checks(
                 ^^^^^^^^^^^^^^^^^^
  File "/Users/ohyunkyo/.local/share/virtualenvs/oh-backend-v1-Vo_OQdRI/lib/python3.11/site-packages/django/core/checks/registry.py", line 88, in run_checks
    new_errors = check(app_configs=app_configs, databases=databases)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #694, opened 2023-03-06
Error in generating schema for GeoDjango models


gurbaaz27:
**Please describe what you are trying to achieve**
I have a Django model with a location field (which is a PointField), and I am trying to generate schema from it using ModelSchema, but I guess django-ninja doesn't have support for these geofields. I wanted to know if I am doing an error, or is there any hack around the same? Thanks.

**Please include code examples (like models code, schemes code, view function) to help understand the issue**

models.py

```python
from django.db import models
from django.contrib.gis.db import models as geomodels
from django.utils.translation import gettext_lazy as _

from uuid import uuid4

# Create your models here.
class Store(models.Model):
    id = models.UUIDField(_("ID"), primary_key=True, editable=False, default=uuid4) 
    name = models.CharField(_("Name"), max_length=200)
    location = geomodels.PointField(_("Location"), geography=True, srid=4326, blank=True, null=True)
    description = models.TextField(_("Description"), null=True, blank=True)
    contact_info = models.JSONField(_("Contact Information"), null=True)
```

schema.py

```python
from ninja import ModelSchema
from stores.models import Store

class StoreSchema(ModelSchema):
    class Config:
        model = Store
        models_fields = "__all__"
```

Error

```
Watching for file changes with StatReloader
Performing system checks...

Exception in thread django-main-thread:
Traceback (most recent call last):
  File "/usr/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/home/gurbaaz/servlog-backend/venv/lib/python3.10/site-packages/django/utils/autoreload.py", line 64, in wrapper
    fn(*args, **kwargs)
  File "/home/gurbaaz/servlog-backend/venv/lib/python3.10/site-packages/django/core/management/commands/runserver.py", line 134, in inner_run
    self.check(display_num_errors=True)
  File "/home/gurbaaz/servlog-backend/venv/lib/python3.10/site-packages/django/core/management/base.py", line 475, in check
    all_issues = checks.run_checks(
  File "/home/gurbaaz/servlog-backend/venv/lib/python3.10/site-packages/django/core/checks/registry.py", line 88, in run_checks
    new_errors = check(app_configs=app_configs, databases=databases)
  File "/home/gurbaaz/servlog-backend/venv/lib/python3.10/site-packages/django/core/checks/urls.py", line 14, in check_url_config
[...]

Issue #335, opened 2022-01-25
Django Ninja + GeoDjango


Jonesus:
Currently using GeoDjango's custom fields breaks Django ninja with the following error:

```
Exception in thread django-main-thread:
...
< Omitted for brevity >
...
File "/opt/pysetup/.venv/lib/python3.9/site-packages/ninja/orm/fields.py", line 121, in get_schema_field
   python_type = TYPES[internal_type]
KeyError: 'PointField'
```

Minimal reproduction models:

`models.py`
```
from django.contrib.gis.db import models
from ninja import ModelSchema

class Restaurant(models.Model):
    location = models.PointField()
    
class RestaurantOut(ModelSchema):
    class Config:
        model = Restaurant
        model_fields = ["location"]
```

and urls:

`urls.py`
```
from django.urls import path
from ninja import NinjaAPI
from .models import Restaurant, RestaurantOut

api = NinjaAPI()

@api.get("/restaurants", response=list[RestaurantOut])
def list_restaurants(request):
    return Restaurant.objects.all()

urlpatterns = [path("api/", api.urls)]
```

Is GeoDjango support something that could be considered to be included in the project? Would using another package like [geojson-pydantic](https://github.com/developmentseed/geojson-pydantic) help?

Jonesus:
I managed to work around this issue for now by referring to https://github.com/vitalik/django-ninja/issues/53, and using a custom property on the model:

```python
import json
from django.contrib.gis.db import models
from geojson_pydantic import Point
from ninja import ModelSchema


class Restaurant(models.Model):
    name = models.CharField()
    location = models.PointField()

    @property
    def location_geometry(self):
        return json.loads(self.location.json)

    
class RestaurantOut(ModelSchema):
    location_geometry: Point

    class Config:
        model = Restaurant
        model_fields = ["name"]
```

However, it would be nice to have proper support for the GeoDjango functionalities: I tried some hacking around `django-ninja` source, and by changing `/ninja/orm/fields.py`:

```python
from geojson_pydantic import Point 

...

TYPES = {
    ...
    "PointField": Point
}

```
I was first able to get the OpenAPI docs to work properly, and then by changing `/ninja/schema.py`:
```python
import json
from django.contrib.gis.geos import Point

...

class DjangoGetter(GetterDict):
    def get(self, key: Any, default: Any = None) -> Any:
        ...
        elif isinstance(result, Point):
            return json.loads(result.json)
```
[...]

PR #873, opened 2023-10-11
Improves create_schema to get types from custom_fields

lucasrcezimbra:
Related to #335, #694, #764

Removes the need to update TYPES for fields passed as custom_fields.

Example:

```python
class CustomField(models.fields.Field):
    pass

class TestModel(models.Model):
    custom = CustomField()

# before
CustomType = TypeVar("CustomType")
TYPES.update({'CustomField': CustomType})
Schema1 = create_schema(TestModel, custom_fields=[("custom", CustomType, ...)])

# with this PR
CustomType = TypeVar("CustomType")
Schema1 = create_schema(TestModel, custom_fields=[("custom", CustomType, ...)])
```

### DIFF
--- a/ninja/orm/factory.py
+++ b/ninja/orm/factory.py
@@ -66,12 +66,15 @@ class SchemaFactory:
         for fld in model_fields_list:
             if fld.name in custom_field_names:
                 continue
-            python_type, field_info = get_schema_field(
-                fld,
-                depth=depth,
-                optional=optional_fields and (fld.name in optional_fields),
-            )
-            definitions[fld.name] = (python_type, field_info)
+            try:
+                python_type, field_info = get_schema_field(
+                    fld,
+                    depth=depth,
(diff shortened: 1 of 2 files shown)


vitalik:
Hi @lucasrcezimbra 

Thank you for PR


Maybe it's also make sense to add some builtin method like "register_field_type("CustomField", MyPydanticType) that will update TYPES

and then error message should suggest to use this function ?

lucasrcezimbra:
> Maybe it's also make sense to add some builtin method like "register_field_type("CustomField", MyPydanticType) that will update TYPES
> 
> and then error message should suggest to use this function ?

Makes sense. I'll add it to this PR.

vitalik:
This is already implemented via register_field:

```python
from ninja.orm import create_schema, register_field

# Register your custom Django field type
register_field("VectorField", list[float])

# Now create_schema will know how to handle it
Schema = create_schema(MyModel)
```

PR #879, opened 2023-10-17
docs: use lowercase for language identifier

ivoire:


ivoire:
Lower cased as discussed in #878.

vitalik:
Thank you

PR #916, opened 2023-11-10
Suppress async-auth Warning "never awaited"

skokado:
Address for a Warning during async-auth request.
detail is here https://github.com/vitalik/django-ninja/issues/44#issuecomment-1805733011

### DIFF
--- a/ninja/operation.py
+++ b/ninja/operation.py
@@ -2,6 +2,7 @@ from typing import (
     TYPE_CHECKING,
     Any,
     Callable,
+    Coroutine,
     Dict,
     Iterable,
     List,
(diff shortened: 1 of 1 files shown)


ninja/operation.py
```diff
@@ -305,9 +306,11 @@ async def _run_authentication(self, request: HttpRequest) -> Optional[HttpRespon
         for callback in self.auth_callbacks:
             try:
                 if is_async_callable(callback) or getattr(callback, "is_async", False):
-                    result = callback(request)
-                    if result is not None:
-                        result = await callback(request)
+                    cor: Optional[Coroutine] = callback(request)
+                    if cor is None:
+                        result = None
```
skokado, line 311: To avoid `UnboundVariable` at L319

Issue #619, opened 2022-12-05
[Question]  django ATOMIC_REQUESTS setting is not working


Chanmoro:
Hello.
I'm trying to django's  ATOMIC_REQUESTS setting with django ninja, but seems does not work.

On the other hand, it works fine using django.db.transaction as decorator instead of setting.

```python
@router.delete("/items/{item_id}")
@transaction.atomic
def delete_item(request: HttpRequest, item_id: int):
    ...
```

How can I use ATOMIC_REQUESTS with django ninja?

vitalik:
@Chanmoro 

django-ninja should not influence ATOMIC_REQUESTS behavour..

maybe you setting it in the wrong palce ? keep in mind that ATOMIC_REQUESTS is part of DATABASES section not a standalone section

Chanmoro:
@vitalik Thank you very much for your reply.

I have tested some behavior ATOMIC_REQUESTS and transaction.atomic decorator.
ATOMIC_REQUESTS setting still doesn't seem to work.

Reproduction code is in this repository.
https://github.com/Chanmoro/django-ninja-619-reproduction

Could you confirm these result?

## Setting

Django's database setting is as below.

```python
DATABASES = {
    'default': {
        'ENGINE': "django.db.backends.postgresql",
        'HOST': os.getenv('DB_HOST'),
        'PORT': os.getenv('DB_PORT'),
        'USER': os.getenv('DB_USER'),
        'PASSWORD': os.getenv('DB_PASSWORD'),
        'NAME': os.getenv('DB_DATABASE'),
        'ATOMIC_REQUESTS': True,
    }
}
```

And output sql log on PostgreSQL using `log_min_duration_statement` setting.

## case1: Test ATOMIC_REQUESTS

Transaction is started before insert, but it has not rollback after exception.
It's not expected behavior.


- API code
```python
@api.post("/test_atomic_requests")
def test_atomic_requests(request):
    Item.objects.create(name="test_atomic_requests")
    raise Exception()
```

- HTTP Request
`$ curl -XPOST localhost:8000/api/test_atomic_requests`

- SQL log
```
django-ninja-619-reproduction-db-1  | 2022-12-13 15:18:25.337 GMT [96] LOG:  duration: 0.254 ms  statement: SET TIME ZONE 'UTC'
django-ninja-619-reproduction-db-1  | 2022-12-13 15:18:25.337 GMT [96] LOG:  duration: 0.019 ms  statement: BEGIN
[...]

PR #924, opened 2023-11-15
Add documentation for debug error handling

Lunrtick:
This is a first attempt at documenting the error handling with DEBUG on or off, as discussed in https://github.com/vitalik/django-ninja/issues/619

### DIFF
--- a/docs/docs/guides/errors.md
+++ b/docs/docs/guides/errors.md
@@ -2,6 +2,14 @@
 
 **Django Ninja** allows you to install custom exception handlers to deal with how you return responses when errors or handled exceptions occur.
 
+## Debug behaviour
+
+It is important to note that when you've set `DEBUG=True` in your Django project, **Django Ninja** will swallow exceptions and return a 500 status error along with a stack trace).
+
+!!! warning "Transactions - `ATOMIC_REQUESTS` and `transaction.atomic`"
+
+    Due to the behaviour described above (when `DEBUG=True`), you may encounter issues with `transaction.atomic` or the `ATOMIC_REQUESTS` setting not working as expected - that's because a successful response is returned (status 500 with a stack trace).
+
 ## Custom exception handlers
 
 Let's say you are making API that depends on some external service that is designed to be unavailable at some moments. Instead of throwing default 500 error upon exception - you can handle the error and give some friendly response back to the client (to come back later)


vitalik:
@Lunrtick thank you

I was actually thinking to add extra check - do not shallow exception if ATOMIC_REQUESTS is set (which I think is not for 95% users)

Lunrtick:
That could definitely work. It would also mean that the user wouldn't get a stack trace back.

They could of course look in the terminal to see the error, so I suppose it doesn't matter too much either way.

Thinking about it now, maybe a Django middleware that comes as an optional extra with django-ninja would be the right way to go. You could keep lines 107 and 108 here and remove lines 109-112. 
https://github.com/vitalik/django-ninja/blob/e5e3a4c34afd136459dc00d6d2240604d3022bb5/ninja/errors.py#L104-L112

Instead, the logic could be
```python
if not settings.DEBUG: 
else:
    raise NinjaAPIException() from exc 
```
Or even just
```python
raise NinjaAPIException() from exc 
```
and then the middleware would be something simple like

```python
# catch NinjaAPIException
if settings.DEBUG:
    return 500, {"a nice": "stack trace"}
else:
    raise exc
```

This would mean that `ATOMIC_REQUESTS` should work as expected either way.

thawancomt:
This behavior definitevily need changes, u know why? Data integrity, currently if we get an erro on the response cycle, da Database dont rollback cause no error get  the middleware...

so erro 500 and user saved? yes this is what is current happening.

My propose.


Current the default behavior is to set this handler:

that can be fixed by

```python

 def _default_exception( 
     request: HttpRequest, exc: Exception, api: "NinjaAPI" 
 ) -> HttpResponse: 
     if not settings.DEBUG: 

     transaction.set_rollback(True) # Just add this line of code solve the entire problem
     logger.exception(exc) 
     tb = traceback.format_exc() 
     return HttpResponse(tb, status=500, content_type="text/plain") 

```

But the above code still getting happeing if user set a custom handler and dont handle the scenerio where a response cycle got an exception with is uncommon since return typacally isnt covered by transaction... maybe this last part didnt make sense, but u get the idea...

In a nutshell the final fix in on the mains,py. line 614

```python

    def on_exception(self, request: HttpRequest, exc: Exc[_E]) -> HttpResponse:
        handler = self._lookup_exception_handler(exc)
        transaction.set_rollback(True) # ensure in exception rollback is called
        if handler is None:
            raise exc
        return handler(request, exc)
```

PR #925, opened 2023-11-16
Fix links on `Motivation` page of Docs.

vincentdavis:
These seem to be old links, I think I got the new URLs correct.


vitalik:
Thank yoyu

Issue #934, opened 2023-11-17
[BUG] extra="forbid" does not work anymore with Django-ninja 1.0


ddahan:
**Describe the bug**
After migrating to Django Ninja 1.0+, and replacing `Config` with `Meta`, the `extra = "forbid"` directive seems to have no effect.

Here is a code example that does not work with Django Ninja 1.0:
```
class BadgeSchemaInUpdate(ModelSchema):
    class Meta:
        model = Badge
        fields = ("expiration", "is_active")
        fields_optional = ("expiration", "is_active")
        extra = "forbid"
```

Here, I can update my `Badge` model using another attribute than `expiration` or `is_active` (not expected).
It did not happened before with this code and older Django Ninja 0.22.2:

```
class BadgeSchemaInUpdate(ModelSchema):
    class Config:
        model = Badge
        model_fields = ("expiration", "is_active")
        model_fields_optional = ("expiration", "is_active")
        extra = "forbid"
```

**Versions**
 - Python version: 3.12
 - Django version:  4.2.7
 - Django-Ninja version: 1.0.1
 - Pydantic version: 2.5.1

vitalik:
@ddahan yeah.. well the thing is that pydantic now asks to configure this using attribute [model_config](https://docs.pydantic.dev/latest/api/base_model/#pydantic.main.BaseModel.model_config)

so for now do it like `model_config = {'extra': 'forbid'}

but I guess that is a bit confusing problem - I will think about it

ddahan:
Thanks vitalik. This works:

```python
class BadgeSchemaInUpdate(ModelSchema):
    class Meta:
        model = Badge
        fields = ("expiration", "is_active")
        fields_optional = ("expiration", "is_active")

    model_config = {"extra": "forbid"}
```

And this works too (I suppose it does the same behind the scene):

```python
class BadgeSchemaInUpdate(ModelSchema):
    class Meta:
        model = Badge
        fields = ("expiration", "is_active")
        fields_optional = ("expiration", "is_active")

    class Config(Schema.Config):
        extra = "forbid"
```

As you said, it's a little confusing, but IMO the main issue is that it's not documented right now.
I can see at least [here](https://django-ninja.dev/tutorial/other/crud/) a usage of the former usage.

I suggest not closing the issue, at least until the documentation is updated.
Thanks for your work.

vitalik:
[...]

PR #945, opened 2023-11-21
Update pydantic model Config classes

adamchainz:
Update `Schema` class and subclasses for the [pydantic config changes](https://docs.pydantic.dev/2.5/migration/#changes-to-config) from its version 2.

I stopped part way since I realized the changes affect the API. Do you think using `model_config` makes sense for subclasses, or would you go with another design like `ModelSchema.Meta`?

### DIFF
--- a/ninja/conf.py
+++ b/ninja/conf.py
@@ -1,5 +1,5 @@
 from django.conf import settings as django_settings
-from pydantic import BaseModel, Field
+from pydantic import ConfigDict, BaseModel, Field
 
 
 class Settings(BaseModel):
@@ -19,9 +19,7 @@ class Settings(BaseModel):
         "ninja.pagination.LimitOffsetPagination", alias="NINJA_PAGINATION_CLASS"
     )
     PAGINATION_PER_PAGE: int = Field(100, alias="NINJA_PAGINATION_PER_PAGE")
-
-    class Config:
-        from_attributes = True
+    model_config = ConfigDict(from_attributes=True)
 
 
 settings = Settings.model_validate(django_settings)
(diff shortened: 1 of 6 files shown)


vitalik:
Hi @adamchainz 

yeah, I'm still thinking about this... (see also discussion here #934 )

basically Pydantic wants to put stuff to model_config, while Django people are very familiar with `class Meta`

maybe the most optional solution would be to allow users to set pydantic settings in `Meta` and then copy them to model_config on class creation

jkgenser:
I think there should be minimal django ninja for wrapping pydantic and use the pydantic primitive. Otherwise users need to know django-ninja + pydantic. But on average users are going to be much more familiar with pydantic as it's widely used

vitalik:
@jkgenser 

but what do you mean by adamchainz's approach ?

to have all handled by Meta:

```Python

class SomeData(ModelSchema):
     class Meta:
           model = User
           fields = '__all__'
           alias_generator=to_camel # < this is a pydantic setting

```

or split with model_config:

```Python
class SomeData(ModelSchema):
     model_config = ConfigDict(alias_generator=to_camel)
     class Meta:
           model = User
           fields = '__all__'
```

KlemenS189:
My two cents here. I dislike `Meta` class approach in django. It's quite common that you have a BaseClass with it's own `Meta` class. Then you create a custom model/schema and inherit the BaseClass. If you then create a new `Meta` class in this new model, it's quite easy to forget inheritance of meta class from parent.

Example:
```python
class BaseSchema(Schema):
    field: str
    
    class Meta:
        config = ConfigDict(populate_by_name=True, alias_generator=to_camel)


class UserSchema(BaseSchema):
    name: str
    class Meta:
        __fields__ = ["name"]
```

In this case, the Meta class in BaseSchema will be ignored.

vitalik:
this seems finally complete

PR #950, opened 2023-11-22
Drop support to unsupported python and django versions

jairhenrique:
https://endoflife.date/python
https://endoflife.date/django


.github/workflows/test_full.yml
```diff
@@ -10,15 +10,9 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12']
-        django-version: ['<3.0', '<3.1', '<3.2', '<3.3', '<4.1', '<4.2', '<4.3']
+        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
+        django-version: ['<3.3', '<4.2', '<4.3']
```
ulgens, line 14: I'm not really sure what's going on with `<` notation (which is my bad), but I don't I'm alone on it and I think it would be better if the version defined in a clear manner.

vitalik:
I would keep for now python3.7 (I still see 1.5% of downloads using it) and it is supported by pydantic

and django ~3.1 is still used in the wild (as I know from some feedbacks)

So i would rather keep deps as it is - to give people less headaches migrating to ninja-v1

on the other hand django 2.x and 3.0 can be dropped (but then need to cleanup ninja.compatibility package)

vitalik:
Thank you

PR #952, opened 2023-11-23
add_api_operation refactor

shimakaze-git:
Hello,

I've refactored the conditional logic in the `add_api_operation` method of the `Router` class for improved readability. The new approach uses the dictionary's `get` method, which simplifies the code.

Here's the updated code snippet:

```python
path_view: Optional[PathView] = self.path_operations.get(path)
if not path_view:
    path_view = PathView()
    self.path_operations[path] = path_view
```

Thank you for considering this change.


vitalik:
thank you - but seems conflicting now

PR #974, opened 2023-12-02
Support django model type annotations

rchoquet:
Hey 👋

First, thanks a lot for this lib! It really helped us deliver better quality software.

We are developing a (close to) fully typed application using Django. In lots of models we explicitly specify the type of the fields when the basic type is not precise enough, for example:

```python
TaskStatus = Literal["todo", "done"]
class TaskConfiguration(TypedDict):
  foo: int
  bar: str

class Task(models.Model):
  status: TaskStatus = models.CharField()
  configuration: TaskConfiguration = models.JSONField()
```

The issue arises when we define the associated Pydantic models, we have to redefine the types for those fields:

```python
class TaskSchema(ModelSchema):
  status: TaskStatus
  configuration: TaskConfiguration

  class Meta:
      model = Task
      fields = "__all__"
```

This PR makes `ModelSchema` aware of the type annotations and allow the previous schema to be defined as
```python
class TaskSchema(ModelSchema):
  class Meta:
      model = Task
      fields = "__all__"
```
... so we are not repeating the type and avoid runtime errors when there is a Django model/Pydantic model mismatch.

This is a draft that would require more work/testing, let me know if you think this is an idea worth pursuing.

Cheers!

### DIFF
--- a/ninja/orm/factory.py
+++ b/ninja/orm/factory.py
@@ -1,3 +1,4 @@
+import inspect
 import itertools
 from typing import Any, Dict, Iterator, List, Optional, Set, Tuple, Type, Union, cast
 
@@ -6,7 +7,7 @@ from django.db.models import ManyToManyRel, ManyToOneRel, Model
 from pydantic import create_model as create_pydantic_model
 
 from ninja.errors import ConfigError
-from ninja.orm.fields import get_schema_field
+from ninja.orm.fields import ModelField, get_schema_field
 from ninja.schema import Schema
 
 # MAYBE:
@@ -71,8 +72,6 @@ class SchemaFactory:
 
         if custom_fields:
             for fld_name, python_type, field_info in custom_fields:
-                # if not isinstance(field_info, FieldInfo):
-                #     field_info = Field(field_info)
                 definitions[fld_name] = (python_type, field_info)
 
         if name in self.schema_names:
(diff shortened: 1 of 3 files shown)


vitalik:
Hi @rchoquet 

Looks good

 - Could you also add some docs / examples-to-documentation
 -  and to support python3.8/3.7 I think you need to import Literal from typing_extensions

ninja/orm/factory.py
```diff
@@ -55,7 +55,7 @@ def create_schema(
         if key in self.schemas:
             return self.schemas[key]
 
-        model_fields_list = self._selected_model_fields(model, fields, exclude)
+        model_fields_list = list(self._selected_model_fields(model, fields, exclude))
```
rchoquet, line 58: @vitalik I added a cast to `list` here to fix this bug:
If `optional_fields` was equal to `__all__`, `optional_fields = [f.name for f in model_fields_list]` a few lines below was exhausting the `model_fields_list` iterator, and made `for fld in model_fields_list` iterate over an empty list, producing an empty Pydantic model.

rchoquet:
@vitalik should be good now.

vitalik:
Hi @rchoquet looks good

could you also add docs example with JsonField - as this is actually a pretty often question

msopacua:
Question: is this common to type your fields instead of using django-stubs?
Also, if you support this, are you going to support stubs in general?

rchoquet:
>Question: is this common to type your fields instead of using django-stubs?

I don't know, but I think you have to if you want more than the primitive types (ex: enums of string Literal, json field, ...)

> Also, if you support this, are you going to support stubs in general?

What do you mean by that?

@vitalik sorry for the delay, I'll do the doc change today.

vitalik:
> Question: is this common to type your fields instead of using django-stubs?
> Also, if you support this, are you going to support stubs in general?

not this has nothing to do with stubs - this let's you overrule default typing annotation when schema is automatically crated from model

rchoquet:
@vitalik done

rchoquet:
Hey @vitalik

Anything left to do on my side to get this merged?

rchoquet:
Ping @vitalik

Issue #1017, opened 2023-12-18
Calling NinjaAPI.urls twice results in error


jlucas91:
Hi @vitalik ! 

First off - what a wonderful library! Thank you for your efforts. 

I have a question related to url resolution and how it intersects with testing. 

To start, calling `.urls` on an API instance twice results in the following error message:

The code - 
```
test = NinjaAPI()
test.urls
test.urls
```

Results in - 
```
ConfigError: Looks like you created multiple NinjaAPIs or TestClients
To let ninja distinguish them you need to set either unique version or urls_namespace
 - NinjaAPI(..., version='2.0.0')
 - NinjaAPI(..., urls_namespace='otherapi')
Already registered: ['api-1.0.0']
```

It appears this is due to the `_validate()` method within `NinjaAPI` that's called by `urls`. It implicitly checks to make sure `.urls` have never been called before. I found this to be a little confusing, particularly given the error message which implied multiple instantiation rather than multiple access. Why is retrieving the `.urls` twice considered to be an error? 

Taking this one step further, my real goal here was an attempt to test the API. Recognizing that testing is undocumented today, I tried to use the `TestClient` client provided by ninja. As an example:

```
from ninja.testing import TestClient
TestClient(test).get('/hello')  # test is the API instance instantiated above
```

I thought this should work given the `TestClient` signature - it expects a NinjaAPI or Router instance. However, this consistently failed with the same `ConfigError` message above. On digging in, it became clear that this was caused by both Django's URL config calling `test.urls` and the `TestClient` calling `test.urls`.

I was able to work around this by using a `Router` and testing the `Router` instead of the API directly - but I was surprised that I couldn't simultaneously test the API and use it in a URL config.

I was left with the following questions:
- Why does the API strictly validate that urls are only retrieved once? If it's incorrect to ever compute the URLs more than once, why not cache instead of strictly validate?
- How, if at all, would one test an API directly rather than a Router? Should the documentation or TestClient be updated if it's not possible or recommended?

If the changes are reasonably straightforward here, happy to help with a small PR.

vitalik:
hi @jlucas91 

basically the problem is that each operation(path) has is it's own url-name - and if you include it twice it will have a clashing naming 

Indeed this is not perfect solution - and I'm looking for some more elegant way to deal wit this

jlucas91:
[...]

Issue #664, opened 2023-01-26
[Feature request] Add `django-ninja` specific `TestCase`


barseghyanartur:
**Is your feature request related to a problem? Please describe.**
A `django-ninja` specific `TestCase` is needed. There are simply too many things to consider, when testing. Authentication is being one of them. CSRF tokens - another. It does not work as is (standard `Django` `TestCase` fails with `4xx` errors ).

**Describe the solution you'd like**
- A specific `TestCase` that does all heavy lifting.
- Proper documentation on testing (ATM - massively missing).

Zio-n:
Any update on official documentation for testing with Django-ninja?

bufke:
What features do folks want? Maybe I could work on this. Right now I write my tests very simplistically. I use django TestCase and post with content_type="application/json"

I would guess features are:

- NinjaTestCase that makes self.client the NinjaClient
- Automatic handling of json serialization - I think NinjaClient does this already
- Maybe a DRF-style .data attribute for response json response? I could go either way on this. It would make converting DRF to ninja easier. It's not that hard to use json().
- What is needed for authentication? I use self.client.force_login myself.
- Document the above

My opinion would be to make something very small and nice and well documented. Avoid too many features.

c4ffein:
@bufke I came up with something custom for 2 items of this list when implementing the [RealWorld Demo App for Django Ninja](https://github.com/c4ffein/realworld-django-ninja/) ( the helper is [here](https://github.com/c4ffein/realworld-django-ninja/blob/master/helpers/headered_client.py) and you can check one of the test files [here](https://github.com/c4ffein/realworld-django-ninja/blob/master/articles/tests.py))

So I reimplemented those slightly differently as 2 PR, as we don't have exactly the same use case if it ends-up here:

- Default headers set at the NinjaClientBase instanciation, that can be partly overwritten during requests: #1259
- `Maybe a DRF-style .data attribute for response json response? I could go either way on this. It would make converting DRF to ninja easier. It's not that hard to use json().`: #1260 

Not sure if this is the right approach, but that's what I ended-up implementing myself in my repo. I'm really not sure we want to have those in Ninja though, I enjoy the framework because it's KISS, as that makes it easier to both learn and adapt to your needs... Anyway, I think splitting this issue into multiple PR to see what gets traction or not may be the right way to do

Issue #403, opened 2022-03-24
How to mock auth?


rsteenwyk:
First off, thanks for the project, it has been a pleasure to use.

I'm running into an issue trying to sort out how to mock successful authentication in my tests. 

Ninja is being instantiated like this:
```
api = NinjaAPI(
    title="Tasks API",
    urls_namespace="tasks",
    parser=ORJSONParser(),
    auth=MyHttpBearerAuth(),
)
```
The `MyHttpBearerAuth` class just overrides the `authenticate` method from `ninja.security.HttpBearer`

```
class MyHttpBearerAuth(HttpBearer):
    def authenticate(self, request, token):
        try:
            at = AccessToken(token)
            at.verify_token()
            if at.decoded_token:
                return at.token_has_role("AllowedAccess")
            return False
        except Exception as e:
            return False
```

I'm using pytest to start writing some tests and trying to figure out how to mock `MyHttpBearerAuth.authenticate` to always return True for some happy path tests.

I'm using the pytest mocker plugin, and I've tried patching like:
```
mocker.patch("tasks.api.MyHttpBearerAuth.authenticate", return_value=True)
```
Where `tasks.api` is the file where Ninja is instantiated and my functions are defined, but I'm still getting 401's.

I also tried a couple different ways of patching the `api` NinjaAPI object directly to just set `auth=None` but couldn't get that to work either.

Not sure where I'm going wrong here, but hoping someone can lend a hand :)

vitalik:
Hi @rsteenwyk 

try overriding default test client... something like :

```Python
from ninja.testing import TestClient as BaseTestClient


class TestClient(BaseTestClient):

    def _build_request(self, method: str, path: str, data: dict, request_params: dict):
        token = <GENERATE YOUR TOKEN HERE> # <--------
        request_params['headers'] = {"Authorization": "Bearer " + token}
        return super()._build_request(method, path, data, request_params)

```

rsteenwyk:
Thanks for the response @vitalik

I don't think that will really help me here because I'd then still need to hit the external auth service (Azure AD), and I'm trying to avoid external dependencies in these tests.

vitalik:
maybe just add some settings switch

```Python
if settings.AUTH == 'test':
      auth = Dummy()
else:
    auth = Azure()

api = NinjaAPI(auth=auth)
```

rsteenwyk:
Yeah I'd thought of that, but was hoping to avoid needing to make changes to the code under test.

[...]

PR #1020, opened 2023-12-19
Add basic testing documentation

jlucas91:
Adding some very basic testing documentation to make it clear, at a minimum, a TestClient is available. 

I don't feel I understand the TestClient's design intentions to go much deeper than this, but I hope this, at least, presents a small guidepost for people looking to write their first test. 

Open questions:
- Is it possible to test an API directly rather than via a router? In my personal work, I was unable to do so. (see: https://github.com/vitalik/django-ninja/issues/1017)
- What if any recommendations should be made for mocking authentication? It's currently possible to mock auth via a combination of patching and setting dummy credentials (see: https://github.com/vitalik/django-ninja/issues/403) but the implementation depends on the auth method being used and the code under test.  

Open questions aside, I hope this is unambiguous and minimal enough to be merged as is as a basis for further improvement.

References: https://github.com/vitalik/django-ninja/issues/664

### DIFF
--- a/docs/docs/guides/testing.md
+++ b/docs/docs/guides/testing.md
@@ -1,0 +1,43 @@
+# Testing
+
+**Django Ninja** provides a test client to make it easy to test your API.
+
+To test the following API:
+```python
+from ninja import NinjaAPI, Schema
+
+api = NinjaAPI()
+router = Router()
+
+class HelloResponse(Schema):
+    msg: str
+    
+@router.get("/hello", response=HelloResponse)
+def hello(request):
+    return {"msg": "Hello World"}
+
+api.add_router("", router)
+```
+
+You can use the Django test class:
+```python
+from django.test import TestCase
+from ninja.testing import TestClient
+
+class HelloTest(TestCase):
+    def test_hello(self):
+        client = NinjaTestClient(router)
+        response = client.get("/hello")
+        
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.json(), {"msg": "Hello World"})
+```
+
+Arbitrary attributes can be added to the request object by passing keyword arguments to the client request methods:
+```python
+class HelloTest(TestCase):
+    def test_hello(self):
+        client = NinjaTestClient(router)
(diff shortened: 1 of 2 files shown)


OtherBarry:
@jlucas91 if you're testing a NinjaAPI instance that is already served by django, you can just test it like you would any other django endpoint, there's no need to pass it to the ninja test client. I think the ninja test client is more for when you're wanting to create an API purely for testing, or if you want to test a router independently of the rest of the API.

jlucas91:
Oh! That perhaps explains my confusion [here](https://github.com/vitalik/django-ninja/issues/1017) if the majority of the ninja community is testing using Django's existing test client.

That said - is that true if you want to test _only_ the view and not Django's entire resolution/middleware stack? Ex. testing similar to Django's [RequestFactory](https://docs.djangoproject.com/en/5.0/topics/testing/advanced/#django.test.client.RequestFactory) or DRF's [APIRequestFactory](https://www.django-rest-framework.org/api-guide/testing/#apirequestfactory). When I attempted to call the view directly, the returned object was not a response - it was the raw response schema or int/schema tuple. The Ninja TestClient handles converting the raw return of the view to a Response object - which was both useful and I believe the correct behavior.

Either way, happy to be educated on the recommended/preferred approach and happy to update the docs accordingly.

OtherBarry:
From personal experience I've always used the django test client (or a custom subclass of it at least) for testing my actual API, and the ninja test client for creating dummy APIs to test things like auth and permissions. Similarly, in your last link, @baseplate-admin is doing the same thing using django's testcase class, despite importing ninja's TestClient and their API instance, they're not actually using them in that code sample.

I can certainly see value in a NinjaRequestFactory that could properly generate everything needed to call an api func, which could then return an instance of the response schema. It would certainly remove a lot of boilerplate and overhead from my existing tests.

I'm not super clear on the use case for wanting to test your real API endpoints without any django aspects involved, but also with ninja converting the response to JSON for you. That being said the existing TestClient should be able to handle that use case in it's current form, so the bug in #1017 is a problem that should be fixed.

PR #1023, opened 2023-12-20
fix: fixing typo in code example for custom docs viewer

jkeyes:


jkeyes:
The failed codestyle step does not look related to the change made to the docs.

Issue #1006, opened 2023-12-14
[BUG] Attribute assignment fails if validate_assignment is set to True


bastiaan85:
**Describe the bug**
Since v1, attribute assignment fails with `AttributeError: 'DjangoGetter' object has no attribute '__dict__'` if validate_assignment is set to True. This may be related to #938.

```
In [52]: class FloatSchema(Schema):
    ...:     float_field_name: float
    ...:     class Config(Schema.Config):
    ...:         validate_assignment = True

In [53]: fs = FloatSchema(float_field_name=1.2)

In [54]: fs.float_field_name = "bla"
---------------------------------------------------------------------------
ValidationError etc etc as expected

In [55]: fs.float_field_name = 2.4
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[55], line 1
----> 1 fs.float_field_name = 2.4

File /opt/app-test/venv/lib/python3.11/site-packages/pydantic/main.py:786, in BaseModel.__setattr__(self, name, value)
    784     attr.__set__(self, value)
    785 elif self.model_config.get('validate_assignment', None):
--> 786     self.__pydantic_validator__.validate_assignment(self, name, value)
    787 elif self.model_config.get('extra') != 'allow' and name not in self.model_fields:
    788     # TODO - matching error
    789     raise ValueError(f'"{self.__class__.__name__}" object has no field "{name}"')

File /opt/app-test/venv/lib/python3.11/site-packages/ninja/schema.py:214, in Schema._run_root_validator(cls, values, handler, info)
    211     handler(values)
    213 values = DjangoGetter(values, cls, info.context)
--> 214 return handler(values)

AttributeError: 'DjangoGetter' object has no attribute '__dict__'
```
Without validate_assignment it works but then the validator isn't run for the field obviously
```
In [1]: from ninja import Schema

In [2]: class FloatSchema(Schema):
   ...:     float_field_name: float
   ...:     class Config(Schema.Config):
   ...:         validate_assignment = False
   ...:

In [3]: fs = FloatSchema(float_field_name=1.2)

In [4]: fs.float_field_name = 2.4

In [5]: fs.float_field_name = "bla"
# no error
In [6]: fs.float_field_name
Out[6]: 'bla'
```

**Versions (please complete the following information):**
 - Python version: 3.11.7
 - Django version: 5.0
 - Django-Ninja version: 1.1.0
 - Pydantic version: 2.52

austinpgraham:
[...]

PR #1024, opened 2023-12-21
Fix   dict   when validate assignment true

austinpgraham:
Issue reported [here](https://github.com/vitalik/django-ninja/issues/1006).

This PR fixes an issue with v1.1.0 where setting validate_assignment = True on a Schema throws an Attribute error for __dict__ on DjangoGetter. Example, from added test with change removed to force the error:
```
FAILED tests/test_schema.py::test_django_getter_validates_assignment - AttributeError: 'DjangoGetter' object has no attribute '__dict__'
```
It appears this functionality was broken when GetterDict was removed as a parent from DjangoGetter, and adding __dict__ in __slots__ seems to be the easiest fix.

I'm very open to feedback, and if this is intended or what a better fix might be, this PR is just me taking a stab at fixing something needed by my company.

Note: It appears running the linting fixed a couple other linting issues unrelated to this change.

### DIFF
--- a/ninja/openapi/schema.py
+++ b/ninja/openapi/schema.py
@@ -36,23 +36,21 @@ class OpenAPISchema(dict):
         self.securitySchemes: DictStrAny = {}
         self.all_operation_ids: Set = set()
         extra_info = api.openapi_extra.get("info", {})
-        super().__init__(
-            [
-                ("openapi", "3.1.0"),
-                (
-                    "info",
-                    {
-                        "title": api.title,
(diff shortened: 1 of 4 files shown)


vitalik:
Hi @austinpgraham 

Thank you

could you reformat some code-style changes with 
```
pip install -U ruff
ruff format ninja tests
```

austinpgraham:
Done, thanks @vitalik

bastiaan85:
@vitalik +1 for me too

PR #1026, opened 2023-12-22
also find "request" param by type annotation

davidszotten:
this re-enables (worked before 1.0) view signatures like

`def view(_: HttpRequest)` to avoid linters complaining about the unused `request` param


vitalik:
Hi @davidszotten thank for PR

I was thinking about a bit different approach for request param

basically now django-ninja supports any django views decorator (with decorate_view helper)

and Django's "protocol" requires(/expects) that first argument is always request object

so maybe implementation should  be the following

 - check only 1st parameter
     - if it has annotation - it must be of a type HttpRequest
     - if no annotation - it must be named "request"

vitalik:
or even simpler - just drop first parameter entirely

davidszotten:
thanks for the quick review! updated with the simpler solution for now (skip first param)

Issue #1011, opened 2023-12-16
paginate decorator causes validation to fail


johnbridstrup:
I have used the `@paginate` decorator successfully multiple times now, but for some reason it is causing validation to fail on a new endpoint (that works without pagination). I have a feeling this is because of the schema, which has a nested schema and a resolved value, because that is the only real difference between it and the working ones. Am I missing something from the documentation or is there a known reason this is happening?

**Schema**

```python
class MyModelResponseSchema(ModelSchema):
    related_model: RelatedModelResponseSchema
    calculated_property: int
    creator: UserLimitedSchema

    @staticmethod
    def resolve_calculated_property(obj):
        return obj.calculated_property

    class Meta:
        model = MyModel
        fields = [
            "related_model",
            "max_amount",
            "max_increment",
            "creator",
        ]
```

**Endpoint**

```python
@router.get(
    "myEndpoint",
    auth=JWTAuth(),
    response={200: List[MyModelResponseSchema], codes_4xx: DefaultError},
    url_name="service_my_endpoint",
)
@paginate  # <--- Works fine without this decorator
def my_swaps(request):
    return 200, MyModel.objects.all()
```

**Output**
```
pydantic_core._pydantic_core.ValidationError: 8 validation errors for NinjaResponseSchema
response.items.0.related_model
  Field required [type=missing, input_value=<ninja.schema.DjangoGette...bject at 0x7f8bbb8852c0>, input_type=DjangoGetter]
response.items.0.calculated_property
  Field required [type=missing, input_value=<ninja.schema.DjangoGette...bject at 0x7f8bbb8852c0>, input_type=DjangoGetter]
response.items.0.creator
  Field required [type=missing, input_value=<ninja.schema.DjangoGette...bject at 0x7f8bbb8852c0>, input_type=DjangoGetter]
response.items.0.max_amount
  Field required [type=missing, input_value=<ninja.schema.DjangoGette...bject at 0x7f8bbb8852c0>, input_type=DjangoGetter]
...
```

Any help would be nice, thanks!

johnbridstrup:
Okay, I've removed the nested schemas and resolver and its still happening. I am now very confused why its specifically happening for this model/endpoint and not the others, which appear to be exactly the same (return `.all()` for a simple query)

johnbridstrup:
I figured out why this is happening. It seems you cannot include an explicit response code when returning from the method. I think this should be probably allowed. I'll look into making a PR at some point

[...]

Issue #940, opened 2023-11-20
Custom pagination for multiple response types?


SonQBChau:
How can I implement custom pagination for multiple response types? I've successfully applied custom pagination with `response=List[MySchema]`, but encountered issues with a structure like `response={200:List[MySchema], 400: Dict[str,str]}`. Here's what I've attempted so far:

```python
class CustomPagination(PaginationBase):
    class Input(Schema):
        page: int = Field(None, ge=1)
        page_size: int = Field(None, ge=1)

    class Output(Schema):
        items: List[Any] = []
        count: Optional[int]
        next_page: Optional[int]
        previous_page: Optional[int]

    def paginate_queryset(self, queryset, pagination: Input, **params):
        DEFAULT_PAGE = 1
        DEFAULT_PAGE_SIZE = 100
        pagination.page = pagination.page or DEFAULT_PAGE
        pagination.page_size = pagination.page_size or DEFAULT_PAGE_SIZE

        offset = (pagination.page - 1) * pagination.page_size

        # trying to catch bad requests here
	if isinstance(queryset, tuple) and queryset[0] == HTTPStatus.BAD_REQUEST:
            return queryset
            
        total_items = queryset.count()
        items = list(queryset[offset : offset + pagination.page_size])

        next_page = (
            pagination.page + 1 if offset + pagination.page_size < total_items else None
        )
        previous_page = pagination.page - 1 if pagination.page > 1 else None

        return {
            "items": items,
            "count": len(items),
            "next_page": next_page,
            "previous_page": previous_page,
        }


class CustomRouterPaginated(RouterPaginated):
    def __init__(self, *args: Any, **kwargs: Any) -> None:
        super().__init__(*args, **kwargs)
        self.pagination_class = CustomPagination
        
router = CustomRouterPaginated()
@router.get("/books", response={200:List[BookSchema], 400: Dict[str,str]})
def list_books(request, filters: BookFilterSchema = Query(...)):
    books = Book.objects.all()
    books = filters.filter(books)
    # custom logic with filter, return error if not satisfied
	    return HTTPStatus.BAD_REQUEST, error_message
    return books
```

vitalik:
Hi @SonQBChau 

yeah.. the pagination router does not support multiple codes (for now I guess)

what you can do at this moment - define response as just list and throw http error with message:


```Python
@router.get("/books", response=List[BookSchema]). # <-----
def list_books(request, filters: BookFilterSchema = Query(...)):
    ...
    if error:
[...]

PR #1028, opened 2023-12-23
[enhancement] Allow returning status explicitly with pagination

johnbridstrup:
I've allowed the pagination decorator to still work if the method returns `status_code, results`. I understand it will almost always be `200` for paginated responses, but I like to explicitly set the codes anyway.

I've added this to `_inject_pagination`
```python
# Check if function returns status code + iterable
code = None
is_two_tuple = isinstance(items, tuple) and len(items) == 2
if (
    is_two_tuple
    and isinstance(items[0], int)
    and isinstance(items[1], Iterable)
):
    code = items[0]
    items = items[1]
...
if code:
    return code, result
return result
```

I suppose I could see a case where someone returns a tuple instead of a list or queryset, and that it could possibly be exactly 2 entries, so if there is a more safe way I'm happy to implement that.

### DIFF
--- a/ninja/pagination.py
+++ b/ninja/pagination.py
@@ -1,7 +1,7 @@
 import inspect
 from abc import ABC, abstractmethod
 from functools import partial, wraps
-from typing import Any, Callable, List, Optional, Tuple, Type
+from typing import Any, Callable, Iterable, List, Optional, Tuple, Type
 
 from django.db.models import QuerySet
 from django.http import HttpRequest
@@ -141,12 +141,26 @@ def _inject_pagination(
 
         items = func(request, **kwargs)
 
+        # Check if function returns status code + iterable
+        code = None
+        is_two_tuple = isinstance(items, tuple) and len(items) == 2
+        if (
+            is_two_tuple
+            and isinstance(items[0], int)
+            and isinstance(items[1], Iterable)
+        ):
+            code = items[0]
+            items = items[1]
+
         result = paginator.paginate_queryset(
             items, pagination=pagination_params, request=request, **kwargs
         )
         if paginator.Output:  # type: ignore
             result[paginator.items_attribute] = list(result[paginator.items_attribute])
             # ^ forcing queryset evaluation #TODO: check why pydantic did not do it here
+
+        if code:
+            return code, result
         return result
 
     contribute_operation_args(
--- a/tests/test_pagination.py
+++ b/tests/test_pagination.py
@@ -143,6 +143,36 @@ def items_9(request):
     return list(range(100))
 
 
+@api.get("/items_10", response=List[int])
+@paginate
+def items_10(request):
+    return 200, ITEMS
+
+
+@api.get("/items_11", response=List[int])
+@paginate(CustomPagination)
+def items_11(request):
+    return 200, ITEMS
+
+
+@api.get("/items_12", response=List[int])
+@paginate(PageNumberPagination, page_size=10)
+def items_12(request, **kwargs):
+    return 200, ITEMS
+
+
+@api.get("/items_13", response={101: int, 200: List[Any]})
+@paginate(PageNumberPagination, page_size=10, pass_parameter="page_info")
+def items_13(request, **kwargs):
+    return 200, ITEMS + [kwargs["page_info"]]
+
+
+@api.get("/items_14", response=List[int])
+@paginate(NoOutputPagination)
+def items_14(request):
+    return 200, list(range(15))
+
+
 client = TestClient(api)
 
 
@@ -340,3 +370,29 @@ def test_config_error_NOT_SET():
         @paginate
         def invalid2(request):
             pass
+
+
+def test_explicit_code_1():
+    response = client.get("/items_10?limit=10").json()
+    assert response == {"items": ITEMS[:10], "count": 100}
+
+
+def test_explicit_code_2():
+    response = client.get("/items_11?skip=5").json()
+    assert response == {"items": ITEMS[5:10], "count": "many", "skip": 5}
+
+
+def test_explicit_code_3():
+    response = client.get("/items_12?page=2").json()
+    assert response == {"items": ITEMS[10:20], "count": 100}
+
+
+def test_explict_code_4():
+    page = 11
+    response = client.get(f"/items_13?page={page}").json()
+    assert response == {"items": [{"page": 11}], "count": 101}
+
+
+def test_explicit_code_5():
+    response = client.get("/items_14?skip=10").json()
+    assert response == [10, 11, 12, 13, 14]


johnbridstrup:
Relevant issue #1011

pmdevita:
Just ran into this problem myself, would be nice to see this merged!

vitalik:
I'm not sure about this one... my thoughts is actually returning tuples with included status was originally not best design idea

 I'm actually leaning into some check that forces user output only querysets in paginated responses

benjaoming:
I didn't notice this PR when I opened https://github.com/vitalik/django-ninja/pull/1148 because I went for fixing #940 :smile: 

re @vitalik 

> I'm not sure about this one... my thoughts is actually returning tuples with included status was originally not best design idea
> 
> I'm actually leaning into some check that forces user output only querysets in paginated responses

I actually like the easiness of returning (status, <iterable>). However, I can see that it doesn't "scale", it's a very implicit data structure, which caused the issue here.

But as you can also see from the fix, it was very simple... but could definitely look more explicit.

I'm not sure how forcing only a queryset would make it impossible for list views that need pagination to benefit from differentiated status codes?

For my case, the reason that differentiated schemas+status codes are important: It enables entrypoints to implement authorization levels and return different schemas depending on access. There may be other similar cases.

My example could be a `book_list` that wants to return `200, BooksForAnonymousSchema` and `222, BooksForReturningCustomerSchema`. The `BooksForReturningCustomerSchema` would include annotations about what the user has rated the book, if they own a copy etc. I'm not sure if this is good API design, but this is currently where I've landed.

OscarVanL:
I'd really appreciate it if this bug could be fixed.

I have a bunch of APIs that have explicit/concrete return Schemas for errors:
```
@router.get("/{uuid:tenant_id}/car", response={200: list[Car], 403: UnauthorizedError})
def get_cars(request: HttpRequest, tenant_id: UUIDStr, filters: CarFilter = Query()):
    allowed = permissions_checker.can(request.user.uuid, permissions.read_cars, tenant_id)
    if not allowed:
        return 403, UnauthorizedError()

    return 200, filters.filter(cars_for_tenant(tenant_id))
```

I recently tried to switch these APIs to use pagination, like this:
```
@router.get("/{uuid:tenant_id}/car", response={200: list[Car], 403: UnauthorizedError})
@paginate
def get_cars(request: HttpRequest, tenant_id: UUIDStr, filters: CarFilter = Query()):
    allowed = permissions_checker.can(request.user.uuid, permissions.read_cars, tenant_id)
    if not allowed:
        return 403, UnauthorizedError()

    return 200, filters.filter(cars_for_tenant(tenant_id))
```

After making the switch, I get pydantic `ValidationError` in that 403 case, and all other cases where I explicitly set the status code.

Obviously, this example is very simple compared to some of my APIs that have rather complex error handling.

Switching everything to raise `HttpError`s (as suggested in the workaround [here](https://github.com/vitalik/django-ninja/issues/940#issuecomment-1820513020)) is a frustratingly large change, especially as I have custom error response schemas/structures. Additionally, using exceptions for errors means the exception's API errors won't show up on my OpenAPI schema.


IMO the semantics around response codes for paginated responses should be the same as non-paginated responses. I don't understand why we'd stop caring about explicit HTTP response codes _just_ because the API is paginated.

# This Fix

I tried including including the fix in this PR on my own Ninja project, but it still didn't solve my use-case and crashes in my Unauthorized case:
```
  File "/service/src/ninja_fork.py", line 288, in view_with_pagination
    result = paginator.paginate_queryset(items, pagination=pagination_params, request=request, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/service/venv/lib/python3.12/site-packages/ninja/pagination.py", line 100, in paginate_queryset
    "items": queryset[offset : offset + limit],
             ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'UnauthorizedError' object is not subscriptable
```

I believe this is because the PR's change to check for `isinstance(items[1], Iterable)` is too loose of a check. For instance, in my scenario, the Ninja `Schema` is considered iterable:
```
from ninja import Schema

class UnauthorizedError(Schema):
    message: str | None

print(isinstance(UnauthorizedError(message="hello world"), Iterable))

> True
```

But obviously when Ninja tries to paginate `UnauthorizedError` that does not work.

Potentially this check could be made stricter by only paginating returned values that are `QuerySet`s and `lists` like this:
```
        # Check if function returns status code + iterable
        code = None
        is_two_tuple = isinstance(items, tuple) and len(items) == 2
        if is_two_tuple and isinstance(items[0], int) and isinstance(items[1], Iterable):
            code = items[0]
            items = items[1]

        if isinstance(items, QuerySet) or isinstance(items, list):
            result = paginator.paginate_queryset(items, pagination=pagination_params, request=request, **kwargs)
            if paginator.Output:  # type: ignore
                result[paginator.items_attribute] = list(result[paginator.items_attribute])
                # ^ forcing queryset evaluation #TODO: check why pydantic did not do it here
        else:
            result = items

        if code:
            return code, result
        return result
```

OscarVanL:
> I'm not sure about this one... my thoughts is actually returning tuples with included status was originally not best design idea
> I'm actually leaning into some check that forces user output only querysets in paginated responses

@vitalik I'm curious if your recent change in 1.6.0 to requiring concrete returns of `return Status(200, body)` is part of the bigger picture to addressing this point?

I still think there is a case to support non-paginated responses in error scenarios like my scenario in the previous comment.

But I understand that maybe you want this to be more explicit...

Could it make sense to differentiate response types by a returned class instead?

* `return Status(403, UnauthorizedError())` for a non-paginated response (added in 1.6.0)
* `return Paginate(200, qs)` (where `qs` is a QuerySet to paginate) for a paginated response?

vitalik:
@OscarVanL - Status responses[ are as well supported in pagination](https://github.com/vitalik/django-ninja/blob/0b67d47637a9e831287447a6691d1c8c37a6eccb/tests/test_status.py#L117-L149) in 1.6+

you should be able to do stuff like

```python

@api.get("/some", response={201: List[UserOut]})
@paginate
def paginated_status(request):
    return Status(201, qs)

```

OscarVanL:
@vitalik Apologies if I was not clear, my paginate function has some error cases that contain non-queryset bodies and these do not function even post-1.6.

For example:
```
@router.get("/{uuid:tenant_id}/car", response={200: list[Car], 403: UnauthorizedError})
@paginate
def get_cars(request: HttpRequest, tenant_id: UUIDStr, filters: CarFilter = Query()):
    allowed = permissions_checker.can(request.user.uuid, permissions.read_cars, tenant_id)
    if not allowed:
        return Status(403, UnauthorizedError())

    return Status(200, filters.filter(cars_for_tenant(tenant_id)))
```

In the error case, I end up with a crash because the `UnauthorizedError` does not implement `len()`:
```
Test permission denied case
----------------------------------------------------------------------
  File "/service/venv/lib/python3.12/site-packages/ninja/pagination.py", line 71, in _items_count
    return queryset.all().count()
           ^^^^^^^^^^^^
  File "/service/venv/lib/python3.12/site-packages/pydantic/main.py", line 991, in __getattr__
    raise AttributeError(f'{type(self).__name__!r} object has no attribute {item!r}')
AttributeError: 'UnauthorizedError' object has no attribute 'all'

During handling of the above exception, another exception occurred:

  File "/service/venv/lib/python3.12/site-packages/django/test/client.py", line 1124, in get
    response = super().get(
               ^^^^^^^^^^^^
  File "/service/venv/lib/python3.12/site-packages/django/test/client.py", line 475, in get
    return self.generic(
           ^^^^^^^^^^^^^
  File "/service/venv/lib/python3.12/site-packages/django/test/client.py", line 671, in generic
    return self.request(**r)
           ^^^^^^^^^^^^^^^^^
  File "/service/venv/lib/python3.12/site-packages/django/test/client.py", line 1087, in request
    self.check_exception(response)
  File "/service/venv/lib/python3.12/site-packages/django/test/client.py", line 802, in check_exception
    raise exc_value
  File "/service/venv/lib/python3.12/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^
  File "/service/venv/lib/python3.12/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/service/venv/lib/python3.12/site-packages/ninja/operation.py", line 661, in sync_view_wrapper
    return self._sync_view(request, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/service/venv/lib/python3.12/site-packages/ninja/operation.py", line 673, in _sync_view
    return operation.run(request, *a, **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/service/venv/lib/python3.12/site-packages/ninja/operation.py", line 225, in run
    return self.api.on_exception(request, e)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/service/venv/lib/python3.12/site-packages/ninja/main.py", line 617, in on_exception
    return handler(request, exc)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/service/venv/lib/python3.12/site-packages/ninja/errors.py", line 129, in _default_exception
    ^^^^^^^^^
  File "/service/venv/lib/python3.12/site-packages/ninja/operation.py", line 216, in run
    result = self.view_func(request, **values)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/service/venv/lib/python3.12/site-packages/ninja/pagination.py", line 669, in view_with_pagination
    result = paginator.paginate_queryset(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/service/src/.../ninja.py", line 274, in paginate_queryset
    "totalCount": self._items_count(queryset),
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/service/venv/lib/python3.12/site-packages/ninja/pagination.py", line 73, in _items_count
    return len(queryset)
           ^^^^^^^^^^^^^
TypeError: object of type 'UnauthorizedError' has no len()
```

Ideally I'd like a way to have my paginated APIs return errors, but still be paginated for the 200 case.

vitalik:
@OscarVanL  the recomended errors mechanism is via exception raising:

```python
from ninja.errors import HttpError

@api.get("/some/resource")
def some_operation(request):
    if True:
        raise HttpError(403, "Unauthorized")
```

custom structure output handler

```python


@api.exception_handler(MyUnauthorizedError)
def custom_error_response(request, exc):
    return api.create_response(
        request,
        {"message": "Unauthorized! Please authorize!"},
        status=403,
    )


# some logic that throws exception

@api.get("/service")
def some_operation(request):
    if random.choice([True, False]):
        raise MyUnauthorizedError()
    ...
```

more details - https://django-ninja.dev/guides/errors/#custom-exception-handlers

vitalik:
@johnbridstrup - I'm closing this PR as we are moving away from tuple results in favor for Status objects (which you can now use in pagination as well)

OscarVanL:
@vitalik 
Thanks for the examples. This approach works, but one of the features my frontend colleagues love about Ninja is the error responses being in the OpenAPI schemas.

When I use "error returns", Nina's validation of API responses keeps me honest that the returned errors match the schema (and if this changes my unit tests will fail), whereas exception-based error responses can be missed in the schema unintentionally since this does not get validated. 

For this reason I try to avoid using exception based error handling as much as possible. 

This use case is why I'd prefer it if Ninja supported "error returns" inside paginated APIs, not just exception based handling. 

Thank you :)

PR #1033, opened 2024-01-02
Implement choices as Literal for PyDantic

msopacua:
Adds the ability to render choices as Literals, with a `choices_exclude` meta configuration to exclude fields that should not be literals.


Zerotask:
Any updates on this? Would be a good addition for django-ninja

PR #1039, opened 2024-01-07
Minor Typing Improvement for `FilterSchema`

max-muoto:
This PR makes a minor typing improving to the `filter` method in `FilterSchema`

Right now, typing information for the queryset's model is lost since we return an arbitrary queryset. This PR uses a generic parameter to ensure typing information is not lost.


vitalik:
Thank you

Issue #965, opened 2023-11-28
Handling duplicate schema names


shughes-uk:
If I have created two schema's with identical names, in the openapi spec ninja will silently use whatever the "last" definition is. This has caused us some headaches.

```python

# schema1.py
class FooSchema():
    x: int

#schema2.py
class FooSchema():
    y: int
```

I personally would be fine with this condition being an import time exception if it happens. Perhaps it could also be avoided by prefixing the schema names with some unique info.

It's a footgun/bug in its current state, I'm not sure there's a legitimate reason to ever want to overwrite schema like this but I've been surprised by things before!

PR #1047, opened 2024-01-13
Add SCHEMA_GENERATION_CLASS to settings, Allow NinjaGenerateJsonSchema to be overridden

Eittipat:
Hi, 

This pull request allow `NinjaGenerateJsonSchema` to be changed via settings

**Usecase**
I encounter a problem like https://github.com/vitalik/django-ninja/issues/965.
while waiting for bug to be resolved. I have a workaround solution by custom  behaviour of `get_defs_ref` inside NinjaGenerateJsonSchema 

Allowing `NinjaGenerateJsonSchema` to be changed via settings enable clean solution to do this.
(I don't want to do monky patch)

### DIFF
--- a/ninja/conf.py
+++ b/ninja/conf.py
@@ -25,6 +25,10 @@ class Settings(BaseModel):
     PAGINATION_PER_PAGE: int = Field(100, alias="NINJA_PAGINATION_PER_PAGE")
     PAGINATION_MAX_LIMIT: int = Field(inf, alias="NINJA_PAGINATION_MAX_LIMIT")
 
+    SCHEMA_GENERATION_CLASS: str = Field(
+        "ninja.schema.NinjaGenerateJsonSchema", alias="NINJA_SCHEMA_GENERATION_CLASS"
+    )
+
     class Config:
         from_attributes = True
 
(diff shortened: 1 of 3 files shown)


vitalik:
Hi @Eittipat 

Thank you for PR - could you please add at at least one test case for this and documentaition ?


Thank you

Eittipat:
@vitalik , Test, Guide and Doc added.

vitalik:
Hi @Eittipat 

thank you

I was wondering #965 should be addressed in the core ? Could you share your implementation to overcome duplicated schema names in different modules ?

vitalik:
I assume you doing something like schema name in json becomes a combination of module-name + class name ? 

I think the ideal could be that name is extened with a counter if duplicate is created (FooSchema, FooSchema2, FooSchema3 ...)

Eittipat:
Yes, I go with approach  `module-name + class name`, I will try follow your suggestion in another pull request.
I think fixing https://github.com/vitalik/django-ninja/issues/965 should be another task.

My implementation is simple:

```python
class CustomNinjaGenerateJsonSchema(NinjaGenerateJsonSchema):

    def get_defs_ref(self, core_mode_ref: CoreModeRef) -> DefsRef:
        module_qualname_occurrence_mode = super().get_defs_ref(core_mode_ref)
        name_choices = self._prioritized_defsref_choices[module_qualname_occurrence_mode]
        name_choices.pop(0)
        name_choices.pop(0)
        self._prioritized_defsref_choices[module_qualname_occurrence_mode] = name_choices
        return module_qualname_occurrence_mode
```

PR #1072, opened 2024-02-06
dict() missed from the deprecated list

xncbf:
The dict method is missing from the PydanticDeprecatedSince20 list.


vitalik:
Thank you


well, pydantic introduces too much of a word "model_" in their api which really confises django users

here we specifically leave .dict on schema method to address both compatibility and avoiding word "model" confusion

PR #1076, opened 2024-02-09
docs(operations): minor adjustments

Zerotask:
- https://docs.python.org/3/library/http.html#http-methods
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods

### DIFF
--- a/docs/docs/guides/input/operations.md
+++ b/docs/docs/guides/input/operations.md
@@ -42,7 +42,7 @@ reference docs for information on what you can pass to any of these decorators.
 If you need to handle multiple methods with a single function for a given path,
 you can use the `api_operation` decorator:
 
-```python hl_lines="3"
+```python hl_lines="1"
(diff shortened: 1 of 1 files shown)


vitalik:
Hi @Zerotask 

I"m not sure about this part
```
@api.api_operation([HTTPMethod.POST, HTTPMethod.PATCH], "/path")

Zerotask:
> Hi @Zerotask 
> 
> I"m not sure about this part
> ```
> @api.api_operation([HTTPMethod.POST, HTTPMethod.PATCH], "/path")

My intention was to provide a default way which supports auto completion and ensures it's written correctly. For example, some people could accidentally write OPTION instead of OPTIONS or they have to lookup all available HTTP methods on the web.
Personally I also prefer enums over raw strings.

Zerotask:
> this makes it longer, and it is not really a place which people use often or make mistakes...
but redabitliy of the code is degrading...

I changed it back.

PR #1096, opened 2024-02-27
Adding create_response method to class Router

atomtables:
The class `Router` does not have the method `create_response` attached to it, which is not consistent with the behavior of the main class `NinjaAPI`. As such, I have added this method in the following pull request, and documented it under Routers in the documentation.

I also changed the default HTTP status-code of the method `create_response` to 200 in both `Router` and `NinjaAPI`, so that the method can be called without error if the parameter `status` is not defined.

Other changes came from reformatting using `make lint`. 

After doing a coverage test on the app, 99.37% of the app is covered, however this 0.63% seems to be from a skipped test from pre-commit, and it seems that this modification is both stable and passing, both with custom app testing, and unit testing.

### DIFF
--- a/docs/docs/guides/routers.md
+++ b/docs/docs/guides/routers.md
@@ -125,6 +125,26 @@ Now, include `api` to your urls as usual and open your browser at `/api/docs`, a
 
 ![Swagger UI Simple Routers](../img/simple-routers-swagger.png)
 
+## Router response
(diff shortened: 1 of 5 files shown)


atomtables:
@vitalik does this look good? just asking

atomtables:
@vitalik bump

Issue #1087, opened 2024-02-18
Is there a way to access extra keys of the query?


Vulwsztyn:
I'm trying to access extra keys of the query.
I have a minimal reproducible example here:
https://github.com/Vulwsztyn/django-ninja-testing/tree/extra_keys_in_query_test

the important part is:
```
class UserSchema(Schema):
    model_config = ConfigDict(extra='allow') <-- this is how I tried to make the extra keys accessible
    username: str

@api.post("/hello", response={200: UserSchema})
def hello(request, body: UserSchema):
    print(body)
    print(body.__pydantic_extra__)
    return 200, body
```

when I send `curl -X POST 'http://127.0.0.1:8000/api/hello' -d '{"username": "asd", "extra_key": 3}' -H "Content-Type: application/json"`

the `extra_key` is lost

Is there a way to access it?

vitalik:
Well schema's tend to describe/document everything that can be passed to it..

maybe it makes more sense to put it to schema as optional field:


```Python
class UserSchema(Schema):
    username: str
    extra_key: int = None

@api.post("/hello")
def hello(request, body: UserSchema): 
     return body.dict(exclude_unset=True) # <-- this will not have extra_key if it not passed
```

if you want to "hide" extra_field from documentation then just access raw body from request

```Python
class UserSchema(Schema):
    username: str

class HiddenSchema(Schema):
    extra_key: int = None

@api.post("/hello")
def hello(request, body: UserSchema): 
     hidden = HiddenSchema(**request.json())
```

I'm not sure yet why `extra='allow'` does not applies when request.json passed to schema - but I think security is involved - pretty dangerous to get everything from user input and then use it blindly

Vulwsztyn:
This is not what the question was about.

I have an endpoint that can accept a json with any keys and in order not to change its implementation with the migration to django ninja I need to access the extra keys.

vitalik:
well my second example should do it

PR #1097, opened 2024-02-28
Allow Schema model_config.extra = "allow"

soccer99:
Allow setting the model_config key `extra="allow"`

```python
# schemas.py

class ExtraAllowedSchema(Schema):
    name: str

    class Config(Schema.Config):
        extra = "allow"

# views.py
@api.post("/extra")
def extra_params(request, payload: ExtraAllowedSchema):
    """
    Test that a Schema with model_config.extra = "allow" will allow keys not defined in schema
    """
    return payload
```

A request with extra keys will be honored and returned

```bash
# Request
POST: /extra
JSON payload {"name": "my name", "age": 100}
Response
{"name": "my name", "age": 100}

```

Also works on response schemas if you return a dictionary with extra keys outside your OutSchema.

Fixes issue: https://github.com/vitalik/django-ninja/issues/1087

### DIFF
--- a/docs/docs/guides/response/config-pydantic.md
+++ b/docs/docs/guides/response/config-pydantic.md
@@ -9,6 +9,46 @@ There are many customizations available for a **Django Ninja `Schema`**, via the
     when using Django models, as Pydantic's model class is called Model by default, and conflicts with
     Django's Model class.
 
+## Allow extra fields in POST request body
+
+You can allow extra fields to be passed to your view using the Pydantic model config option `extra="allow"`
+
+```python
+from ninja import Schema
+
+class UserInSchema(Schema):
+    name: str
+
+    class Config(Schema.Config):
+        extra = "allow"
+
+# ---
+# views.py
+
+@api.post("/users")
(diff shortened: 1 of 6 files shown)


p-baum:
This change causes extra fields to be included in the main parsed data - clearly not what we want.  We need the extra data stored in the  [model_extra ](https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_extra) attribute.

Vulwsztyn:
> This change causes extra fields to be included in the main parsed data - clearly not what we want.

Not really. It's ok.

soccer99:
Including the extra fields at the top level is the default pydantic behavior when using extra = "allow".  I suppose we could only expose those extra fields in an `extra` param on the view function if thats what is decided.

Vulwsztyn:
@vitalik Could you take a look?

I'm sorry if it is not ok that I ping you, but I'm interested in this feature

vitalik:
Still thinking about it (need more testing on like how it will behave wit schema have resolvers or handles objects outputs instead of dicts...)

Issue #1019, opened 2023-12-19
[BUG] TypeError: cannot specify both default and default_factory


jaeyongjaykim:
**Describe the bug**

**Versions (please complete the following information):**
 - Python version: 3.12
 - Django version: 5.0
 - Django-Ninja version: 1.1.0
 - Pydantic version: 2.5.2

Note you can quickly get this by runninng in `./manage.py shell` this line:
```
import django; import pydantic; import ninja; django.__version__; ninja.__version__; pydantic.__version__
```

# TypeError: cannot specify both default and default_factory
```
  File "/Users/jaeyong/_git/asdf-server/asdf/apps/tracking/v1/schema.py", line 8, in <module>
    class ReferrerIn(ModelSchema):
  File "/Users/jaeyong/.local/share/virtualenvs/asdf-server-0h85sao-/lib/python3.12/site-packages/ninja/orm/metaclass.py", line 106, in __new__
    model_schema = create_schema(
                   ^^^^^^^^^^^^^^
  File "/Users/jaeyong/.local/share/virtualenvs/asdf-server-0h85sao-/lib/python3.12/site-packages/ninja/orm/factory.py", line 65, in create_schema
    python_type, field_info = get_schema_field(
                              ^^^^^^^^^^^^^^^^^
  File "/Users/jaeyong/.local/share/virtualenvs/asdf-server-0h85sao-/lib/python3.12/site-packages/ninja/orm/fields.py", line 169, in get_schema_field
    FieldInfo(
  File "/Users/jaeyong/.local/share/virtualenvs/asdf-server-0h85sao-/lib/python3.12/site-packages/pydantic/fields.py", line 194, in __init__
    raise TypeError('cannot specify both default and default_factory')
TypeError: cannot specify both default and default_factory
```

# ReferrerIn(ModelSchema):
```
class ReferrerIn(ModelSchema):
    class Meta:
        model = Referrer
        exclude = ["id", "created_at", "created_by", "updated_at", "updated_by", "is_deleted"]
        fields_optional = "__all__"
```

# Short Explanation
I am migrating Django-Ninja from `0.22.2` to `1.1.0` and I encountered this error.
It seems like `model` + `exclude` + `fields_optional` combination raises TypeError.

vitalik:
@jaeyongjaykim 

there is related bug #1016 that I cannot reproduce

maybe you can show me full listing for the `Referrer` model ?

jaeyongjaykim:
@vitalik 

[...]

Issue #1080, opened 2024-02-13
[Question] TypeError: cannot specify both default and default_factory


unique1o1:
I'm unable to  use `fields_optional='__all__'` in some models? 
 
here are my models:
```python
class Topic(models.Model):
	name = models.CharField(max_length=100, db_index=True)
	certificate = models.ForeignKey(
		to=Certificate, on_delete=models.CASCADE, related_name='topics', db_index=True
	)
	is_free = models.BooleanField(default=False)
	updated_at = models.DateTimeField(
		default=timezone.localtime,
	)
	created_at = models.DateTimeField(
		default=timezone.localtime,
	)
	description = models.TextField(null=True, blank=True)
	updated_by = models.CharField(
		max_length=100,
		db_index=True,
	)
	number_of_questions = models.IntegerField(default=0)
	image = models.ImageField(upload_to='topic_image', null=True, blank=True)


class Question(models.Model):
	topic = models.ForeignKey(
		to=Topic, on_delete=models.CASCADE, related_name='questions', db_index=True
	)
	# publish_date = models.DateField(default=datetime.now)
	question = models.TextField()
	image = models.ImageField(upload_to='question_image', null=True, blank=True)
	answers = models.JSONField(default=default_answers)
	# multiple_right_answer = models.BooleanField(default=False)
	correct_answer = ArrayField(
		models.CharField(max_length=10, blank=True),
		help_text='Please choose from the above option you entered',
	)
	question_type = models.ForeignKey(
		to=QuestionType, on_delete=models.CASCADE, related_name='questions'
	)
	updated_at = models.DateTimeField(
		default=timezone.localtime(),
	)
	created_at = models.DateTimeField(
		default=timezone.localtime(),
	)
	updated_by = models.CharField(
		max_length=100,
	)
```

Schema:
```python

class TopicUpdateSchema(ModelSchema):
	class Meta:
		model = Topic
		exclude = ['id', 'updated_by', 'updated_at', 'created_at']
		fields_optional = '__all__' <------------------works

class QuestionUpdateSchema(ModelSchema):
	class Meta:
		model = Question
		exclude = ['id', 'updated_by', 'updated_at', 'created_at']
                 fields_optional="__all__" <----------------doesn't work

```

Using `fields_optional='__all__'` in QuestionUpdateSchema does not work but it works in TopicUpdateSchema. Even adding the following fields in QuestionUpdateSchema  didn't work .
```python
fields_optional = [
			'certificate',
			'topic',
[...]

PR #1100, opened 2024-02-28
fix: also set default_factory to None when fields are optional

fk128:
This also set `default_factory = None` when using `fields_optional='__all__'`.

It is related to this issue https://github.com/vitalik/django-ninja/issues/1080


bartoszgrabski:
Fixes https://github.com/vitalik/django-ninja/issues/1019 as well

PR #1102, opened 2024-03-07
Fix TestClient name in testing docs

adriaanwm:
Could also go the other way and do:

```
from ninja.testing import TestClient as NinjaTestClient
```

### DIFF
--- a/docs/docs/guides/testing.md
+++ b/docs/docs/guides/testing.md
@@ -26,7 +26,7 @@ from ninja.testing import TestClient
 
 class HelloTest(TestCase):
     def test_hello(self):
-        client = NinjaTestClient(router)
+        client = TestClient(router)
         response = client.get("/hello")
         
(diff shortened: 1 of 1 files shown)


mikezolo:
The fix is indeed necessary, as the `TestClient` was imported from the `ninja.testing`. 
The solution in the comment above suggesting importing `TestClient as NinjaTestClient` is good too, but the initial fix is more elegant, in my humble opinion.
The only thing I'd consider adding to the request is a bit more explicit description to focus the reviewer on the issue gist.

mikezolo (approved):
The fix is indeed necessary, as the `TestClient` was imported from the `ninja.testing`.
The solution in the comment above suggesting importing `TestClient as NinjaTestClient` is good too, but the initial fix is more elegant, in my humble opinion.
The only thing I'd consider adding to the request is a bit more explicit description to focus the reviewer on the issue gist.

Issue #469, opened 2022-06-11
[BUG] 


eldardamari:
**Describe the bug**
Failing to apply Pydantic `alias_gnerator` config, to convert fields to camelCase in response body.

```python
api = NinjaAPI()


def to_camel(string: str) -> str:
    return ''.join(word.capitalize() for word in string.split('_'))


class AddResult(Schema):
    total_sum: int

    class Config(Schema.Config):
        alias_generator = to_camel
        allow_population_by_field_name = True


@api.get("/add", response=AddResult)
def add(request, a: int, b: int):
    return AddResult(total_sum=a + b)
```

### Expected response: 
```
{
  "totalSum": 3
}
```

### Docs

![image](https://user-images.githubusercontent.com/2379595/173184393-58bde335-dc6f-4376-99ba-96bcb6cc442b.png)

**Versions (please complete the following information):**
 - Python version: [3.9]
 - Django version: [4.0.4]
 - Django-Ninja version: [0.18.0]

vitalik:
Hi @eldardamari 

Pydantic's `by_alias` is not default, you need to enforce it:

```Python
class AddResult(Schema):
    total_sum: int

    class Config:
        alias_generator = to_camel
        allow_population_by_field_name = True


@api.get("/add", response=AddResult, by_alias=True)   # <--- !!! by_alias=True
def add(request, a: int, b: int):
    return {'total_sum': a + b}

```

eldardamari:
Thanks @vitalik!
How can I add `by_alias` to specific api Router? 

```python
app_router = Router()

@app_router.get('/data', by_alias=True) 
def get_data():
  // code

@app_router.get('/') 
def get_index():
  // code



```

vitalik:
I guess you can extend Router class:

```Python
class MyRouter(Router):
    def api_operation(self, *a, **kw):
           kw['by_alias'] = True
           return super().api_operation(*a, **kw)

router = MyRouter()

```

migueltorrescosta:
> I guess you can extend Router class:
> 
> ```python
> class MyRouter(Router):
>     def api_operation(self, *a, **kw):
>            kw['by_alias'] = True
>            return super().api_operation(*a, **kw)
> 
> router = MyRouter()
> ```

Is there a similar solution that would change the default value of `by_alias` to all routers (i.e. to the `NinjaAPI` object` ) rather than applying it on a router by router basis?

vitalik:
@migueltorrescosta 

[...]

PR #1111, opened 2024-03-22
Use property to alias ForeignKey fields instead of Pydantic alias

### LINKED ISSUE #828: [BUG] ForeignKey field in modelSchema do not use the related alias_generator
**Describe the bug**
When using an alias_generator in the config of a `modelSchema` the id's returned for ForeignKey Fields do not use that generator

 - Python version: [ 3.11.4]
 - Django version: [4.1.5]
 - Django-Ninja version: [0.20.0]
 - Pydantic version: [1.10.4]

I have this `Dealer` model

```
class Dealer(AbstractDisableModel, AddressBase, ContactInfoBase):
    ...
    distributor = models.ForeignKey(
        Distributor, on_delete=models.DO_NOTHING, related_name="dealers"
    )
    ...
```

Which adheres to this modelschema

```
class DealerSchema(ModelSchema):
    ...

    class Config(CamelModelSchema.Config):
        ...
```

which uses this schema that converts properties to camelcase

```
class CamelModelSchema(Schema):
    class Config:
        alias_generator = to_camel
        allow_population_by_field_name = True
```

All of this works for the fields directly attached to the instances, but foreignkey field (that end in `_id`) don't seem to be converted. Would it be possible to have the generator adjust ALL fields?

E.g. this is the schema of the response of a dealer instance

image

`distributor_id` is still not camelcased although the rest of the keys are.

pmdevita:
Fixes #828 and helps with #469. Full explanation [here](https://github.com/vitalik/django-ninja/issues/828#issuecomment-2013444804). Apologies if this explanation isn't quite clear, I'm definitely struggling to explain this properly lol.

Currently for ModelSchema, Ninja uses Pydantic aliases to alias between a ForeignKey field's property name as defined by the user, and the attribute name which holds the ID referenced by the foreign key (`author` vs `author_id`). However, this prevents users from using Pydantic's `alias_generator` with ForeignKey field names.

This PR removes the alias and then changes the property name on the Pydantic model to match the attribute name on the Django model (`author_id`), which should allow for data dumped out with `my_schema.model_dump()` to still correctly match up to Django fields as it does now. It then adds `property` fields to alias between the attribute name and the property name, so any accesses on the model, like `my_schema.author`, will get aliased to the real property name, `my_schema.author_id`.

Let me know if there are any changes I need to make. I'm going to also write some tests to polish this off.

### DIFF
--- a/ninja/orm/factory.py
+++ b/ninja/orm/factory.py
@@ -6,7 +6,7 @@ from django.db.models import ManyToManyRel, ManyToOneRel, Model
 from pydantic import create_model as create_pydantic_model
 
 from ninja.errors import ConfigError
-from ninja.orm.fields import get_schema_field
+from ninja.orm.fields import get_schema_field, get_field_property_accessors
 from ninja.schema import Schema
 
 # MAYBE:
@@ -62,12 +62,12 @@ class SchemaFactory:
 
         definitions = {}
         for fld in model_fields_list:
-            python_type, field_info = get_schema_field(
+            field_name, python_type, field_info = get_schema_field(
                 fld,
                 depth=depth,
                 optional=optional_fields and (fld.name in optional_fields),
             )
-            definitions[fld.name] = (python_type, field_info)
+            definitions[field_name] = (python_type, field_info)
 
         if custom_fields:
             for fld_name, python_type, field_info in custom_fields:
@@ -96,6 +96,12 @@ class SchemaFactory:
         # **field_definitions: Any,
         self.schemas[key] = schema
         self.schema_names.add(name)
+
+        for fld in model_fields_list:
+            if fld.is_relation:
+                field_name, prop = get_field_property_accessors(fld)
+                setattr(schema, field_name, prop)
+
         return schema
 
     def get_key(
--- a/ninja/orm/fields.py
+++ b/ninja/orm/fields.py
@@ -108,6 +108,7 @@ def get_schema_field(
     field: DjangoField, *, depth: int = 0, optional: bool = False
 ) -> Tuple:
     "Returns pydantic field from django's model field"
+    name = field.name
     alias = None
     default = ...
     default_factory = None
@@ -118,14 +119,14 @@ def get_schema_field(
 
     if field.is_relation:
         if depth > 0:
-            return get_related_field_schema(field, depth=depth)
+            return name, *get_related_field_schema(field, depth=depth)
 
         internal_type = field.related_model._meta.pk.get_internal_type()
 
         if not field.concrete and field.auto_created or field.null:
             default = None
 
-        alias = getattr(field, "get_attname", None) and field.get_attname()
+        name = getattr(field, "get_attname", None) and field.get_attname()
 
         pk_type = TYPES.get(internal_type, int)
         if field.one_to_many or field.many_to_many:
@@ -165,6 +166,7 @@ def get_schema_field(
     title = title_if_lower(field.verbose_name)
 
     return (
+        name,
         python_type,
         FieldInfo(
             default=default,
@@ -199,3 +201,12 @@ def get_related_field_schema(field: DjangoField, *, depth: int) -> Tuple[OpenAPI
             title=title_if_lower(field.verbose_name),
         ),
     )
+
+
+@no_type_check
+def get_field_property_accessors(field: DjangoField):
+    attribute_name = getattr(field, "get_attname", None) and field.get_attname()
+    property_name = field.name
+    fget = lambda self: getattr(self, attribute_name)
+    fset = lambda self, value: setattr(self, attribute_name, value)
+    return property_name, property(fget, fset)


vitalik:
Hi @pmdevita 

Yeah I guess it worth trying - just wondering will ti work for case like 

```Python
class SomeSchema(ModelSchema):
    fkfield_id: int
    
    class Meta:
           model = Some
           fields = ['id', 'fkfield']
```

pmdevita:
Yeah I've been kinda thinking over whether this is going to cause some kind of backwards compatibility problem. But I tested the current master branch and it looks like doing this with a schema already causes trouble.

```python
class Author(models.Model):
    id = models.AutoField(primary_key=True)
    name = models.CharField(max_length=50)

    class Meta:
        app_label = "tests"

class Book(models.Model):
    id = models.AutoField(primary_key=True)
    name = models.CharField(max_length=100)
    author = models.ForeignKey(Author, on_delete=models.CASCADE)

    class Meta:
        app_label = "tests"

class BookSchema(ModelSchema):
    author_id: int = Field(...)
    class Meta:
        model = Book
        fields = "__all__"


pprint(BookSchema.json_schema())
test = BookSchema(author_id=1, name="asdfioj")
print(test)
```
    
outputs
    
```
{'properties': {'author_id': {'title': 'Author', 'type': 'integer'},
                'id': {'anyOf': [{'type': 'integer'}, {'type': 'null'}],
                       'title': 'Id'},
                'name': {'maxLength': 100, 'title': 'Name', 'type': 'string'}},
 'required': ['author_id', 'name', 'author_id'],
 'title': 'BookSchema',
 'type': 'object'}
author_id=1 id=None name='asdfioj' author=1
```

The schema seems a bit funny, there's only one `author_id` property now and it's mentioned twice in `required`. We can also see that setting the `author_id` also sets both.

Same thing here with models.

```python
author_test = Author(name="J. R. R. Tolkien", id=1)
model_test = Book(author=author_test, name="The Hobbit", id=1)
schema_test = BookSchema.from_orm(model_test)
print(schema_test)
```

outputs

```
author_id=1 id=1 name='The Hobbit' author=1
```

It might be reasonable to assume that because this behavior just duplicates data, it probably isn't useful and it's unlikely anyone would be relying on it. 

Here is the output from that test on my branch for comparison.

```
{'properties': {'author_id': {'title': 'Author Id', 'type': 'integer'},
                'id': {'anyOf': [{'type': 'integer'}, {'type': 'null'}],
                       'title': 'Id'},
                'name': {'maxLength': 100, 'title': 'Name', 'type': 'string'}},
 'required': ['author_id', 'name'],
 'title': 'BookSchema',
 'type': 'object'}
author_id=1 id=None name='asdfioj'
author_id=1 id=1 name='The Hobbit'
```

pmdevita:
I left the class `Config` in the docs but I just realized Pydantic deprecated that, is `model_config` now the recommended way to go for Ninja?

pmdevita:
Anything else I can do to help move this along?

pmdevita:
@vitalik Could you take a look at this again and let me know if I need to fix anything? I'd really like to get this merged since I'm using to_camel for all of my API schemas. Thank you!

Tatsh:
@pmdevita does this fix this this issue?

```python
class SomeSchema(Schema):
    field: datetime = Field(None, alias='user.date')
    
class AnotherSchema(Schema):
    some: SomeSchema = None
```

Outputs:

```json
{
  "some": {"user.date": ""}
}
```

But we want:

```json
{
  "some": {"field": ""}
}
```

pmdevita:
It doesn't, this solves an issue specific to foreign keys and alias generators. I'm not sure why you're aliasing like that but maybe AliasPath and AliasChoice can help you if you're trying to access a property on the data being validated

SaschaKrug:
Is there any reason this hasn't been merged? The handling of alias generators is very buggy currently with foreign keys and it looks like this PR would fix it?

pmdevita:
@vitalik Is there anything you'd like to me to change with this? It would be nice to have it merged

pmdevita:
I have made a mistake by leaking unrelated commits from a fork with my rebase earlier, I think it also accidentally backlinked some stuff too in the repo. I'm really sorry about that, I'll try to see if I can clean it up some how.

pmdevita:
Some extra context I wrote up since it's been a while and I don't remember this all to well anymore.

Sort of an example just with Pydantic, the old way would look like this

```python
class MySchema(ModelSchema):
    author_id: int = Field(..., alias="author")  # generated by ModelSchema
    
    class Meta:
        model = Book
        fields = ["author"]
```

which would of course then cause a problem with use of automatic alias generation. The new way is effectively equivalent to

```python
class MySchema(ModelSchema):
    author_id: int = Field(...)  # generated by ModelSchema
    
    @property                        # generated by ModelSchema
    def author(self):
        return author_id

    @author.setter                 # generated by ModelSchema
    def author(self, value):
        self.author_id = value
    
    class Meta:
        model = Book
        fields = ["author"]
```        

which now frees up alias to be changed. In all testing I've done, this is effectively supersetting the old functionality, as dumping the model would make it use "author_id" anyways, and manually defining an "author_id" field with an "author" was probably going to get you in trouble anyways.

In the time since I've written this, I ran into actually that exact issue vitalik mentioned here https://github.com/vitalik/django-ninja/pull/1111#issuecomment-2014514028. I would propose to fix this we simply don't attach the @property functions if the user was already trying to put something there themselves. Manual property definitions should be the only source of conflict.

yetty:
Hi @vitalik, any updates on this pull request? It seems like there are several people experiencing issues related to alias generators and foreign keys. Just curious if there's anything blocking this from being merged. Thanks!

PR #1112, opened 2024-03-26
DRAFT: Naively convert from Config class to ConfigDict

jrocketfingers:


jrocketfingers:
I assume that this can't be merged until support for both Pydantic 1.x is dropped? Feel free to close this if you have different plans for addressing the deprecation, I've done this as an exercise for myself, mostly to understand if my schemas introduce these warnings.

vitalik:
Hi @jrocketfingers 

well generally pydantic 1.x is not supported in django-ninja 1.x

My idea was to migrate `Class` to `Meta` naming, but first Meta should automatically set pydantic.model_config 
(That way django world only works with Meta on everything, Forms, Models, Schemas, Filters)

jrocketfingers:
> My idea was to migrate Class to Meta naming

Ah, that makes sense.

Do you happen to know about Pydantic's motivation for moving away from Config classes? Might it be worthwhile to lean into Pydantic's way of doing things?

The more I work with Django Ninja the more I feel like it's not _really_ a part of the Django ecosystem. Many apps/libs create Django views, not Ninja views; auth libraries work with Django auth, not Ninja auth and so on. 

Am I mistaken in thinking this? ORM/urlpatterns seems to be one of the few contact points with Django proper.

Let me know if I should move the philosophical discussion elsewhere 😅

jrocketfingers:
Gotcha, in that case I'll close this if it's not salvageable as-is. Feel free to reopen it and ping me if you want this pushed through to completion.

gertsteyn:
1.5 years later, maybe time to get rid of the PydanticDeprecatedSince20 warning, Pydantic 1 is very old now ;-)

PR #1126, opened 2024-04-12
Add Generic Typing for `decorate_view` to resolve Pyright Issues

max-muoto:
For projects using Django Ninja that utilize [Pyright](https://github.com/microsoft/pyright), the typing for `decorate_view` leads to a `reportUntypedFunctionDecorator` error.

Let's take this example from the documentation:

```python
from django.views.decorators.cache import cache_page

@api.get("/test")
@decorate_view(cache_page(5)) # Untyped function decorator obscures type of function
def test_view(request):
    return str(datetime.now())
```

By having `decorate_view` utilize generics, we can can avoid this. I also went about generally cleaning up the callable types, `Callable[..., Any]` is more descriptive than `Callable`. In Pyright the former would be preferred to avoid unknowns, as `Callable == Callable[..., Unknown]` instead of being `Any`. There have been times where I've imported private functions from ninja internals, while not often, it would be nice to avoid unknowns in these cases.

### DIFF
--- a/ninja/decorators.py
+++ b/ninja/decorators.py
@@ -1,7 +1,8 @@
 from functools import partial
-from typing import Callable, Tuple
+from typing import Any, Callable, Tuple
 
 from ninja.operation import Operation
+from ninja.types import TCallable
 from ninja.utils import contribute_operation_callback
 
 # Since @api.method decorator is applied to function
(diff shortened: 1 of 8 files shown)


max-muoto:
@vitalik Was curious if there was anything else I could to help get this merged!

PR #1127, opened 2024-04-12
Fix `paginate` Decorator Typing

max-muoto:
Paginating your endpoint transforms the collection return type into one that returns a dict, however, this isn't reflected in the final type.

Let's take this example:

```python
@api.get("/invalid1", response={200: None})
@paginate
def invalid1(request: HttpRequest) -> list[Any]: 
    return [1, 2, 3]

reveal_type(invalid1) # Type of "invalid1" is "(...) -> Unknown
```

This also leads to typing errors when passing in a pagination class (with Pyright):

```python
@api.get("/invalid1", response={200: None})
 @paginate(PageNumberPagination) # Untyped function decorator obscures type of function; ignoring decorator
def invalid1(request: HttpRequest) -> list[Any]: 
    return [1, 2, 3]
```

This PR sets up overloading for `paginate` so that type can be correctly inferred in both cases. Operating  under the assumption that assumption that ninja already makes that `paginate_queryset` returns `dict[str, Any]`

For our first example, we now get:

```python
@api.get("/invalid1", response={200: None})
@paginate
def invalid1(request: HttpRequest) -> list[Any]: 
    return [1, 2, 3]

reveal_type(invalid1) # (HttpRequest) -> Dict[str, Any]
```

## Other Changes

- While unlikely, it's not enforced that you paginate a queryset, technically any collection with `len` is supported. Therefore, I switched some of the `QuerySet` types to `Sequence` to better reflect this.
- Added overload decorated functions to be excluded from coverage requirements, since they're purely for typing.

### DIFF
--- a/ninja/pagination.py
+++ b/ninja/pagination.py
@@ -1,20 +1,40 @@
 import inspect
 from abc import ABC, abstractmethod
 from functools import partial, wraps
-from typing import Any, Callable, List, Optional, Tuple, Type
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    List,
+    Optional,
+    Sequence,
+    Tuple,
+    Type,
+    TypeVar,
+    Union,
+    overload,
+)
 
-from django.db.models import QuerySet
+from django.db import models
 from django.http import HttpRequest
 from django.utils.module_loading import import_string
+from typing_extensions import Concatenate, ParamSpec, TypeAlias, TypeGuard
 from typing_extensions import get_args as get_collection_args
 
 from ninja import Field, Query, Router, Schema
 from ninja.conf import settings
-from ninja.constants import NOT_SET
+from ninja.constants import NOT_SET, NOT_SET_TYPE
 from ninja.errors import ConfigError
 from ninja.operation import Operation
 from ninja.signature.details import is_collection_type
 from ninja.utils import contribute_operation_args, contribute_operation_callback
+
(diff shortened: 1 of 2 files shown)


ninja/pagination.py
```diff
@@ -109,37 +152,53 @@ def my_view(request):
...
-    def view_with_pagination(request: HttpRequest, **kwargs: Any) -> Any:
+    def view_with_pagination(
+        request: Req, *args: P.args, **kwargs: P.kwargs
+    ) -> Dict[str, Any]:
         pagination_params = kwargs.pop("ninja_pagination")
         if paginator.pass_parameter:
             kwargs[paginator.pass_parameter] = pagination_params
 
-        items = func(request, **kwargs)
+        items = func(request, *args, **kwargs)
```
max-muoto, line 201: To properly utilize `ParamSpec` (which lets us preserve the call signature), we need to pass in `*args`. I'm guessing this wouldn't be problematic, but if will be, I can rethink the approach here.

vitalik:
Hi @max-muoto 

Thank you for contribution
Could you fix conflicts with the latest version ?

max-muoto:
> Hi @max-muoto
> 
> Thank you for contribution Could you fix conflicts with the latest version ?

I'll have to do a bit of reworking with the support for async pagination, but I'll ping you once this is ready again, I'll move it back to a draft in the mean-time.

PR #1150, opened 2024-05-03
feat(django_getter): use function cache to increase performance

yorickr:
Hi! 

After using ninja a lot, I noticed some performance benefit could be made in DjangoGetter, by caching the actual function being called and using a lookup, instead of doing up to 4 checks each time.

This MR implements this simple cache. 
I didn't specifically add tests for this, as schema is tested very well throughout the rest of the tests. Let me know if you'd like to see this differently.

Let me know what you think!

Some data:
```
Old code: Average time across 20 runs of 100k nested objects is 5.516484143049911
New code: Average time across 20 runs of 100k nested objects is 4.045871758400063
This is 1.5s saved, or 27% saved.

Old code: Average time across 20 runs of 10k nested objects is 0.4554609017498933
New code: Average time across 20 runs of 10k nested objects is 0.35757346044997573
This saved 22%.
```

### DIFF
--- a/ninja/schema.py
+++ b/ninja/schema.py
@@ -18,14 +18,15 @@ dotted attributes and resolver methods. For example::
 
 """
 
+from __future__ import annotations
+
 import warnings
+from functools import partial
 from typing import (
     Any,
     Callable,
-    Dict,
-    Type,
+    ClassVar,
     TypeVar,
-    Union,
     no_type_check,
 )
 
@@ -39,6 +40,7 @@ from pydantic.functional_validators import ModelWrapValidatorHandler
 from pydantic.json_schema import GenerateJsonSchema, JsonSchemaValue
 from typing_extensions import dataclass_transform
 
+from ninja.constants import NOT_SET
 from ninja.signature.utils import get_args_names, has_kwargs
 from ninja.types import DictStrAny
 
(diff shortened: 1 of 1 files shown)


yorickr:
Hi! Is this something you would like to see included? Let me know

yorickr:
@vitalik is there any interest in merging or introducing this? If not, I can close it. Please let me know :)

vitalik:
Hi @yorickr 
Thank you for PR - I am looking into this (just afraid can be some edge cases with cache key)
On the other hand there might be a chance that pydantic will give us a nicer hooks (as DjangoGetter is actually a bit dirty workaround) and that may speed things up even more

yorickr:
Thanks for your reply :) 
From what I can tell, the cache key should work fine for everything. Perhaps not for builtins, as they do not have a name. But I am unsure whether they can even end up being fed in through the constructor though.
For what it's worth, this has seen some production use and hasn't been giving us issues yet.

I agree with DjangoGetter being a bit of a dirty workaround. Until we get nicer hooks, I think this is a decent approach.

yorickr-sendcloud:
I've been running this exact version for about 10 months now, and have yet to see issues with it. But I am closing this, as there doesn't seem to be interest in merging this approach. Feel free to contact me to re-open.

yorickr:
I've been running this exact version for about 10 months now, and have yet to see issues with it. But I am closing this, as there doesn't seem to be interest in merging this approach. Feel free to contact me to re-open.

Issue #960, opened 2023-11-24
Calculated fields are not visible in the schema 


leiserfg:
The schema of the models uses `validation` instead of `serialization` and that causes the computed fields not to be shown (see https://docs.pydantic.dev/latest/api/json_schema/#pydantic.json_schema.JsonSchemaMode).

aurbano:
@vitalik is there a known workaround for this, or a plan to address?

I can try to send a PR if you point me in the direction of where to fix

vitalik:
@aurbano sure - go ahead

PR #1162, opened 2024-05-11
Use serialization aliases for responses in OpenAPI schema

camuthig:
When generating JSON schemas from the Pydantic types, the `mode` must be set in order for the correct aliases to be used. The default is to generate the schema for validation mode, which means that for building response schemas, the mode needs to be explicitly set to serialization.

This fixes issues I brought up in discussion #1132

### DIFF
--- a/ninja/openapi/schema.py
+++ b/ninja/openapi/schema.py
@@ -4,6 +4,8 @@ import warnings
 from http.client import responses
 from typing import TYPE_CHECKING, Any, Dict, Generator, List, Optional, Set, Tuple
 
+from pydantic.json_schema import JsonSchemaMode
+
 from ninja.constants import NOT_SET
 from ninja.operation import Operation
 from ninja.params.models import TModel, TModels
@@ -207,6 +209,7 @@ class OpenAPISchema(dict):
         model: TModel,
         by_alias: bool = True,
         remove_level: bool = True,
+        mode: JsonSchemaMode = "validation",
     ) -> Tuple[DictStrAny, bool]:
         if hasattr(model, "__ninja_flatten_map__"):
             schema = self._flatten_schema(model)
@@ -215,6 +218,7 @@ class OpenAPISchema(dict):
                 ref_template=REF_TEMPLATE,
                 by_alias=by_alias,
                 schema_generator=NinjaGenerateJsonSchema,
+                mode=mode,
             ).copy()
 
         # move Schemas from definitions
@@ -284,7 +288,7 @@ class OpenAPISchema(dict):
             if model not in [None, NOT_SET]:
                 # ::TODO:: test this: by_alias == True
                 schema = self._create_schema_from_model(
-                    model, by_alias=operation.by_alias
+                    model, by_alias=operation.by_alias, mode="serialization"
                 )[0]
                 details[status]["content"] = {
                     self.api.renderer.media_type: {"schema": schema}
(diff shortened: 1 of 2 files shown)


pmdevita:
This might be the same thing as https://github.com/vitalik/django-ninja/pull/1139

camuthig:
@pmdevita thanks for calling that out. There is some duplication here in that both pulls depend on leveraging the `mode` argument in `model_json_schema`. However, the other issue/pull request are also tackling a more nuanced issue around definition of "required" field.

I'm open to closing the pull request, but considering the other one is also facing some open questions of backwards compatibility, I don't want to slow down this fix because of those unrelated topics.

nstonic:
@camuthig So, what about this PR?

camuthig:
@nstonic I'm not sure what the situation here is. @vitalik would you consider this pull request as an simpler version of the referenced #1139, that one is tackling a much deeper problem. This pull request has a smaller scope and only makes sure that the alias configurations are respected in the JSON schema generation. I think the other PR can still be considered as a progression of what this one accomplishes.

tu-pm:
This also fixes the problem of computed fields missing from the schema, as reported by issue  #1292 and #960

oubeichen:
@vitalik any update on this?

rayansostenes:
Had to solve this at work, here's the monkey patch I applied to get make it work.

```python
# pyright: basic
from http.client import responses
from typing import Any

import ninja
import pydantic
from ninja.constants import NOT_SET
from ninja.openapi.schema import REF_TEMPLATE
from ninja.openapi.schema import OpenAPISchema
from ninja.schema import NinjaGenerateJsonSchema
from pydantic.json_schema import JsonSchemaMode


def _create_schema_from_model(
    self,
    model: pydantic.BaseModel,
    by_alias: bool = True,  # noqa: FBT001, FBT002
    remove_level: bool = True,  # noqa: FBT001, FBT002
    mode: JsonSchemaMode = "validation",
) -> tuple[dict[str, Any], bool]:
    if hasattr(model, "__ninja_flatten_map__"):
        schema = self._flatten_schema(model)
    else:
        schema = model.model_json_schema(
            ref_template=REF_TEMPLATE,
            by_alias=by_alias,
            schema_generator=NinjaGenerateJsonSchema,
            mode=mode,
        ).copy()

    # schema = _rewrite_schema(self.schemas, schema, mode)

    if schema.get("$defs"):
        self.add_schema_definitions(schema.pop("$defs"))

    if remove_level and len(schema["properties"]) == 1:
        name, details = next(iter(schema["properties"].items()))

        required = name in schema.get("required", {})
        return details, required
    return schema, True


def _responses(self, operation: Any) -> dict[int, dict[str, Any]]:
    assert bool(operation.response_models), f"{operation.response_models} empty"

    result = {}
    for status, model in operation.response_models.items():
        if status == Ellipsis:
            continue  # it's not yet clear what it means if user wants to output any other code

        description = responses.get(status, "Unknown Status Code")
        details: dict[int, Any] = {status: {"description": description}}
        if model not in [None, NOT_SET]:
            schema = self._create_schema_from_model(
                model,
                by_alias=operation.by_alias,
                mode="serialization",
            )[0]
            details[status]["content"] = {self.api.renderer.media_type: {"schema": schema}}
        result.update(details)

    return result


def apply_ninja_monkey_patch():
    version_tuple = tuple(map(int, ninja.__version__.split(".")))
    if version_tuple > (1, 3, 0):
        msg = "This monkey patch is only tested for ninja<=1.3.0"
        raise ImportError(msg)
    OpenAPISchema._create_schema_from_model = _create_schema_from_model  # noqa: SLF001
    OpenAPISchema.responses = _responses
```

camuthig:
@vitalik bumping this back to your attention. Is there any chance of getting this review and merged in?

sghng:
This might be related to #1485, which is about using Pydantic builtin `model_dump_json` method to serialize responses. Currently it seems like the the Pydantic configs are not respected in serialization.

camuthig:
This PR is no longer necessary now that https://github.com/vitalik/django-ninja/pull/1606 is merged

Issue #1094, opened 2024-02-23
add dataclass_transform to enable better typehints


landeholt:
**Is your feature request related to a problem? Please describe.**
When you inherit modelmetaclass, the dataclass_transform decorator gets lost, resulting in LSP not being able to source fields for schemas

**Describe the solution you'd like**

add @dataclass_transformer(kw_only_default=True, field_specifiers=(Field,)) to the new metaclass here: https://github.com/vitalik/django-ninja/blob/08af4b9b571c92a7d14edaeb53cf54bb673bea0d/ninja/schema.py#L149

Can either be sourced from pydantic or dataclasses. Field is the schema field.

Issue #1053, opened 2024-01-17
Django-ninja doesn't work with mypy?


JinLisek:
Hi,

I'm trying to start a new project using django-ninja.

I installed mypy, wanted to see how it works with django-ninja... But somehow I can't make it work together. Mypy correctly checks Pydantic models, but it doesn't work with Django-ninja schemas (even though they inherit pydantic.BaseModel).

Here is a little example with pydantic and django-ninja models instantiated incorrectly:
```python
import ninja
import pydantic


class N(ninja.Schema):
    v: int


class P(pydantic.BaseModel):
    v: int


n = N(v="z")
p = P(v="z")

```

Mypy notices an error when creating P, but doesn't whine about incorrect creation of N:
```
aaa.py:14: error: Argument "v" to "P" has incompatible type "str"; expected "int"  [arg-type]
Found 1 error in 1 file (checked 1 source file)
```

What am I doing wrong? :|

I'm using:
python 3.12.1
django-ninja 1.1.0
pydantic 2.5.3
mypy 1.8.0

vitalik:
related #1050

Issue #1050, opened 2024-01-17
Missing editor hints when creating new instance


torx-cz:
Hello

I would like to ask if it is possible to "turn on" (somehow) the ability to see editor hints when creating new instances of Schemas.

For example, see this code:

```python
import pydantic
import ninja

class PydanticModel(pydantic.BaseModel):
    a: int = 0
    b: str

class NinjaSchema(ninja.Schema):
    a: int = 0
    b: str

PydanticModel()
NinjaSchema()
```

When i write this code in VS Code i can see this for `PydanticModel()` (when hover over + red line under it, signaling something is wrong!)

![pydantic_model](https://github.com/vitalik/django-ninja/assets/22475728/cd87e97e-f99a-484d-b1df-51cbd105192d)

But for `NinjaSchema()` i see only this:

![ninja_schema](https://github.com/vitalik/django-ninja/assets/22475728/6da78f4d-924f-4986-b708-0a518ca3cf86)

---

My question is: **Is there some way to turn on the same behavior for NinjaSchemas like it is for PydanticModel?**
Namely:
- when hover over - to be able to see which params need to be defined (not just  `**data: Any`) 
- red line under when something is wrong

Thank you very much!

vitalik:
Hi @torx-cz 

yeah indeed - I did not notice that before - and it is interesting to find out what's the trick (as Schema is just inherited from the BaseModel... maybe metaclass helps here)

PR #1165, opened 2024-05-15
Fix Ninja Schema Typing

max-muoto:
Fix Ninja schema typing by adding a dataclass transformer, as specified in [PEP 681](https://peps.python.org/pep-0681/). This transformer matches that of Pydantic's model metaclass.

Now, instead of Ninja schemas being typed as `**kwargs: Any`, one will get static type-checking on each field.

Fixes: https://github.com/vitalik/django-ninja/issues/1094, https://github.com/vitalik/django-ninja/issues/1050, https://github.com/vitalik/django-ninja/issues/1053

Example:

CleanShot 2024-05-14 at 23 28 35@2x

### DIFF
--- a/ninja/schema.py
+++ b/ninja/schema.py
@@ -19,7 +19,15 @@ dotted attributes and resolver methods. For example::
 """
 
 import warnings
-from typing import Any, Callable, Dict, Type, TypeVar, Union, no_type_check
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Type,
+    TypeVar,
+    Union,
+    no_type_check,
+)
 
 import pydantic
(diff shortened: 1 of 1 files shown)


ninja/schema.py
```diff
@@ -29,6 +37,7 @@ def resolve_name(obj):
 from pydantic._internal._model_construction import ModelMetaclass
 from pydantic.functional_validators import ModelWrapValidatorHandler
 from pydantic.json_schema import GenerateJsonSchema, JsonSchemaValue
+from typing_extensions import dataclass_transform
```
max-muoto, line 40: Until the minimum is 3.11, we'll have to import from typing extensions.

Issue #1045, opened 2024-01-12
using patch with None, and optional fields


cltrudeau:
Hello there,

I don't normally bother with a PATCH operation, but need to this time and so am running into something I haven't before. I've built a ModelSchema using the Meta attribute `fields_optional="__all__"`, but the associated payload includes all keys, even if they're aren't sent in the request. Keys that aren't sent are set to None. This is problematic, as there is no way to differentiate a client wanting to set a value to None from a value not being sent.

Am I doing it wrong?

While on the topic, it'd be great to have a shortcut for creating a schema based on another one, but with a change to its attributes. For example, having a Car model, I would have a CarInSchema for the PUT, and a CarPatchSchema for the PATCH. Rather than copy-paste the class, it'd be great to have a way to create CarPatchSchema based on CarInSchema, but with the `fields_optional` attribute set. I attempted to use inheritance to accomplish this, but it didn't seem to work. Probably something to do with the inner class and how the metaclass is creating the class.

(Actually, on that topic, it'd be great to be able to just add a field as well. 90% of the time my "In" and "Out" Schemas only differ by the ID field, so having a base "In" Schema, and an inheriting Out and Patch Schema where only the changes were needed would be ideal)

Qadosch:
Better late than never
https://django-ninja.dev/reference/operations-parameters/#exclude_unset

you could use it like this:
```pathon
@api.patch("/{id}", response=RespSchema)
def patch_view(request, employee_id, body: ReqSchema):
    employee = get_object_or_404(Employee, id=employee_id)
    filtered_dict = body.dict(exclude_unset=True)
    for attr, value in filtered_dict.items():
            setattr(employee, attr, value)
```

`exclude_unset` seems to filter the dict for not set items

Qadosch:
As far as I understand, the concept of Schemas is to deal with unset values, is by setting them. Pydantic deals with this this way and by extension FastAPI and DjangoNinja.

It feels like the PATCH method is not welcome lol

cltrudeau:
Might be able to be handled with a function instead. A helper method that takes the Schema specifying the possible fields/validation, and the request, and once the data has been verified normalizes it for changed fields only. Essentially, doing the Schema work first, then removing fields that weren't in the request body.

Could also be a method on the class, in addition to `.as_dict()`, you could have `as_patch_dict()`.

[...]

PR #1173, opened 2024-05-21
Docs: Adds an example to show how exclude_unset works

benjaoming:
Fixes: #1045

I believe the right thing to do here is to iterate over fields to patch and exclude all the unsupplied fields. Preceding documentation already shows how to use `fields_optional`, but it remains somewhat confusing what to do next.


horizon365:
good example

vitalik:
Thank you @benjaoming

PR #1178, opened 2024-05-28
[FIX] Updated Swagger doc to use csrftoken cookie value for custom csrf token header

viralj:
Added fix for django_auth where custom csrftoken header doesn't get updated when csrftoken cookie is set. This fix/update will allow swagger ui to test secure endpoint with built in `django_auth`. The fix simply looks into cookie first and if cookie is set, it will automatically use that instead of the csrf token which gets generated by django.


pjtpj:
This same change needs to be made to swagger.html for those of using django_auth without the CDN version (by adding 'ninja' to INSTALLED_APPS).

PR #1208, opened 2024-06-26
Throttling

vitalik:
Throttles allows to control the rate of requests that clients can make to an API. Django Ninja allows to set custom throttlers globally (across all operations in NinjaAPI instance), on router level and each operation individually.

Django Ninja’s throttling feature is pretty much based on what Django Rest Framework (DRF) uses, which you can check out [here](https://www.django-rest-framework.org/api-guide/throttling/). So, if you’ve already got custom throttling set up for DRF, there’s a good chance it’ll work with Django Ninja right out of the box. They key difference is that you need to pass initialized Throttle objects instead of classes (which should give a better performance)

### Usage

**Global**

The following example will limit unauthenticated users to only 10 requests per second, while authenticated can make 100/s

```Python
from ninja.throttling import AnonRateThrottle, AuthRateThrottle

api = NinjaAPI(
    throttle=[
        AnonRateThrottle('10/s'),
        AuthRateThrottle('100/s'),
    ],
)
```

!!! tip
    `throttle` argument accepts single object and list of throttle objects

**Router level**

Pass `throttle` argument either to `add_router` function

```Python
api = NinjaAPI()
...

api.add_router('/sensitive', 'myapp.api.router', throttle=AnonRateThrottle('100/m'))
```

or directly to init of the Router class:

```Python
router = Router(..., throttle=[AnonRateThrottle('1000/h')])
```

**Operation level**

If `throttle` argument is passed to operation - it will overrule all global and router throttles:

```Python
from ninja.throttling import UserRateThrottle

@api.get('/some', throttle=[UserRateThrottle('10000/d')])
def some(request):
    ...
```


vitalik:
Hi @eadwinCode - JFYI throttling was a very requested feature by lot of users - so this implementation will probably land in the next release (making django-ninja-extra throttling a bit broken I guess)

PR #1212, opened 2024-06-28
Allow generic filter types

eugenenelou:
This MR allows creating custom filter type with pydantic to reuse across filter classes, by making the field_name implicit.

like:
```py
IContainsField = Annotated[Optional[str], Field(None, q='__icontains')]

    name: IContainsField


eugenenelou:
fixed the ruff and mypy errors.

Issue #1223, opened 2024-07-10
[BUG] If validate assignment is true, the value will not change.


milksys:
If validate assignment is true, the value will not change.

this is sampel code

**Describe the bug**
```python
from ninja import Schema

class Model(Schema):
    name: str
    is_bool: bool

    model_config = ConfigDict(validate_assignment=True)


model = Model(name="NewJeans",  is_bool=False)

print(model)  # => name="NewJeans" is_bool=False

model.name= "Ive"
model.is_bool = True

# Not change value
print(model)  # => name="NewJeans" is_bool=False
```

In the previous 1.1.0 version, that PR threw an error, but now the value does not change (https://github.com/vitalik/django-ninja/pull/1024)

**Versions (please complete the following information):**
 - Python version: 3.11.1
 - Django version: 5.0.7
 - Django-Ninja version: 1.2.1
 - Pydantic version: 2.8.2

austinpgraham:
I've been debugging this over the past couple of hours, and seems to be related how pydantic updates fields after they've been validated; appears pydantic-core is updating the __dict__ property directly: https://github.com/pydantic/pydantic-core/blob/0e6b377d2bef52e744deff22314269c8751f7b00/src/validators/dataclass.rs#L368 (I'm not super familiar with the pydantic codebase so I may be looking at the wrong place)

However, I did validate this by debugging through the existing test added in the PR you linked @milksys. 
```
tests/test_schema.py 
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB set_trace (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /Users/me/django-ninja/ninja/schema.py(225)_run_root_validator()
-> return handler(values)
(Pdb) c

[...]

PR #1232, opened 2024-07-15
Fix unchanging field bug when validate assignment true

austinpgraham:
# Problem
This PR addresses a bug reported here, where if validate_assignment is set to True on a schema, then Pydantic will correctly validate the new values, but the schema object will not persist the new value.

# RCA
This is caused by the fact that [pydantic-core sets these new values on the object (once validated) directly on the object __ dict __](https://github.com/pydantic/pydantic-core/blob/0e6b377d2bef52e744deff22314269c8751f7b00/src/validators/dataclass.rs#L368). As it was, this means that the new value will be set onto the DjangoGetter object, rather than the schema object being validated.

# Solution
Added an extra condition in the root validator of a Schema object that will call `handler` on the original values before the conversion to a DjangoGetter to ensure the new value is persisted properly on the original object. Comment is also updated to reflect this change.

# Testing
Added assertions into an existing test that a reassigned value is persisted when validate_assignment = True, and an additional test to test for the lack of ValidationError when validate_assignment is False or None.

# Open Questions
This seems odd in the following way: I understand the use of DjangoGetter of course to handle resolvable fields. However, it not behaving like a dictionary is causing some churn where we now have to validate the object twice in certain conditions just to ensure fields are persisted/validated properly. I don't know immediately if there's a better solution, or if there's background here I don't know. Welcome to feedback.

### DIFF
--- a/ninja/schema.py
+++ b/ninja/schema.py
@@ -215,9 +215,13 @@ class Schema(BaseModel, metaclass=ResolverMetaclass):
     def _run_root_validator(
         cls, values: Any, handler: ModelWrapValidatorHandler[S], info: ValidationInfo
     ) -> Any:
-        # when extra is "forbid" we need to perform default pydantic validation
-        # as DjangoGetter does not act as dict and pydantic will not be able to validate it
-        if cls.model_config.get("extra") == "forbid":
+        # If Pydantic intends to validate against the __dict__ of the immediate Schema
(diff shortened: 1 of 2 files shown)


austinpgraham:
@vitalik potential fix to issue mentioned here: https://github.com/vitalik/django-ninja/issues/1223

vitalik:
Thank you

PR #1249, opened 2024-08-02
Fix primary key always being marked as nullable in schema

### LINKED ISSUE #1227: [BUG] Primary keys are always marked as nullable
**Describe the bug**
As a result of #1181, all primary keys are now marked as nullable in ModelSchema, regardless of whether the field is set to nullable or not. I'm not sure why this was done, but I would expect it to defer to the field's own nullable setting instead if for some reason you have a nullable primary key. If the intention was to make this optional for request schemas, I think it should be marked in `fields_optional` instead in the Meta class.

This is a result of these lines here in fields.py

https://github.com/vitalik/django-ninja/blob/8a499928dafb68626e1bcf5c231f23933e4ec241/ninja/orm/fields.py#L153-L155

 - Python version: 3.12
 - Django version: 5.0.7
 - Django-Ninja version: 1.2.1
 - Pydantic version: 2.8.2

pmdevita:
Fixes #1227

A regression occurred with #1181 that made it so primary keys were always marked as nullable in a Schema. This fixes that regression and then also fixes the tests that were incorrectly expecting it that way.

I have no idea why you would lol, but if for some reason you had a primary key field that was marked as nullable on the model, then the Schema should respect that, so I added a new test to ensure that happens.

This will change the default behavior of the schema generation, but it just restores it to the correct pre-1.2.0 behavior for primary keys. For any users that want their primary key to be nullable/optional, they should be either setting the model field to `null=True` or marking the field as optional with `optional_fields` on their ModelSchema.

### DIFF
--- a/ninja/orm/fields.py
+++ b/ninja/orm/fields.py
@@ -150,7 +150,7 @@ def get_schema_field(
         internal_type = field.get_internal_type()
         python_type = TYPES[internal_type]
 
-        if field.primary_key or blank or null or optional:
+        if blank or null or optional:
             default = None
             nullable = True
 
--- a/tests/test_orm_metaclass.py
+++ b/tests/test_orm_metaclass.py
@@ -46,10 +46,10 @@ def test_simple():
         "title": "SampleSchema2",
         "type": "object",
         "properties": {
-            "id": {"anyOf": [{"type": "integer"}, {"type": "null"}], "title": "ID"},
+            "id": {"title": "ID", "type": "integer"},
             "firstname": {"title": "Firstname", "type": "string"},
         },
-        "required": ["firstname"],
+        "required": ["id", "firstname"],
     }
 
 
@@ -119,9 +119,9 @@ def test_optional():
             fields = "__all__"
             fields_optional = "__all__"
 
-    assert OptSchema.json_schema().get("required") == ["extra"]
+    assert OptSchema.json_schema().get("required") == ["id", "extra"]
     assert OptSchema.json_schema()["properties"] == {
-        "id": {"anyOf": [{"type": "integer"}, {"type": "null"}], "title": "ID"},
+        "id": {"title": "ID", "type": "integer"},
         "title": {"anyOf": [{"type": "string"}, {"type": "null"}], "title": "Title"},
         "other": {"anyOf": [{"type": "string"}, {"type": "null"}], "title": "Other"},
         "extra": {"title": "Extra", "type": "integer"},
@@ -164,14 +164,14 @@ def test_fields_all():
         "title": "SomeSchema",
         "type": "object",
         "properties": {
-            "id": {"anyOf": [{"type": "integer"}, {"type": "null"}], "title": "ID"},
+            "id": {"title": "ID", "type": "integer"},
             "field1": {"title": "Field1", "type": "string"},
             "field2": {
                 "anyOf": [{"type": "string"}, {"type": "null"}],
                 "title": "Field2",
             },
         },
-        "required": ["field1"],
+        "required": ["id", "field1"],
     }
 
 
--- a/tests/test_orm_schemas.py
+++ b/tests/test_orm_schemas.py
@@ -32,12 +32,12 @@ def test_inheritance():
         "title": "ChildModel",
         "type": "object",
         "properties": {
-            "id": {"anyOf": [{"type": "integer"}, {"type": "null"}], "title": "ID"},
+            "id": {"title": "ID", "type": "integer"},
             "parent_field": {"type": "string", "title": "Parent Field"},
             "parentmodel_ptr_id": {"type": "integer", "title": "Parentmodel Ptr"},
             "child_field": {"type": "string", "title": "Child Field"},
         },
-        "required": ["parent_field", "parentmodel_ptr_id", "child_field"],
+        "required": ["id", "parent_field", "parentmodel_ptr_id", "child_field"],
     }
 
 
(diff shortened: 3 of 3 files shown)


vitalik:
Well there is a bit of controversy here...

when you create schema from model (with ModelSchema) and then want to use it as input payload you do not want to primary key field to be required... I was thinking about it - I do not yet see some perfect solution

pmdevita:
I see what you mean, that could be troublesome for users that are using the same schema for a request and a response data type. But for those cases, they could add the primary key to `fields_optional` and get the same functionality. This PR will enable users to decide what they want on their schemas, but the current behavior does not allow for any choice.

In general, I think it's a bit odd to always translate primary keys as null when every other model field's setting is respected by ModelSchema. ModelSchema is otherwise consistent in translating your database models, so I think this could be surprising behavior to users.

vitalik:
I would also try to play with `Optional` or `readonly` (which makes sense logically - automatic primary keys can be read, but cannot be set in most cases) I think pydantic now handles that

pmdevita:
Sorry to ask, but could you provide a bit more context for what you mean here? How are you suggesting we use `Optional`? Also, as far as I can tell, there isn't a read only property for either Django or Pydantic models.

pmdevita:
Sorry for the late reply. To try this out, I set the read only property manually on a schema I have in a hobby project. This is using Ninja 1.1.0.

image

Which looks correct on the API schema.

image

And looks correct on Swagger UI (before/after)

image

image

However, the immediate problem that results from this is that Pydantic doesn't actually understand that the field is read only, and validation fails during the request. Read only isn't supported by Pydantic and for reference, FastAPI gets around this by [recommending separate request and response schemas](https://fastapi.tiangolo.com/tutorial/extra-models/), and the Ninja docs also [do the same](https://django-ninja.dev/guides/response/). The read only property also has poor support in client library generators, neither [openapi-generator](https://github.com/OpenAPITools/openapi-generator/issues/4190) or the TypeScript client I use, [openapi-client-axios](https://www.npmjs.com/package/openapi-client-axios) support it. In theory, I agree this would be a big part of an ideal solution, but support is just not there, both from Pydantic and the client side. 

Just to give a bit of context on my usecase as well, my workplace is using Ninja now (which despite a few rough edges, is a really lovely experience so thank you!) and we're following the recommended strategy of using different schemas for request and response types. The current behavior of marking all primary keys as nullable would mean that our client teams would have to do an unnecessary null check every time they wanted to use the corresponding field on a response. The schema would not really reflect the API, because the API will never respond null for a primary key. 

Since read only isn't a viable solution, I agree that users should be able to make primary keys nullable so that it would be possible to reuse their schemas for request and response. But ModelSchema's default behavior has always been to follow your model first, and to allow customization and exceptions to be configured from there. Making primary keys nullable by default breaks this otherwise consistent pattern. Given all of this, I think the best option here is still to follow the model field by default, and allow users to set their primary keys as nullable with `fields_optional`, it's consistent but still gives flexibility to those who need it.

Ksauder:
In regards to the readonly discussion, there may be a chance to fix it with some custom typing in pydantic. I don't think it fixes everything, as I'm not sure how you would know in the schema whether it's an incoming payload or outgoing resource. However, it could at least be Optional and create the proper json schema.

I have also taken a crack at fixing the optional primary_key issue. It's rolled into broader changes, but the short and sweet of it is a new parameter `primary_key_optional` to both ModelSchema.Meta and create_schema() in #1281 . This allows the current functionality to exist, or you can make it `False` to revert to pre 1.2.0. Doesn't necessarily deal with this whole issue, and maybe the regression is only a bug in which case I'll remove that extra param.

pmdevita:
It may be possible to emulate proper readonly functionality once we have proper validation vs serialization separation #1139. OpenAPI ecosystem support for it is very weak though, so it might be best to offer it as an option alongside a simpler option for null vs nonnullable primary key.

I think your solution is definitely preferable to the current behavior. The ability to create a custom ModelSchema parent class with custom configuration would smooth over the issue of having to set all of your primary keys as nullable manually (as my current solution would make you do). I still think primary keys being nullable by default is surprising and inconsistently opinionated behavior, and I would recommend it to be toggled off by default. But, even if it is on by default, it should at least provide a path forward for everyone.

brendan-morin:
@vitalik Would you be able to weigh in on if you find either this PR or https://github.com/vitalik/django-ninja/pull/1281 to be a decent solution to this?

I think we're in agreement that the current behavior is limiting to some use cases, so some changes would be needed. Perhaps we could clarify your vision or reservations for this particular functionality, which might help us continue to drive this forward. 

If you are not convinced this is an issue that needs addressing, I think we can focus the conversation towards aligning on expected supported behaviors. If reservations are related to technical specifics (appropriate default behavior, style, configuration etc), that would also help to know and we can work to address that.

If you are busy and don't have the bandwidth to dive into this, perhaps @pmdevita, @Ksauder and myself (and any other interested community members) can align on a narrowly scoped recommendation for sign off.

Ksauder:
I believe this MR should be adopted due to it's direct resolution of the issue. I don't think the function to turn a model field into a pydantic field, `get_schema_field`, should have an opinion on what happens if that field happens to be a primary key. I'm not really sure we need it to be an option like I made it either.

brendan-morin:
Taggin @julienc91, who originally authored [this commit](https://github.com/vitalik/django-ninja/commit/a1a1e5076ba35c95c16cec0c4fb103433700000e#diff-996ad41391ec8d35fe4995efee2131fcbeeb9c226cfc325cb5d159c12eec6ee4), which imposed the constraint to force primary keys to be nullable. Julien, I'd be interested in your thoughts on the above issue and proposed solution, given you originally introduced this particular feature.

pmdevita:
Julien responded in the original issue for this PR here https://github.com/vitalik/django-ninja/issues/1227

Ksauder:
Any interest in resolving this? I support either this solution or adding an argument to `create_schema` and friends to make make pk nullable.

PR #1255, opened 2024-08-05
Fix no URL validation for ModelSchema URLFields

### LINKED ISSUE #1157: [BUG] `models.URLField` serialized & validated as `str`
**Describe the bug**
When using a `URLField` in a Django model and then using a `ModelSchema` for serialization/validation on a `POST` request, the url is just validated as a `str`. Most other fields are covered in Ninja, but this one seems to be missing.

 - Python version: [e.g. 3.6] 3.11.9
 - Django version: [e.g. 4.0] 5.0
 - Django-Ninja version: [e.g. 0.16.2] 1.1.0
 - Pydantic version: [e.g. 1.9.0] 2.7.1

lapinvert:
Fixes #1157

Since Django's implementation of of `get_internal_type()` returns 'CharField' for `URLField`, we were not able to assign a Pydantic type to it, that would be better suited for URLs, such as `AnyUrl`. 

This PR fixes this by adding a special case for 'URLField' to check it's object type with and assign an `internal_type` of 'URLField'.

Related tests were also updated.


lapinvert:
I added a commit because the previous one would break JSON serialization, since `AnyUrl` doesn't serialize as a string. It would throw `Object of type Url is not JSON serializable`.

The problem was discussed [over here](https://github.com/pydantic/pydantic/issues/7186) in Pydantic github and I implemented one of the solutions suggested there by [pavradev](https://github.com/pavradev) and updated the tests.

lapinvert:
I had to improve on the previous solution which would throw `Input should be a valid string` as it would appear that validation on individual objects is done (= when `response=PostSchema`) whereas it's not done when returning lists of objects (`response=List[PostSchema]`).

Issue #1218, opened 2024-07-04
Distinguish between AuthenticationError and AuthorizationError


lordi:
**Is your feature request related to a problem? Please describe.**

Right now, when a user is authenticated but lacks permissions, I raise `AuthenticationError`. This is not totally correct, because it returns 401, but for authorization failures I expect 403 (Forbidden).

**Describe the solution you'd like**

I suggest to add another exception to `ninja.errors` that specifically denotes an Authorization error and results in a 403 response.

c4ffein:
@vitalik @lordi I made #1257 which should answer that need, please tell me if anything is missing

PR #1257, opened 2024-08-06
Distinguish between AuthenticationError and AuthorizationError

c4ffein:
Implements #1218


lordi:
As far as I'm concerned, this looks good, thx

BlaiseMuhirwa:
Can we merge this soon? I also think it looks good and I need this distinction as well since I currently use `django-ninja`.

vitalik:
I think this might be done better...


`AuthenticationError`, `AuthorizationError` should  be both inherited from HttpError (with corresponding codes and messages)

that way only one default http-error handler is needed (and users should be able to overwrite custom handlers for each class individualy)

c4ffein:
> I think this might be done better...
> 
> `AuthenticationError`, `AuthorizationError` should be both inherited from HttpError (with corresponding codes and messages)
> 
> that way only one default http-error handler is needed (and users should be able to overwrite custom handlers for each class individualy)

Didn't want to cause too much changes with this PR, but definitely the way to go

NikolozCh:
Hi, thanks for the updates. When are you guys expecting these changes? Release 1.3.0 doesn't contain AFAIK? Please, correct me if I am wrong. Thanks!

c4ffein:
@NikolozCh you may check [Django Shinobi](https://github.com/pmdevita/django-shinobi), the community fork that is currently maintained
The plan is to merge it back if possible, and if we have no news from @vitalik we will still make the framework more community oriented so the current situation doesn’t happen again

NikolozCh:
Thanks @c4ffein. Django Shinobi worked perfectly well.

PR #1259, opened 2024-08-08
Add default headers for NinjaClientBase instance

c4ffein:
Those changes let you set default headers for instances of NinjaClientBases, which are passed through all requests, unless overwritten.
Related to #664 

- [x] documentation update


vitalik:
Thank you

PR #1261, opened 2024-08-09
Pass serialisation context to model_dump (fixes #1233)

### LINKED ISSUE #1233: Pass serialization context (similar to validation context) with request to `model_dump`
**Is your feature request related to a problem? Please describe.**
We're trying to post-process response entities to erase data from responses based on user permissions. 

We're using custom framework to do this (somewhat similar to ninja permissions, but applied at field level). Based on model class and user permissions it generates list of fields to exclude to use with `model_dump`.
```
excludes = get_excludes_by_permissions(api_response_schema, request.user)
json_response = api_response_schema.model_dump_json(exclude=excludes)
```

Pydantic starting from 2.7 added serialization context support (https://github.com/pydantic/pydantic/pull/8965)

Ninja already passes similar context to validation.
`Operation._result_to_response` https://github.com/vitalik/django-ninja/blob/master/ninja/operation.py#L258-L269

It would be useful to have request in serialization context as well.

scorpp:
Pydantic starting with 2.7 added [support for context param in `model_dump` and `model_dump_json`](https://docs.pydantic.dev/latest/changelog/#new-features_2) allowing to pass arbitrary additional data into custom serializers.

Passing request in serialisation context may be useful in cases where customising serialisation is necessary.

### DIFF
--- a/ninja/operation.py
+++ b/ninja/operation.py
@@ -21,7 +21,7 @@ from django.http.response import HttpResponseBase
 from ninja.constants import NOT_SET, NOT_SET_TYPE
 from ninja.errors import AuthenticationError, ConfigError, Throttled, ValidationError
 from ninja.params.models import TModels
-from ninja.schema import Schema
+from ninja.schema import Schema, pydantic_version
 from ninja.signature import ViewSignature, is_async
 from ninja.throttling import BaseThrottle
 from ninja.types import DictStrAny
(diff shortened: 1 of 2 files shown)


vitalik:
Hi @scorpp 

Could you as well add some test case(s) for for this ?

vitalik:
Thank you

PR #1275, opened 2024-08-22
Allow multiple returned content types from one renderer

bradykieffer:
I love the project - a request we commonly see from users is the ability to obtain our data in separate formats. I've seen my fair share of endpoints that are essentially identical, where one spits out JSON and another spits out CSV. Maintaining these forked views is pretty annoying, so I figured enabling APIs to render responses based on the `Accept` header would be a cool feature 😄 

I picked `Dynamic` in about 30 seconds, so very open to renaming this. But would love to get this type of functionality in.


bradykieffer:
The missing pagination coverage is in: https://github.com/vitalik/django-ninja/pull/1276 which I don't want to blindly copy-paste. Would be great to know if this is a feasible change, or what else would be needed.

LaurierMantel:
Can we get a look at this? @vitalik

PR #1276, opened 2024-08-23
Fixed coverage on Python>=3.11

c4ffein:
`make test-cov` is not at 100% on Python>=3.11 when executed locally.
Nothing is wrong with the pipeline as it is still using Python 3.9 but it is a minor incovenience during development.

It is due to the fact that the handling of `typing.Any` becomes unnecessary with this change in cpython: python/cpython#88690

This PR resolves this situation by adding a quick UT that is not so useful for Python<3.11 but covers that specific `try` - `except` block.

We may (or may not) also add support for Python>=3.11 in the coverage pipeline, in this PR or another one.

### DIFF
--- a/ninja/pagination.py
+++ b/ninja/pagination.py
@@ -282,8 +282,7 @@ def make_response_paginated(paginator: PaginationBase, op: Operation) -> None:
     try:
         new_name = f"Paged{item_schema.__name__}"
     except AttributeError:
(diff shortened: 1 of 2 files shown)


c4ffein:
I guess the change to `typing.Any` is introduced between 3.10 and 3.11, but better just ask for a first review anyway

PR #1281, opened 2024-08-27
RFC: Modelschema rework

Ksauder:
I was having trouble getting inheritance to work in a logical way with the ModelSchema, so I took a stab at updating the ModelSchemaMetaclass, MetaConf, and some of the SchemaFactory. The added inheritance test shows a concrete example. 

I think this rework allows more flexibility and ability to customize moving forward, would love some feedback.

- [ ] rework ModelSchemaMetaclass so it does not use create_schema() under the hood
- [ ] rework the defaults from MetaConf through SchemaFactory methods to be more explicit
- [ ] add `primary_key_optional` option to MetaConf - this toggles how Django fields with `primary_key` are handled


Ksauder:
@vitalik Is this MR anywhere in the ballpark of changes you'd like to see?

vitalik:
Hi @Ksauder 

Sorry have been busy on some other "projects", and if you mark PR as draft it goes a bit under radars
great work on such compact change set

If all is good we can start testing - basically our goal for the next release 

 - better inheritances
 - final get rid of "Config" class  from ninja 0.x
 - primary key option to to give users ability to use one schema for both create and put methods (not yet 100% sure this PR fully helps with that but can proceed)

Ksauder:
Perfect timing, I have been updating and reviewing django ninja on my project today. I'll review and clean this branch up

Ksauder:
@vitalik I have some docs and test coverage yet to update, but I think this is all for this MR.

I have a branch where I'm working through some refactoring to get rid of the Config class, detangle some logic, and get rid of the need for a separate ModelSchema if that is of interest.

PR #1291, opened 2024-09-02
Adding OrderingSchema for ordering QuerySets

acuriel:
## Problem

The `FilterSchema` definition is a simple but effective approach to centralize logic around filtering QuerySets. A similar approach could be followed for ordering.
Current for ordering and filtering in the same handler we would the following:

```python
@api.get("/books")
def list_books(request, filters: BookFilterSchema = Query(...), order_by: list[str] | None = None):
    books = Book.objects.order_by(**order_by)
    books = filters.filter(books)
    return books
```

But what if we want to limit the API to allow ordering by just some of the fields. Or if we want to customize ordering based on custom fields and logic. What if we want to have a similar approach for other data sources, for example ElasticSearch.

## Proposal

This PR propose to include a helper schema class, similar to FilteringSchema, but for ordering. It';s a simple schema class, with only one field: `order_by`, that accepts a list of string.
The allowed fields can be specified through the Config inner class, and a Pydantic validator will check that the provided query values are part of the allowed fields.
The schema then will provide a `.sort()` method (similar to `.filter()`) that we can use to pass the query set, and expect it ordered as a returned value.
The values can be provided using django standard behavior for descending order.

### Example

Using it with out-of-the-box definition, allowing all fields from the model.

```python hl_lines="4"
from ninja import OrderingSchema

@api.get("/books")
def list_books(request, filters: BookFilterSchema = Query(...), ordering: OrderingSchema = Query(...)):
    books = Book.objects.all()
    books = filters.filter(books)
    books = ordering.sort(books)
    return books
```

Using it with custom definition of allowed fields

```python hl_lines="4"
from ninja import OrderingSchema

class BookOrderingSchema(OrderingSchema):
    class Config(OrderingSchema.Config):
        allowed_fields = ['name', 'created_at']  # Leaving out `author` field

@api.get("/books")
def list_books(request, filters: BookFilterSchema = Query(...), ordering: BookOrderingSchema = Query(...)):
    books = Book.objects.all()
    books = filters.filter(books)
    books = ordering.sort(books)
    return books
```

## Other ideas not followed

I also considered to have a default value field in the config, but decided to go with field default definition on custom schema level
Another consideration was to create a class method factory in the OrderingSchema, so it can be define inline, but I wasn't sure if it would be used:

```python
@api.get("/books")
def list_books(
    request,
    filters: BookFilterSchema = Query(...),
    ordering: OrderingSchema.with_allowed_fields('name', 'created_at') = Query(...)
):
    ...
```
```

## Notes

I didn't add field validation with Model definition, to keep the practices followed in the FilterSchema definition.
Also, I think is a good idea to keep this helpers class as simple as possible, and give room to personalization for more complex scenarios. However, let me know if validating `allowed_fields` with Model fields is something we would like to have, and I can update the PR.

This was really useful in a personal project, were we needed to provide different ordering behaviors for a QuerySet and for an OpenSearch query. We had to do similar personalizations for pagination and filtering.

### DIFF
--- a/docs/docs/guides/input/ordering.md
+++ b/docs/docs/guides/input/ordering.md
@@ -1,0 +1,51 @@
+# Ordering
+
+If you want to allow the user to order your querysets by a number of different attributes, you can use the provided class `OrderingSchema`. `OrderingSchema`, as a regular `Schema`, it uses all the
+necessary features from Pydantic, and adds some some bells and whistles that will help use transform it into the usual Django queryset ordering.
+
+You can start using it, importing the `OrderingSchema` and using it in your API handler in conjunction with `Query`:
+
+```python hl_lines="4"
+from ninja import OrderingSchema
+
+@api.get("/books")
+def list_books(request, ordering: OrderingSchema = Query(...)):
+    books = Book.objects.all()
+    books = ordering.sort(books)
+    return books
+```
+
+Just like described in [defining query params using schema](./query-params.md#using-schema), Django Ninja converts the fields defined in `OrderingSchema` into query parameters. In this case, the field is only one: `order_by`. This field will accept multiple string values.
+
+You can use a shorthand one-liner `.sort()` to apply the ordering to your queryset:
+
+```python hl_lines="4"
+@api.get("/books")
+def list_books(request, ordering: OrderingSchema = Query(...)):
+    books = Book.objects.all()
+    books = ordering.sort(books)
+    return books
+```
+
+Under the hood, `OrderingSchema` expose a query parameter `order_by` that can be used to order the queryset. The `order_by` parameter expects a list of string, representing the list of field names that will be passed to the `queryset.order_by(*args)` call. This values can be optionally prefixed by a minus sign (`-`) to indicate descending order, following the same standard from Django ORM.
+
+## Restricting Fields
+
+By default, `OrderingSchema` will allow to pass any field name to order the queryset. If you want to restrict the fields that can be used to order the queryset, you can use the `allowed_fields` field in the `OrderingSchema.Config` class definition:
+
+```python hl_lines="3"
+class BookOrderingSchema(OrderingSchema):
+    class Config(OrderingSchema.Config):
+        allowed_fields = ['name', 'created_at']  # Leaving out `author` field
+```
+
+This class definition will restrict the fields that can be used to order the queryset to only `name` and `created_at` fields. If the user tries to pass any other field, a `ValidationError` will be raised.
+
+## Default Ordering
+
+If you want to provide a default ordering to your queryset, you can assign a default value in the `order_by` field in the `OrderingSchema` class definition:
+
+```python hl_lines="2"
+class BookOrderingSchema(OrderingSchema):
+    order_by: List[str] = ['name']
+```
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -59,6 +59,7 @@ nav:
           - guides/input/file-params.md
           - guides/input/request-parsers.md
           - guides/input/filtering.md
+          - guides/input/ordering.md
       - Handling responses:
           - Defining a Schema: guides/response/index.md
           - guides/response/temporal_response.md
--- a/ninja/__init__.py
+++ b/ninja/__init__.py
@@ -9,6 +9,7 @@ from ninja.files import UploadedFile
 from ninja.filter_schema import FilterSchema
 from ninja.main import NinjaAPI
 from ninja.openapi.docs import Redoc, Swagger
+from ninja.ordering_schema import OrderingSchema
 from ninja.orm import ModelSchema
 from ninja.params import (
     Body,
@@ -54,6 +55,7 @@ __all__ = [
     "Schema",
     "ModelSchema",
     "FilterSchema",
+    "OrderingSchema",
     "Swagger",
     "Redoc",
     "PatchDict",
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

**Comment by shijl0925** [comment]
How about `OrderingSchema.sort` support list type sort?
```
class OrderingSchema(OrderingBaseSchema):
    def sort(self, items: Union[QuerySet, List]) -> Union[QuerySet, List]:
        if self.order_by:
            if isinstance(items, QuerySet):  # type:ignore
                return items.order_by(*self.order_by)
            elif isinstance(items, list) and items:
                def multisort(xs: List, specs: List[Tuple[str, bool]]) -> List:
                    getter = itemgetter if isinstance(xs[0], dict) else attrgetter
                    for key, reverse in specs:
                        xs.sort(key=getter(key), reverse=reverse)
                    return xs

                return multisort(
                    items,
                    [
                        (o[int(o.startswith("-")):], o.startswith("-"))
                        for o in self.order_by
                    ],
                )

        return items
```

**Comment by aryan-curiel** [comment]
> How about `OrderingSchema.sort` support list type sort?

Adding support for list type is definitely a good idea. However, not sure adding it to this sort method is the most convenient approach. 
For example, if we want to support other collections, then we would need to add another type hint, and another `elif`.
Any extension for supporting other collections or types, might be more suitable as a subclass of the base OrderingBaseSchema. However, we might need to make some small arrangements in the base class for a better definition.

**Comment by shijl0925** [comment]
May I ask if there is any progress on this OrderingSchema? I found same failure in these pull request actions. But I can not see these failed logs. What happened about this pull request?

**Comment by aryan-curiel** [comment]
> May I ask if there is any progress on this OrderingSchema?

Honestly, I'm not sure. I haven't received any review, and find no sense in the failed tests related to schemas in the pipeline, so I haven't been able to fix it. If you need it with some level of urgency, feel free to bring the  added code to your code base.

**Comment by aryan-curiel** [comment]
@crbunney Wow, thanks! Didn't know about it. I'm also doing a library for supporting filtering, ordering and pagination for Ninja and FastAPI with multiple data sources. I'm not a big fan of using decorators for this :( It feels like it;s obfuscating the logic. In any case, it's amazing that we have already something like ninja extras!

**Comment by lthon-sha** [comment]
Hi @aryan-curiel, just wanted to say a huge thank you for this PR! The OrderingSchema implementation is incredibly clean, elegant. It's exactly the solution I was looking for. Amazing work!

**Comment by lthon-sha** [comment]
I noticed that when order_by is empty ([]), calling `queryset.order_by()` clears the model's default ordering from Meta.ordering.

 Maybe we should only call `order_by()` when there are actual fields to sort by?

**Comment by aryan-curiel** [comment]
> I noticed that when order_by is empty ([]), calling `queryset.order_by()` clears the model's default ordering from Meta.ordering.
> 
> Maybe we should only call `order_by()` when there are actual fields to sort by?

@lthon-sha Thanks for that amazing feedback! I will update the PR according to your recommendation and add the needed tests

**Comment by aryan-curiel** [comment]
@vitalik I would love a feedback from any of the more experienced contributors. It has been almost a year since the PR was opened, and I would love to move forward with it if it's something that you all consider it might bring some value.

**Comment by vitalik** [comment]
Hi @aryan-curiel 

Sorry for dragging this for this long - but we finally getting rid of "Config" child class in favour of Meta (as it will be "broken" in future pedantic releases 

Could you rework Config -> Meta  like https://github.com/vitalik/django-ninja/blob/52d510ac9d50794d1935086127132d917da670c9/ninja/filter_schema.py#L44-L47

**Comment by aryan-curiel** [comment]
Hi @vitalik 

That's awesome! 
PR updated with the requested changes. Thanks for the feedback.

Let me know if anything else is needed.

**Comment by cassandra-aq** [comment]
Hi, @vitalik do you have any plan to merge this PR ? It seems like @aryan-curiel has made the change you asked )

**Comment by brendan-morin** [comment]
This looks great, @aryan-curiel, thank you for putting this PR together! 

For the ordering param name, I think because this would be public to end users, we would want to have the query parameter itself be configurable rather than hardcoding to `order_by`. Use cases could be preferences/requirements to use other field naming/consistency conventions (sort/rank/etc), as well as users who have APIs localized into languages other than English. Having a default `order_by` seems reasonable, but allowing an override would improve flexibility (perhaps something like the would be possible):

```
class Meta:
    ordering_query_param = "order_by"
        
...

@model_validator(mode="after")
    def validate_ordering(self) -> "OrderingBaseSchema":
        field_name = getattr(self.Meta, "ordering_query_param")
        value = getattr(self, field_name, [])
        ...
        
```

Similarly, it may also not always be desirable to expose direct model relationships in ordering to end users, so facilitating some type of mapping of declared filters to publicly facing options would be preferable. Unless I'm misunderstanding, this seems to basically just pass through to the ORM with some optional validation, but if we have a field with a lot of relationships or strange internal naming conventions, we would ideally be able to alias this for end users. 

For example, we could have a valid relationship `product__variant__brand__formatted_name`, but may want to simplify in our api as the query param `brand`. Perhaps allowing `allows_fields` to represent a mapping would make sense (or `allowed_fields_map` if mixed list/dict types feel wrong):

```
class Meta:
    allowed_fields = {
        "brand": "product__variant__brand__formatted_name"
    }

aryan-curiel:
@brendan-morin 
I updated the PR based on your requests. Please take a look, and let me know your thoughts on the applied changes
Tests and docs also updated accordingly.

PR #1293, opened 2024-09-08
Fix `SynchronousOnlyOperation` error when `paginate` is used with async views returning Django querysets.

qodot:
When applying the `paginate` decorator to an async view that returns a Django queryset, a `SynchronousOnlyOperation` error occurs.

To resolve this, I've added `_inject_async_pagination` within the `paginate` and `RouterPaginated` and differentiated the behavior based on the view type.

### DIFF
--- a/ninja/pagination.py
+++ b/ninja/pagination.py
@@ -1,3 +1,4 @@
+from collections.abc import AsyncIterable, Iterable
 import inspect
 from abc import ABC, abstractmethod
 from functools import partial, wraps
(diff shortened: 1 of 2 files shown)


ninja/pagination.py
```diff
@@ -170,69 +171,45 @@ def my_view(request):
...
-                kwargs[paginator.pass_parameter] = pagination_params
-
-            items = await func(request, **kwargs)
-
-            result = await paginator.apaginate_queryset(
-                items, pagination=pagination_params, request=request, **kwargs
-            )
-
-            async def evaluate(results: Union[List, QuerySet]) -> AsyncGenerator:
-                for result in results:
```
michalasia10, line 207: This can be fixed by adding `async` to this for loop.

Anyway, bump beacuse I had also this issue ;d

PR #1317, opened 2024-10-14
Small typo fixes and doc mention router import

michaelg100:
Added a small comment in the testing doc to call out to the user to import the router. Then just some other typo fixes I saw.


vitalik:
Thank you

Issue #1109, opened 2024-03-14
[BUG] 


scott-8:
**Describe the bug**
Django-ninja uses `get_internal_type` to determine field type for schemas. This works with all of the default fields, but does not work with custom fields. For fields that don't define it, Django will return `self.__class__.__name__`. For fields that define a custom `get_internal_type` or `db_type`, the type parsed by django-ninja might not necessarily be correct.

For example, I have a custom field that stores data in the database as a string, but converts that string to a bool when the data is read from the database. Since `get_internal_type` returns `CharField`, django-ninja generates a schema with `string`, while ideally it would be a bool.

This is fixable by adding annotations for schemas using these custom fields, but it would be nice to have a solution on the field level. The code in `get_schema_field` gets the field type like this:
```python
internal_type = field.get_internal_type()
python_type = TYPES[internal_type]
```

I'd propose changing it to this:
```python
if hasattr(field, "get_python_type"):
    python_type = field.get_python_type()
else:
    internal_type = field.get_internal_type()
    python_type = TYPES[internal_type]
```

That would let a user define `get_python_type` on the custom field. If this is something you'd be interested in I can make the change.

**Versions (please complete the following information):**
 - Python version: 3.12
 - Django version: 5.0.2
 - Django-Ninja version: 1.1.0
 - Pydantic version: 2.6.3

PR #1320, opened 2024-10-18
Add support for custom django fields (#1109)

RobertKolner:
## Motivation

Using exclusively `django.db.fields.Field.get_internal_type` to determine which Pydantic field to use in the built schemas makes it difficult to create custom django fields. This PR adds support for a new method `get_schema_type`, which allows using `get_internal_type` for its intended purpose (determining database column types, etc.), while at the same time allowing specifying which Pydantic schema field to use.

## Concerns

- There is of course always a possibility that the users already define `get_schema_type` function in their field classes. Right now there is no checks or fallbacks for this scenario. 
- The name `get_schema_type` is somewhat generic. I have no better ideas though.
- ~~Monkey-patching `ninja.orm.fields.TYPES` is not a great solution for future-proofing the solutions using the provided pattern to define custom fields. The should probably be an additional function in `ninja.orm.fields` to do this.~~ This is no longer relevant.

## Notes

I don't currently have time to test the code, but see no reason why it shouldn't work. If you want to wait, I can write proper tests in a couple of weeks.

### DIFF
--- a/ninja/orm/fields.py
+++ b/ninja/orm/fields.py
@@ -151,11 +151,7 @@ def get_schema_field(
             python_type = field.get_schema_type()
         else:
             internal_type = field.get_internal_type()
-
-            try:
-                python_type = TYPES[internal_type]
-            except KeyError as e:
-                raise KeyError("Type '{0}' isn't registered in ninja.orm.fields.TYPES".format(internal_type)) from e
+            python_type = TYPES[internal_type]
 
         if field.primary_key or blank or null or optional:
             default = None


docs/docs/guides/response/django-pydantic.md
```diff
@@ -147,9 +147,48 @@ def modify_data(request, pk: int, payload: PatchDict[GroupSchema]):
...
+To achieve that, we can add a `get_schema_type` function to the django field and add it to supported
+types in Django Ninja:
+
+```python
+class TranslatedTextField(models.JSONField):
+    ...
+    def get_schema_type(self):
+        return "TranslatedTextField"
+
+ninja.orm.fields.TYPES["TranslatedTextField"] = TranslatedTextFieldSchema
```
jerivas, line 193:
```suggestion
class TranslatedTextField(models.JSONField):
    ...
    def get_schema_type(self):
        return TranslatedTextFieldSchema
```

Do we really need to add the method AND modify the `TYPES` list? I think the method should return the schema class itself and ninja should use it as the `python_type` right away.

pmdevita:
Looks like a different solution was decided on #1352

PR #1326, opened 2024-10-25
document how to use an optional file input

toudi:
It took me a couple of google searches before I understood how to make the optional file input so I thought I'd help other people by extending the docs :)


vitalik:
thank you

what was you intuitive feeling on how to make it optional ?

toudi:
I tried couple of things. First thing I tried was to do `file: UploadedFile = None`, then  `UploadedFile | None = File(...)`, (but this one I think exploded on generating schema) then I tried `UploadedFile = File(required=False)` but then realized that required was not one of the parameters. I think it was due to my lack of experience. I then found the solution by pure accident by browsing trough issues of django-ninja and so I thought that I'd save some other developers some head-scratching time :)

PR #1338, opened 2024-11-17
feat: support returning pydantic model as result

mh-alahdadian:
allow returning pydantic model as result and pass converting and validation

### DIFF
--- a/ninja/operation.py
+++ b/ninja/operation.py
@@ -255,28 +255,31 @@ class Operation:
             # Empty response.
             return temporal_response
 
-        if isinstance(result, response_model):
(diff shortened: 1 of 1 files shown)


vitalik:
Hi @mh-alahdadian 

could you also add test case for multiple response codes ?

```Python

@router.get("/check_pydantic", response={200: UserModel, 201: Message})
def check_pydantic(request):
    if something:
          return 201, {"message": "Created"}
    return 200, UserModel(id=1, user_name="John")
```

vitalik:
well coverage shows that it never actually calls the needed line

PR #1339, opened 2024-11-18
Add auser mock support for TestClient (#1252)

### LINKED ISSUE #1252: [BUG] Mock Request Generated by Test Client Doesn’t Support auser()
**Describe the bug**

In Django 5.0, `request.auser()` is added to access the authenticated user in async style. This can be helpful for use case like async authentication callback. But Django Ninja's test client doesn't support it.

Example Code:

```python
from django.contrib.auth.models import AnonymousUser, User
from ninja import NinjaAPI
from ninja.security import SessionAuth
from ninja.testing import TestClient


class AsyncSessionAuth(SessionAuth):
    async def __call__(self, request):
        return await self.authenticate(request, None)

    async def authenticate(self, request: HttpRequest, key: str | None) -> User | AnonymousUser | None:
        user = await request.auser()
        return user if user.is_authenticated else None


def test_foobar(api: NinjaAPI):
    @api.get("/foobar", auth=AsyncSessionAuth())
    def foobar(request) -> str:
        return "foobar"

    client = TestClient(api)
    response = client.get("/foobar")
    assert response.status_code == 200
    assert response.json() == "foobar"
```

Accessing the view in an actual API works as expected. But when running the test the following error is raised: `TypeError: object Mock can't be used in 'await' expression` during the line `user = await request.auser()`.

 - Python version: 3.12
 - Django version: 5.0.7
 - Django-Ninja version: 1.2.2
 - Pydantic version: 2.8.2

LuYanFCP:
As a new open-source contributor, I recently encountered the same issue described in #1252 . I've created a simple patch to address this problem. fixes #1252

### DIFF
--- a/ninja/testing/client.py
+++ b/ninja/testing/client.py
@@ -1,7 +1,7 @@
 from json import dumps as json_dumps
 from json import loads as json_loads
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
-from unittest.mock import Mock
+from unittest.mock import Mock, AsyncMock
 from urllib.parse import urljoin
 
 from django.http import QueryDict, StreamingHttpResponse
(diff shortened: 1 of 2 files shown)


vitalik:
Hi @LuYanFCP 

The AsyncMock seems introduced in python 3.8+

I guess you can put custom async mock implementation (or just copy from 3.8) into https://github.com/vitalik/django-ninja/tree/master/ninja/compatibility folder and use it

PR #1340, opened 2024-11-19
chore: Refactor async pagination

Amr-elwetaidy:
## Async Paginator

The evaluate AsyncGenerator causes:

`django.core.exceptions.SynchronousOnlyOperation: You cannot call this from an async context - use a thread or sync_to_async`

This part of the code is the reason:
Screenshot_1

### **In fact,** it's unnecessary at all as django already handles the queryset and evaluates it as an AsyncGenerator once you call it with `async for`

---

Similarly with the `count()` and `acount()` methods, you can directly call them on the queryset and no need to call `all()` first.

Screenshot_2

### DIFF
--- a/ninja/pagination.py
+++ b/ninja/pagination.py
@@ -2,7 +2,7 @@ import inspect
 from abc import ABC, abstractmethod
 from functools import partial, wraps
 from math import inf
-from typing import Any, AsyncGenerator, Callable, List, Optional, Tuple, Type, Union
+from typing import Any, Callable, List, Optional, Tuple, Type, Union
 
 from django.db.models import QuerySet
 from django.http import HttpRequest
@@ -53,7 +53,7 @@ class PaginationBase(ABC):
         """
         try:
             # forcing to find queryset.count instead of list.count:
-            return queryset.all().count()
+            return queryset.count()
         except AttributeError:
             return len(queryset)
 
@@ -70,7 +70,7 @@ class AsyncPaginationBase(PaginationBase):
 
     async def _aitems_count(self, queryset: QuerySet) -> int:
         try:
-            return await queryset.all().acount()
+            return await queryset.acount()
         except AttributeError:
             return len(queryset)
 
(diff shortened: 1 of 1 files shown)


vitalik:
I think .all there used to foce method that does not exist in a `list`

SCR-20241119-jvcn

Amr-elwetaidy:
> I think .all there used to foce method that does not exist in a `list`
> 
> SCR-20241119-jvcn

I missed that one, but still why not use `if not isinstance(queryset, list):  ...`

meequz:
Why it is closed? The issue is still in the codebase, and is pretty annoying

Amr-elwetaidy:
@meequz i got busy but i am planning to make a better pr for this issue soon.

meequz:
> @meequz could you provide code example where you have issue ?

Sure, here is my endpoint:

```
# api.py
from ninja import ModelSchema, Router
from ninja.pagination import paginate

from .models import Book


router = Router()


class BookOut(ModelSchema):
    class Meta:
        model = Book
        fields = 'title', 'year'


@router.get(
    '/books',
    response={200: list[BookOut]},
)
@paginate
async def get_user_books(request):
    user = await request.auser()
    books = Book.objects.filter(user=user).order_by('-created_at')
    return books
```

With this code, the error I get is `django.core.exceptions.SynchronousOnlyOperation`.

What helped is to edit the `evaluate` function inside a _inject_pagination function in `ninja/pagination.py` (line 207) to this (adds `async` before the `for result in results`):
```
            async def evaluate(results: Union[List, QuerySet]) -> AsyncGenerator:
                async for result in results:
                    yield result
```

but since I'd prefer not to maintain custom changes in the ninja lib, I resolved the issue by applying custom pagination class which looks like:

```
from ninja.conf import settings
from ninja.pagination import LimitOffsetPagination


class MyPagination(LimitOffsetPagination):

    async def apaginate_queryset(self, queryset, pagination, **params):
        """
        Added async list comprehension
        """
        offset = pagination.offset
        limit: int = min(pagination.limit, settings.PAGINATION_MAX_LIMIT)
        return {
            "items": [item async for item in queryset[offset : offset + limit]],
            "count": await self._aitems_count(queryset),
        }
```

It only changes one line, adds `async for` list comprehension to 'items' in the returned dict.

Honestly, I don't quite understand why there is no `async for` in `evaluate`, because it's applied exactly when async call is found (`is_async_callable(func)` is True). I'm glad I can help to resolve this issue.

PR #1354, opened 2024-12-07
fix: async pagination evaluate fix for queryset object

WellingtonNico:
Hello guys, you are doing a great job! 

I found a problem using default `PageNumberPagination` with async endpoint returning a queryset like this:

```python
    @route.get("/", response=list[TodoListSchema])
    @paginate(PageNumberPagination)
    async def list_todos(self, filters: TodoFilterSchema = Query()):
        todos = filters.filter(
            Todo.objects.filter(user=self.context.request.user).values("id", "title")
        )
        return todos
```

I was getting that error about calling sync method, so I made a small change in your `PageNumberPagination` and `LimitOffsetPagination`.
It was not a big change, I also adjusted your tests and everything is working just fine.

The reason why the error was being raised is the same as if count was called in async loop, so that change was really needed. 

I don't hope you to merge my pull request, but that change is really needed and you should merge it to properly work with`QuerySet` object.

### DIFF
--- a/ninja/conf.py
+++ b/ninja/conf.py
@@ -11,6 +11,7 @@ class Settings(BaseModel):
         "ninja.pagination.LimitOffsetPagination", alias="NINJA_PAGINATION_CLASS"
     )
     PAGINATION_PER_PAGE: int = Field(100, alias="NINJA_PAGINATION_PER_PAGE")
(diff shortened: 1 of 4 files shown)


WellingtonNico:
I forgot to mention that I also changed the way to configure page_size, more like a way I think we are already used to do in django.

Issue #1055, opened 2024-01-18
Nested routers w/ path parameters?


akotsampaseris:
# Use Case

I have several objects that belong to a certain customer, who has a unique **customer_id**. I want to include **"/customers/{customer_id}/"** as a prefix for all subsequent routers that handle the objects that belong to that customer. And I need the **customer_id** to ensure that the user/agent requesting access to these objects belongs to the same customer.

# Example

**In expense/api.py:**

```python
router = Router(tags=["Expense"])
router.add_router("customers/{customer_id}/", products_router)
```

**In products/api.py:**
```python
router = Router(tags=["Products"])

@router.get("/products", response={200: List[Products], 400: Error})
def get_products(request: WSGIRequest, customer_id: int):
    products = ProductService.get_products()
    return 200, products
``` 

**Error Response:**
```javascript
{
    "detail": [
        {
            "loc": [
                "query",
                "customer_id"
            ],
            "msg": "field required",
            "type": "value_error.missing"
        }
    ]
}
``` 

# How I want to solve it

I have created a Router object for each Django app, eg app_router, and I am adding each of those routers to the main Router object, eg main_router, using **main_router.add_router("customers/{customer_id}", app_router)**. *_(The router names are used just as an example)._

# Problem

I need to access the **customer_id** as a path parameter in my views. The problem is that the view does not recognize it as a path parameter, because it belongs to a higher level component, even though the endpoint is created correctly. When I include **"/customers/{customer_id}/"** directly to the same exact view, it obviously sees the path parameter and it works correctly. But this is an issue when you have a big number of endpoints that you need to include that.

# Questions

I saw that FastAPI initially had a similar issue and the path parameters from nested routers were not included in the view level. But then they fixed that and included that functionality in a later release. Is there a way to do that with Django-Ninja already? If not, is it possible to include that to a later version? Is there a reason that this structure would not be advised? Thank you!

eznix86:
Not sure, but Maybe try to create a [global auth](https://django-ninja.dev/guides/authentication/#global-authentication)

``` python
router = Router(tags=["Expense"], auth=AuthHere) # <---
router.add_router("customers/{customer_id}/", products_router)
```

Then inside of this
```diff
router = Router(tags=["Products"])

@router.get("/products", response={200: List[Products], 400: Error})
-def get_products(request: WSGIRequest, customer_id: int):
+def get_products(request: WSGIRequest):
+   user_id = request.auth
    products = ProductService.get_products()
[...]

PR #1365, opened 2024-12-18
Updated docs for url parameters in nested routers

kentrh:
As mentioned in this thread: https://github.com/vitalik/django-ninja/issues/1055, and it took me a while to find that thread, so thought it was alright to update the docs to reflect how this could be done with a working example.


vitalik:
Thank you

PR #1371, opened 2024-12-25
Add HttpToken class for authenticating tokens with 'Token' prefix

Mooshiiee:
Just a copy-and-paste of the HttpBearer class that instead checks for 'Token' prefix.

'Token' Prefix comes standard with Django DRF TokenAuthentication  and should be included for more seamless use within DRF ecosystem.

### DIFF
--- a/ninja/security/http.py
+++ b/ninja/security/http.py
@@ -42,6 +42,28 @@ class HttpBearer(HttpAuthBase, ABC):
         pass  # pragma: no cover
 
 
+class HttpToken(HttpAuthBase, ABC):
+    openapi_scheme: str = "token"
+    header: str = "Authorization"
+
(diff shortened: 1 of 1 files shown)


vitalik:
Hi @Mooshiiee 

I think you can simply extend bearer class:

```Python
class HttpToken(HttpBearer):
    openapi_scheme: str = "token"
```

jnoring-pw:
I used @vitalik 's precise solution.

It's also worth noting DRF is incorrect to use `"token"`; that is not covered by any official IETF standard and is essentially a custom scheme name. Also, the openapi documentation barfs if you try to use `"token"`.

Mooshiiee:
I appreciate the response! Extending the class is an embarrassing oversight. Happy New Year!

PR #1373, opened 2024-12-28
Support multi-period intervals in throttling

robhudson:
This pull request enhances the throttling functionality to support multi-period intervals, such as `5/30s`, `10/5m`, and `100/2h`. Previously, the throttling implementation only allowed single-period intervals (e.g., `5/s`, `100/d`), which limited the granularity and flexibility of rate limits.

The new implementation expands the allowed syntax for throttle rates to accommodate use cases requiring multi-unit periods, making it easier to define more precise rate limits.

### Motivation

While single-period intervals work for many scenarios, they lack flexibility for more complex use cases. For example:

* Higher granularity: A rate like `20/5m` (20 requests every 5 minutes) cannot be expressed with the current system.
* Short-term bursts: Rates like `50/2h` allow finer control over bursts without spanning a full day.


vitalik:
Thank you

Issue #882, opened 2023-10-18
[BUG] ModelSchema does not validate Blank fields properly


jam13:
**Describe the bug**
Given a model field:
```Python
description = models.TextField(blank=True)
```

`ModelSchema` treats it as an optional field (which is I guess correct), but doesn't default it to an empty string. This results in integrity errors when trying to save to the db.

Excluding the field from `model_fields` and explicitly declaring it as:
```Python
project_description: str = ""
```
is a workaround.

From what I understand, Django encourages the use of `blank` without `null` for text fields to avoid a mix of empty values.

**Versions (please complete the following information):**
 - Python version: 3.11
 - Django version: 4.2
 - Django-Ninja version: 1.0b2
 - Pydantic version: 2.4

michalwerner:
I also stumbled upon this issue.
The fix seems to be simple but I'm not sure if it won't result in other issues or if and if so why this was intentional.
See draft PR: https://github.com/vitalik/django-ninja/pull/1374

PR #1374, opened 2024-12-29
Don't make field optional if blank=True

michalwerner:
Draft related to https://github.com/vitalik/django-ninja/issues/882


sephii:
Is there anything that prevents this PR from being merged? There’s no reason to set a field to null if it’s not nullable.

PR #1385, opened 2025-01-09
CONTRIBUTING.md - recommend `pipx install flit` instead pip

Fak3:
unlike pip, pipx separates flit dependencies from system-wide site-packages


vitalik:
Hi @Fak3 
We are actually planning to migrate all tooling to uv

Issue #1386, opened 2025-01-09
[BUG] Async authenticate() in HttpBearer causes crash


mmkhitaryan:
**Describe the bug**
I expect that if I write the same `authenticate` handlers in async and sync, they behave the same. But in reality, I get proper 401 with SyncBearerAuth and crash for AsyncBearerAuth with the same handling logic.
```python
Traceback (most recent call last):
  File "C:\Users\Marat\Desktop\csmbackend\.venv\Lib\site-packages\ninja\operation.py", line 190, in _run_authentication
    result = async_to_sync(callback)(request)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Marat\Desktop\csmbackend\.venv\Lib\site-packages\asgiref\sync.py", line 254, in __call__
    return call_result.result()
           ^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Marat\AppData\Local\Programs\Python\Python312\Lib\concurrent\futures\_base.py", line 449, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Marat\AppData\Local\Programs\Python\Python312\Lib\concurrent\futures\_base.py", line 401, in __get_result
    raise self._exception
  File "C:\Users\Marat\Desktop\csmbackend\.venv\Lib\site-packages\asgiref\sync.py", line 331, in main_wrap
    result = await self.awaitable(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: object NoneType can't be used in 'await' expression
``` 

**Versions (please complete the following information):**
 - Python version: 3.12.8
 - Django version: 5.1.4
 - Django-Ninja version: 1.3.0
 - Pydantic version: 2.10.5

Note you can quickly get this by runninng in `./manage.py shell` this line:
```python
import pytest

from ninja.security import (
    HttpBearer,
)
from ninja import NinjaAPI
from ninja.testing import TestClient, TestAsyncClient

class SyncBearerAuth(HttpBearer):
    def authenticate(self, request, token: str):
        return

class AsyncBearerAuth(HttpBearer):
    async def authenticate(self, request, token: str):
        return

api = NinjaAPI()

@api.get("/sync/", auth=SyncBearerAuth())
def sync_auth(request):
    return

@api.get("/async/", auth=AsyncBearerAuth())
def async_auth(request):
    return

[...]

PR #1387, opened 2025-01-09
Fix of TypeError when returning None in AuthBase.__call__ from async code #1386

mmkhitaryan:
# The root cause
The function `_run_authentication` is made to run `AuthBase.__call__` and execute what if returned. But for an example, the [HttpBearer](https://github.com/vitalik/django-ninja/blob/39f40b30ebcc368471a8c1449355990fa63ec0e2/ninja/security/http.py#L22) returns None in cases when a token is misplaced in the request:
```
    def __call__(self, request: HttpRequest) -> Optional[Any]:
        headers = request.headers
        auth_value = headers.get(self.header)
        if not auth_value:
            return None
        parts = auth_value.split(" ")

        if parts[0].lower() != self.openapi_scheme:
            if settings.DEBUG:
                logger.error(f"Unexpected auth - '{auth_value}'")
            return None
        token = " ".join(parts[1:])
        return self.authenticate(request, token)
```
Even though None is returned, the function `_run_authentication` tries to call `async_to_sync` without checking if the coroutine is None.

Because of the call to async_to_sync with None, instead of a coroutine, this TypeError is raised:
```
  File "C:\Users\Marat\Desktop\csmbackend\.venv\Lib\site-packages\ninja\operation.py", line 190, in _run_authentication
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Marat\Desktop\csmbackend\.venv\Lib\site-packages\asgiref\sync.py", line 254, in __call__
    return call_result.result()
           ^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Marat\AppData\Local\Programs\Python\Python312\Lib\concurrent\futures\_base.py", line 449, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "C:\Users\Marat\AppData\Local\Programs\Python\Python312\Lib\concurrent\futures\_base.py", line 401, in __get_result
    raise self._exception
  File "C:\Users\Marat\Desktop\csmbackend\.venv\Lib\site-packages\asgiref\sync.py", line 331, in main_wrap
    result = await self.awaitable(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: object NoneType can't be used in 'await' expression
```

# Fix
Because we know that the coroutine is async:
```
```
We can assume that calling it without putting the coroutine into the eventloop won't trigger the coroutine body. So we can safely call the coroutine without putting it into the eventloop, and check if the result is a coroutine, not a function:
```
if inspect.iscoroutine(initial_result):
    # If it's a coroutine, use async_to_sync to resolve it
    resolved_result = async_to_sync(callback)(request)
else:
    # If it's not a coroutine, use the initial result
    resolved_result = initial_result
```
That way we won't call `async_to_sync` with None, thus triggering TypeError.


vitalik:
Hi @mmkhitaryan 

Thank you for PR


a few issues I see

1) missing 100% coverage 


2) I thinik it can be simplified to 


```python
    def _run_authentication(self, request: HttpRequest) -> Optional[HttpResponse]:
            try:
                    result = callback(request)
                else:
                    result = callback(request)
            except Exception as exc:
                return self.api.on_exception(request, exc)

            if result:
                request.auth = result  # type: ignore
                return None
        return self.api.on_exception(request, AuthenticationError())
```

keep in mind that we check bool authentication result (not only None) `if result:`

vitalik:
Hi @mmkhitaryan 

I do not see any check run for this PR for some reason - maybe try push another commit or something to initiate it ?

vitalik:
Thank you

PR #1388, opened 2025-01-10
Adding the event source method to NinjaAPI and Router (SSE)

rroblf01:
With this PR I propose to make it easier to use SSE in Django Ninja.

### DIFF
--- a/docs/docs/guides/input/operations.md
+++ b/docs/docs/guides/input/operations.md
@@ -9,6 +9,7 @@ An `operation` can be one of the following [HTTP methods](https://developer.mozi
 - PUT
 - DELETE
 - PATCH
+- EVENT SOURCE
 
 **Django Ninja** comes with a decorator for each operation:
 
@@ -31,6 +32,10 @@ def delete_operation(request):
 
 @api.patch("/path")
 def patch_operation(request):
+    ...
+
+@api.event_source("/path")
+def event_source_operation(request):
     ...
 ```
 
--- a/docs/docs/guides/input/operations.md
+++ b/docs/docs/guides/input/operations.md
@@ -38,6 +38,8 @@ def patch_operation(request):
 def event_source_operation(request):
     ...
 ```
+> [!WARNING]
+> Note that using synchronous SSE may block the WSGI worker; use asynchronous SSE with the ASGI worker for better performance.
 
 See the [operations parameters](../../reference/operations-parameters.md)
 reference docs for information on what you can pass to any of these decorators.
--- a/ninja/testing/client.py
+++ b/ninja/testing/client.py
@@ -1,3 +1,5 @@
+import types
+from collections.abc import AsyncIterable, Iterable
 from json import dumps as json_dumps
 from json import loads as json_loads
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
@@ -145,10 +147,12 @@ class NinjaClientBase:
         request.META = request_params.pop("META", {"REMOTE_ADDR": "127.0.0.1"})
         request.FILES = request_params.pop("FILES", {})
 
-        request.META.update({
-            f"HTTP_{k.replace('-', '_')}": v
-            for k, v in request_params.pop("headers", {}).items()
-        })
+        request.META.update(
+            {
+                f"HTTP_{k.replace('-', '_')}": v
+                for k, v in request_params.pop("headers", {}).items()
+            }
+        )
 
         request.headers = HttpHeaders(request.META)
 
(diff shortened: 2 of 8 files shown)


vitalik:
Could you also add tests for async views ? 

the sync mode does not make much sense actually as it can easily  occupy all wsgi threads/processes

rroblf01:
@vitalik I need to investigate, when I test with html views it works, but when I run the tests, it fails saying

"TypeError: object StreamingHttpResponse can't be used in 'await' expression"

I will continue investigating the reason.

rroblf01:
Hello, @vitalik I have added a test to check the asynchronous endpoint.

docs/docs/guides/input/operations.md
```diff
@@ -32,6 +33,10 @@ def delete_operation(request):
 @api.patch("/path")
 def patch_operation(request):
     ...
+
+@api.event_source("/path")
+def event_source_operation(request):
+    ...
```
baseplate-admin, line 39: Could this be made async?

baseplate-admin (changes requested):
One small nitpick

ninja/testing/client.py
```diff
@@ -191,14 +195,35 @@ def __init__(self, http_response: Union[HttpResponse, StreamingHttpResponse]):
...
     def json(self) -> Any:
         return json_loads(self.content)
 
+    def __get_content(
+        self, streaming_content: Union[AsyncIterable[bytes], Iterable[bytes]]
+    ) -> bytes:
+        if isinstance(streaming_content, types.AsyncGeneratorType):
+            return self.__get_async_content(streaming_content)  # type: ignore
+        else:
+            return self.__get_sync_content(streaming_content)  # type: ignore
```
baseplate-admin, line 212: I think we can use [`asgiref.iscoroutinefunction`](https://django.readthedocs.io/en/stable/topics/async.html#async-views) instead of this, its the same approach `django` uses

docs/docs/guides/input/operations.md
rroblf01: Yes, it works with async

ninja/testing/client.py
rroblf01: You're right, I've already changed the code to use isasyncgen to check if it's an async generator.

docs/docs/guides/input/operations.md
baseplate-admin: My apologies, I meant can we refractor this to make it like,

```python
@api.event_source("/path")
async def event_source_operation(request):
    ...
```

baseplate-admin (changes requested):
Overall looks good, just one small change request

baseplate-admin:
Also, i have a suggestion, should we allow only async server events? Otherwise we might get complaints in the future about this `sync`/`async` approach.

docs/docs/guides/input/operations.md
rroblf01: It's possible, but Django's implementation already prevents that blocking, right?

rroblf01:
> Also, i have a suggestion, should we allow only async server events? Otherwise we might get complaints in the future about this `sync`/`async` approach.

I understand that both should be allowed. Even the StreamingHttpResponse works better when it is synchronous (I imagine it is due to the implementation underneath)

docs/docs/guides/input/operations.md
rroblf01: Although a StreamingHttpResponse with an async generator will only work properly if they are using asgi instead of wsgi.
Should we still make async the default in the documentation?

rroblf01:
Good morning @vitalik @baseplate-admin , What changes would be good to merge this PR? Do I put in the documentation that the method is asynchronous but with a warning that it should be used with asgi when it is asynchronous?

rroblf01:
hello @vitalik @baseplate-admin How could we continue with this PR?

baseplate-admin:
Apologies, i think @vitalik needs to give his final opinion on this PR.

> Do I put in the documentation that the method is asynchronous but with a warning that it should be used with asgi when it is asynchronous?

Sounds good to me, add the comment made by vitalik, something like 

```
Note that using sync SSE might block the worker, please use async SSE for best performance 
```

baseplate-admin:
Okay, thank you.

@vitalik will make the final call on this

rroblf01:
Hello @vitalik Could we merge this pr or do you think it is necessary to add something else?

onursoyer:
@rroblf01  what happened? Why did you delete your PR?

rroblf01:
Good morning @onursoyer , after five months of no response and much insistence, I understand that my PR is not necessary and that you have no interest in the project. It's best to close this PR and not let it generate noise among the PRs that are interested.

onursoyer:
@rroblf01 Sorry to hear that. It looks like you put significant effort into this PR. It seems that others, including myself, are interested in this PR as well.
@vitalik has probably just been too busy and overlooked this PR.

Rafterman1217:
Very sad, sorry that you felt like deleting your pr! :/

vitalik:
@rroblf01 @Rafterman1217 @onursoyer @baseplate-admin 

sorry for delay - well this feature was a bit tricky one - I wanted it to be for of all close to standards, easy to use(short and clean) and support not only GET methods (browsers only) but as well other methods (e.g. MCPs) 

on OpenAPI part I went the same way FastAPI did (just introducing schema that not really displayed on swagger/redoc) but hopefully soon will be

the usage is much simpler and clean

details here - https://github.com/vitalik/django-ninja/pull/1685 let me know if you see any issues

Issue #1391, opened 2025-01-14
I want to add parser in router


bamjun:
I can create a custom parser when create the `NinjaAPI` but I can't create a custom parser in routers. How can I do this ?

## problem
---
I want to add parser in router like blow.

```python
from ninja import NinjaAPI
from ninja.parser import JSONParser, FormParser

api = NinjaAPI()

@api.post("/json/", parser=JSONParser())
def handle_json(request, data: dict):
    return {"received_data": data}

@api.post("/form/", parser=FormParser())
def handle_form(request, name: str, age: int):
    return {"name": name, "age": age}
```

## I CAN DO
---
```python
from ninja.parser import Parser
from lxml import etree

class XMLParser(Parser):
    def parse_body(self, request):
        body = request.body.decode()
        root = etree.fromstring(body)
        return {child.tag: child.text for child in root}

from ninja import NinjaAPI

api = NinjaAPI(parser=XMLParser())

@api.post("/xml/")
def handle_xml(request, data: dict):
    return {"received_data": data}
```

## I WANT TO DO, BUT I CAN NOT
---
```python
from ninja import NinjaAPI
from ninja.parser import JSONParser, FormParser

api = NinjaAPI()

@api.post("/json/", parser=JSONParser())
def handle_json(request, data: dict):
    return {"received_data": data}

@api.post("/form/", parser=FormParser())
def handle_form(request, name: str, age: int):
    return {"name": name, "age": age}
```

PR #1392, opened 2025-01-14
Add parser support to Operation and Router classes

bamjun:
- Introduced a new optional `parser` parameter in the `Operation` and `Router` classes to allow for customizable request body parsing.
- Updated the `run` method in `Operation` to utilize the parser for handling request bodies.
- Modified various HTTP method decorators in `Router` to accept and pass the parser parameter to the associated operations.

#1391


vitalik:
for some reason GH did not run any workflow here - can you make another commit or something to force it ?

bamjun:
Did I do something wrong?

PR #1396, opened 2025-01-22
Default error schema 422 

begyy:
Added a default schema for 422 status codes, similar to FastAPI. The functionality ensures that a schema is only added if http422 is not already present in the response.

```python
if http422 not in result:
    model = operation._create_response_model_multiple({http422: DefaultValidationError})[http422]
    schema = self._create_schema_from_model(model)[0]
    result[http422] = {
        "description": "Validation error",
        "content": {self.api.renderer.media_type: {"schema": schema}},
    }

```

The function `operation._create_response_model_multiple({http422: DefaultValidationError})[http422]` is a method within the class that generates the model. If you know of a better approach to create this model without directly invoking the class method, please suggest improvements, and I will update the implementation accordingly. 🙂

### DIFF
--- a/ninja/openapi/schema.py
+++ b/ninja/openapi/schema.py
@@ -1,14 +1,14 @@
 import itertools
 import re
 from http.client import responses
-from typing import TYPE_CHECKING, Any, Dict, Generator, List, Optional, Set, Tuple
+from typing import TYPE_CHECKING, Any, Dict, Generator, List, Optional, Set, Tuple, Union
 
 from django.utils.termcolors import make_style
 
 from ninja.constants import NOT_SET
 from ninja.operation import Operation
 from ninja.params.models import TModel, TModels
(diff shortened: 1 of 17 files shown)


begyy:
image

ninja/openapi/schema.py
```diff
@@ -167,7 +178,7 @@ def _extract_parameters(self, model: TModel) -> List[DictStrAny]:
             p_schema: DictStrAny
             p_required: bool
             for p_name, p_schema, p_required in flatten_properties(
-                name, details, is_required, schema.get("$defs", {})
+                    name, details, is_required, schema.get("$defs", {})
```
luckydonald, line 181: probably best to not change the indent. This makes the diff bigger for no gain.

luckydonald (changes requested):
Probably it makes sense to revert all changes in indent, so only the relevant changes are part of the MR.

PR #1407, opened 2025-02-07
Validate responses from exception handlers to avoid inconsistent OpenAPI specification

matez0:
In the OpenAPI specification of the request handlers, it is possible to declare response schemas for different response status, but if the response is generated by an exception handler, it was not validated like normal responses.
This could lead to inconsistent OpenAPI specification and implementation.


matez0:
If the idea is accepted, does this change need to be mentioned in the documentation?

PR #1408, opened 2025-02-09
CI: Use Ubuntu 22 for 3.7 support in tests

pmdevita:
`ubuntu-latest` just swapped to Ubuntu 24, which drops support for Python 3.7. There should probably be a separate discussion about dropping about 3.7, but this will fix the problem for now.


vitalik:
thank you

PR #1409, opened 2025-02-11
Add method to check is user either staff or superuser.

adambirds:
Add method to check is user either staff or superuser.

### DIFF
--- a/ninja/security/__init__.py
+++ b/ninja/security/__init__.py
@@ -1,6 +1,6 @@
 from ninja.security.apikey import APIKeyCookie, APIKeyHeader, APIKeyQuery
 from ninja.security.http import HttpBasicAuth, HttpBearer
-from ninja.security.session import SessionAuth, SessionAuthSuperUser
+from ninja.security.session import SessionAuth, SessionAuthSuperUser, SessionAuthIsStaff
 
 __all__ = [
     "APIKeyCookie",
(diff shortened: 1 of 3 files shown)


adambirds:
Would be gerat if this method could be added @vitalik, let me know if it needs any amendments.

vitalik:
I think the check of is_superuser (and naming it so) is obsolete

in django world if user is superuser that it's automatically is staff and has all the permissions


https://github.com/django/django/blob/41239fe34d64e801212dccaa4585e4802d0fac68/django/contrib/auth/models.py#L384-L397


I would just call it IsStaff : 


```Python
class SessionAuthSuperUser(APIKeyCookie):
    "Reusing Django session authentication & verify that the user is a super user"

    param_name: str = settings.SESSION_COOKIE_NAME

    def authenticate(self, request: HttpRequest, key: Optional[str]) -> Optional[Any]:
        is_superuser = getattr(request.user, "is_superuser", None)
        if request.user.is_authenticated and is_superuser:
            return request.user

        return None


class SessionAuthSuperUserOrStaff(SessionAuthSuperUser):
    
def authenticate(self, request: HttpRequest, key: Optional[str]) -> Optional[Any]:
        result = super().authenticate(request, key)
        if result:
              return result
        is_staff = getattr(request.user, "is_sstaff", None)
        if request.user.is_authenticated and (is_superuser or is_staff):
            return request.user

        return None
```

adambirds:
@vitalik I've made those changes.

PR #1414, opened 2025-02-25
Improve Performance for Async Pagination

tim-hub:
django ninja is a good framework, when I work on the project, I found the pagination on async view is much lower than sync view for performance wise, like row 1 and row 3, so I have made some update for the pagination class.

| Type | Name                 | # Requests | # Fails | Average (ms) | Min (ms) | Max (ms) | Average size (bytes) | RPS  | Failures/s |
| ---- | -------------------- | ---------- | ------- | ------------ | -------- | -------- | -------------------- | ---- | ---------- |
| async with origin pagination  | /api/bff/trade       | 108        | 12      | 11294.07     | 5764     | 23045    | 38551.11             | 0.86 | 0.1        |
| async with new pagination change  | /api/bff/trade/async | 108        | 0       | 75.9         | 25       | 475      | 43370                | 0.86 | 0          |
| sync pagination  | /api/bff/trade/sync  | 108        | 0       | 193.67       | 24       | 2175     | 43370                | 0.86 | 0          |

In another word, if use the origin pagination on async view, it is actually does not bring any befenits for performance but slow it down. This PR is to resolve this.

### DIFF
--- a/ninja/pagination.py
+++ b/ninja/pagination.py
@@ -6,7 +6,6 @@ from typing import Any, AsyncGenerator, Callable, List, Optional, Tuple, Type, U
 
 from django.db.models import QuerySet
 from django.http import HttpRequest
-from django.shortcuts import aget_list_or_404
 from django.utils.module_loading import import_string
 from typing_extensions import get_args as get_collection_args
 
@@ -109,7 +108,7 @@ class LimitOffsetPagination(AsyncPaginationBase):
         offset = pagination.offset
         limit: int = min(pagination.limit, settings.PAGINATION_MAX_LIMIT)
         return {
-            "items": await aget_list_or_404(queryset[offset : offset + limit]),
+            "items": [obj async for obj in queryset[offset : offset + page_size]],
             "count": await self._aitems_count(queryset),
         }  # noqa: E203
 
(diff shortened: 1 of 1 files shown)


vitalik:
hi @tim-hub
could you run your test with `async for` and latest version from master ?

tim-hub:
> aget_list_or_404 seems aded only in 5.x django
> 
> maybe it just worth adding some await
> 
> ```python
> [obj async for obj in queryset[offset: offset + self.page_size]]
> ```
> 
> throwing 404 if empty here also does not make sense actually

This makes sense, I have updated it.

tim-hub:
> hi @tim-hub could you run your test with `async for` and latest version from master ?

I have run it again,

image

It basically kept the same pattern, that new async pagination( and sync pagination) performance better than origin async pagination without the changes.

not surprisingly `[obj async for]` and `aget_list_or_404` do not performance much different, because as we now`aget` is just a wrapper + 404. anyway, I agree 404 does not make sense here.

tim-hub:
latest commit to fix the failing test

vitalik:
@tim-hub 

i think if-statement should be other way around

```python
if isinstance(queryset, list):
```
to
```
if isinstance(queryset, QuerySet)
   do await
else: # can be list, tuple, string etc - that does not have alist
   ...
```

tim-hub:
> @tim-hub
> 
> i think if-statement should be other way around
> 
> ```python
> if isinstance(queryset, list):
> ```
> 
> to
> 
> ```
> if isinstance(queryset, QuerySet)
>    do await
> else: # can be list, tuple, string etc - that does not have alist
>    ...
> ```

I see, make sense

PR #1415, opened 2025-03-03
Update default json_dumps_params at renderers.py

ulissesjdeo:
I think this would be nice by default to save bandwhich.

### DIFF
--- a/ninja/renderers.py
+++ b/ninja/renderers.py
@@ -19,7 +19,7 @@ class BaseRenderer:
 class JSONRenderer(BaseRenderer):
     media_type = "application/json"
     encoder_class: Type[json.JSONEncoder] = NinjaJSONEncoder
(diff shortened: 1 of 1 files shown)


vitalik:
Hi @ulissesjdeo do you have any reference on what it does /

pmdevita:
I think what this would do is take out the unnecessary spaces left in the JSON. So 

```
{"a": 123, "b": 456}
```

would become

```
{"a":123,"b":456}
```

vitalik:
well I do not see much sense on saving spaces for bandwidth...

if if that the case for some users they can [overwrite default JSON renderer ](https://django-ninja.dev/guides/response/response-renderers/)

```python
class NoSpacesRenderer(JSONRenderer):
    json_dumps_params: Mapping[str, Any] = {'separators': (',', ':')}

api = NinjaAPI(renderer=NoSpacesRenderer())
```

PR #1419, opened 2025-03-09
Fix tests on Pydantic 2.11

pmdevita:
The CI tests automatically use the newest version of Pydantic and 2.11 introduced some changes to the JSON schema. Might be a good idea to set or limit the Pydantic version


vitalik:
@pmdevita 
Thank you
did not see that - already fixed

PR #1422, opened 2025-03-14
Add query parameters in request_params for testing

zhaozigu:
# Enhanced Support for Query Parameters in GET Requests
Currently, Django Ninja's client.get only supports query parameters directly embedded in the URL, such as `/customers/details/?name=fred&age=7` . 

However, it lacks support for passing query parameters as a separate dictionary like 
```python
c.get("/customers/details/", query_params={"name": "fred", "age": 7})
```

This pull request enhances the _build_request method to extract query parameters from the request_params dictionary, providing a more flexible and cleaner API for testing GET requests with query parameters. 

The implementation maintains backward compatibility with the existing URL-based approach while adding this new functionality.

Now we can implement code like this:

```python
from ninja.testing import TestClient

response = client.get("/customers/details/", query_params={"name": "fred", "age": 7})
```

### DIFF
--- a/ninja/testing/client.py
+++ b/ninja/testing/client.py
@@ -172,7 +172,11 @@ class NinjaClientBase:
             if query_params:
                 query_dict = QueryDict(mutable=True)
                 for k, v in query_params.items():
-                    query_dict[k] = v
+                    if isinstance(v, list):
+                        for item in v:
+                            query_dict.appendlist(k, item)
+                    else:
+                        query_dict[k] = v
                 request.GET = query_dict
             else:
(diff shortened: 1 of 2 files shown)


vitalik:
Hi @zhaozigu 

Nice, do you mind adding test case for this ?

zhaozigu:
> Hi @zhaozigu
> 
> Nice, do you mind adding test case for this ?

Haha, thanks for the reminder! I've added the test cases now, and everything stays consistent with the existing functionality, including support for parameters with the same name.

vitalik:
Thank you

PR #1423, opened 2025-03-14
Fix for request.FILES Population on Non-POST Methods

vitalik:
This PR addresses a longstanding [Django issue](https://code.djangoproject.com/ticket/12635) where request.FILES is only populated when the request method is POST, ignoring file uploads for other HTTP methods.

While this behavior may have aligned with how browsers handled file uploads in the past, modern browsers and HTTP clients allow file uploads with any method.

This PR introduces a check for non-POST methods that accept file uploads. If such a request is detected, Django-Ninja will raise an error with a suggestion to enable a specific middleware that corrects this behavior by ensuring request.FILES is properly populated.

can be disabled with `settings.NINJA_FIX_REQUEST_FILES_METHODS = []`

CC @baseplate-admin  (https://github.com/baseplate-admin/ninja_put_patch_file_upload_middleware)

### DIFF
--- a/ninja/compatibility/files.py
+++ b/ninja/compatibility/files.py
@@ -1,18 +1,12 @@
-from collections.abc import Awaitable, Callable
-from typing import Any, List, Union
+from typing import Any, List
 
 from asgiref.sync import iscoroutinefunction, sync_to_async
 from django.conf import settings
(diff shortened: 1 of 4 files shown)


baseplate-admin:
Thank you for including this.

All good from my side :D

PR #1439, opened 2025-04-08
fix: added missing await to get response  fix_request_files_middlewar…

caspel26:
fix_request_files_middleware is broken with async views i just fixed this issue adding a missing await


vitalik:
good catch !

Issue #1446, opened 2025-04-13
Option to exclude Django Field blank option from determination of 'nullable'


tomatod:
**Is your feature request related to a problem? Please describe.**
When blank=True is specified at a Djang Field, ModelSchema treats the field as nullable even if null=False is also specified there.  So, ModelSchema passes null to the Django model and it raise NOT NULL constraint exception.

```
class CharFieldVerification(models.Model):
    char_field = models.CharField(max_length=200, blank=True, null=False)

class TextFieldVerification(models.Model):
    text_field = models.TextField(blank=True, null=False)

class DateFieldVerification(models.Model):
    date_field = models.DateField(blank=True, null=False)

class FileFieldVerification(models.Model):
    file_field = models.FileField(blank=True, null=False)

class BooleanFieldVerification(models.Model):
    boolean_field = models.BooleanField(blank=True, null=False)
```

![Image](https://github.com/user-attachments/assets/c85a357d-1fd2-457c-b905-67a9e9998bbc)

The combination of blank=True and null=False seems to be common in string-based Django fields, which is mentioned in [Django Model field reference](https://docs.djangoproject.com/en/5.2/ref/models/fields/#null)

**Other related Issue**
- https://github.com/vitalik/django-ninja/issues/882

tomatod:
According to the [Django reference](https://docs.djangoproject.com/en/5.2/ref/models/fields/#blank), the blank option is mainly for form validation. Django is originally made for form application and target content-type is application/x-www-form-urlencoded, so I think DjangoField has the blank option because the content-type cannot express null explicitly. Json can express null originally, so I think as an option it would be good if ModelSchema could simply look only at the null option.

PR #1447, opened 2025-04-13
Add option to exclude Django Field blank option from determination of 'nullable'

tomatod:
# Summary
This is related to #1446. I added an option (is_blank_nullable) as an attribute to ModelSchema, which will be used to determine whether a blank value is considered when deciding a field's nullability.

I have taken compatibility with previous implementations into consideration. The default behavior remains unchanged unless is_blank_nullable is explicitly set to False.

To ensure compatibility, I first added tests that check the current behavior in the first commit. Then, in the second commit, I introduced the new option.

# Note
I read CONTRIBUTING.md and run the commands (make test, make lint, etc.), but make test-cov is failed. It seems to has been already failed from previous release, so I ignored for now. Off course, the coverage of the new parts of this PR is checked.

I'll be glad to have any suggestions. If there are many points that should be considered and fixed, I can remake this PR anew.

Thanks,

### DIFF
--- a/ninja/orm/factory.py
+++ b/ninja/orm/factory.py
@@ -41,6 +41,7 @@ class SchemaFactory:
         fields: Optional[List[str]] = None,
         exclude: Optional[List[str]] = None,
         optional_fields: Optional[List[str]] = None,
+        is_blank_nullable: bool = True,
         custom_fields: Optional[List[Tuple[str, Any, Any]]] = None,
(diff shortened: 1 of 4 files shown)


tomatod:
I got to think this sollution effecting all kind of field is not better after thinking about that again, because Django assumes the combination of blank=True and null=False is only for a few kinds of field. But continually, I want know good solutions if any.

PR #1450, opened 2025-04-17
Add PatchSchema

adriaanwm:
Disclaimer: The code in this PR was written entirely with AI. Based on my tests it is working. However, I am not very familiar with Pydantic.

Suppose you have the following schema:

```
class MyProfileIn(Schema):
    bio: str
    email: Optional[str]
```

If you use PatchDict[MyProfileIn], then the following patch body passes validation:

```
{ "bio": null, "email": "example@example.com" }
```

This should return a 422, because if bio is included, it should be a string. This will cause issues when updating models where non-nullable fields are being set to null.

For this reason I made PatchSchema (`PatchSchema[MyProfileIn]`) which would result in the following:

```
{ "bio": null, "email": "example@example.com" }  # 422
{ "email": null }  # 200
{ "bio": "Example" }  # 200
{ "bio": "Example", "email": null }  # 200
```


vitalik:
@adriaanwm - we already have [PatchDict](https://django-ninja.dev/guides/response/django-pydantic/#patchdict)

adriaanwm:
@vitalik PatchDict allows `null` to be provided for a field where `null` is not a valid value.

if you have

```
class MyProfileIn(Schema):
    bio: str
```

then `PatchDict[MyProfileIn]` means that this payload passes validation:

```
{ 'bio': null }
```

Ideally this should return a 422 because null is not valid for bio (if it is included). This is why I made `PatchSchema`.

vitalik:
I would say this is questinoable/undefinded behaviour - sending  `{ 'bio': null }` via PATCH request clearly tells that client wants to set some value as None

anyway - I would look into extending `PatchDict` with some option instead of introducing class with very similar name and a bit different behaviour

Issue #1452, opened 2025-04-18
[BUG] Ninja uses a deprecated function to override the UUID converter in django


bulldogukbcm:
**Describe the bug**
When upgrading to Django 5.1, I'm getting the following warnings:

``` 
File "/home/vscode/.local/lib/python3.11/site-packages/ninja/signature/utils.py", line 87, in <module>
    register_converter(NinjaUUIDConverter, "uuid")
  File "/home/vscode/.local/lib/python3.11/site-packages/django/urls/converters.py", line 62, in register_converter
    warnings.warn(
django.utils.deprecation.RemovedInDjango60Warning: Converter 'uuid' is already registered. Support for overriding registered converters is deprecated and will be removed in Django 6.0.
```
This warning is blocking my pytests from running - is it on the roadmap to be addressed?

**Versions (please complete the following information):**
 - Python version: 3.11.12
 - Django version: 5.1.8
 - Django-Ninja version: 1.4.1
 - Pydantic version: 2.11.3

Note you can quickly get this by runninng in `./manage.py shell` this line:
```
import django; import pydantic; import ninja; django.__version__; ninja.__version__; pydantic.__version__
```

vitalik:
duplicate #1266

Issue #1266, opened 2024-08-18
[BUG] Converter `uuid` is already registered - deprecated in django 6.0


mathiasertl:
**Describe the bug**

When running `python -Wd ca/manage.py check` I get a deprecation warning (paths trimmed for privacy):

```
user@host:~/src/...$ python -Wd ca/manage.py check
.../lib/python3.12/site-packages/ninja/signature/utils.py:96: RemovedInDjango60Warning: Converter 'uuid' is already registered. Support for overriding registered converters is deprecated and will be removed in Django 6.0.
  register_converter(NinjaUUIDConverter, "uuid")
```

django-ninja registers a converter called `uuid`, which is already registered by Django. Support for overriding a converter will be removed in Django 6.0. This converter is also problematic as it alters the behavior of views unrelated to django-ninja: A standard Django view using the `uuid` converter will receive a different type of value depending on if django-ninja is used or not - this was also already raised in #280.

From the looks of it, I think the converter can be safely removed. If not, it has to be renamed. Please advice on your preferred course of action, I'm happy to make a PR.

**Versions (please complete the following information):**
 - Python version: 3.12 (should not matter)
 - Django version: 5.1
 - Django-Ninja version: 1.3.0
 - Pydantic version: 2.8.2 (should not matter)

Xdynix:
Sharing a monkeypatch I used to suppress this behavior. The behavior of the `uuid` converter has always been to return a UUID instance since Django 2.0. I don't really think it should behave differently in Django Ninja than the `int` converter (i.e. returning `str` instead of the actual type).

Add the following code to your `settings.py`.

```python
def monkeypatch_ninja_uuid_converter() -> None:
    import importlib
    import sys

    import django.urls

    module_name = "ninja.signature.utils"
    sys.modules.pop(module_name, None)

    original_register_converter = django.urls.register_converter

    def fake_register_converter(*_: Any, **__: Any) -> None:
        pass

    django.urls.register_converter = fake_register_converter
    importlib.import_module(module_name)

    django.urls.register_converter = original_register_converter


monkeypatch_ninja_uuid_converter()
```

lapinvert:
Are we going to fix this before the release of django 6? otherwise django-ninja will stop being compatible with latest version of django.

antoniogamizbadger:
Any update on this? :D

ncrozz:
I have problems also with the workaround by @Xdynix, Django can't start at the moment exiting with Pydantic warning.
[...]

PR #1453, opened 2025-04-19
uuidstr url converter instead of shadowing django uuid

vitalik:
Before django ninja shadowed django's uuid to be able to to get string value from url:

```python

@api.get('/some/{uuid:id}')
def some(request, id):
     return {'id': id, 'type': str(type(id))}  # type=str
```

Now added custom converted `uuidstr`

```python
@api.get('/some/{uuidstr:id}')
def some(request, id):
    # id will be str type here
```

django 6.0 will not allow shadowing 

addresses #1452 #1266

### DIFF
--- a/ninja/router.py
+++ b/ninja/router.py
@@ -1,3 +1,4 @@
+import re
 from typing import (
     TYPE_CHECKING,
     Any,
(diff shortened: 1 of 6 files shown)


ninja/router.py
```diff
@@ -320,6 +320,13 @@ def add_api_operation(
         include_in_schema: bool = True,
         openapi_extra: Optional[Dict[str, Any]] = None,
     ) -> None:
+        if "{uuid:" in path:
```
baseplate-admin, line 323: Do you think a regex based check would be better?

vitalik: @baseplate-admin good question... 

wondering if django allows doing stuff like `/path/<UUID:var>` or `/path/< uuid :var>` (space)

baseplate-admin: [I dont think django allows that.](https://chatgpt.com/share/6803ab9c-bb60-8011-a9c4-e4322d111cbe)

By default, Django's URL dispatcher is case-sensitive.

baseplate-admin: But i still think that regex based check is far better. It helps with edge cases.

For example what if someone does: `{uuid::` // Note the double ":"

vitalik: ok, just in case replaced it with ignore case regex

PR #1457, opened 2025-05-02
fix _run_authentication creating the coroutine result twice

eugenenelou:
This fix a problem where `_run_authentication` in sync mode and async callback instantiate the coroutine twice and awaits the second one.

This raises a "coroutine not awaited" warning, which is annoying.


ninja/operation.py
```diff
@@ -207,7 +207,11 @@ def _run_authentication(self, request: HttpRequest) -> Optional[HttpResponse]:
...
                     result = callback(request)
                     if inspect.iscoroutine(result):
-                        result = async_to_sync(callback)(request)
+
+                        async def await_result(cor: Coroutine) -> Any:
+                            return await cor
+
+                        result = async_to_sync(await_result)(result)
                 else:
                     result = callback(request)
```
matez0, line 216: In the `try` block,
```python
                result = callback(request)

                    async def await_result(cor: Coroutine) -> Any:
                        return await cor

                    result = async_to_sync(await_result)(result)
```
would not be enough since `result` is created the same way in all cases and we may want to await `result` independently of the properties of `callback`?

vitalik:
@eugenenelou could you show example code(authenticator?) that triggers  "coroutine not awaited" warning ?

Ideally include it tests

PR #1467, opened 2025-05-24
[WIP] fix: OpenAPI schema with duplicated payload name

lucasrcezimbra:
I couldn't find any issue/PR reporting this.

When we have two schemas with the same name, only one goes to the OpenAPI schema. I added a test case to show the issue, but I don't know the expected behavior. Options: 

1. Add a number in the end Payload, Payload2, PayloadN.
1. Raise an exception during the startup.
1. Add a warning.
1. Any other ideas?


OscarVanL:
I have ran into the same issue too.

PR #1476, opened 2025-06-02
Added Message_JSONRenderer for enriched JSON API responses

mammhoud:
## What’s New

- Introduced `Message_JSONRenderer`, which wraps API responses in a standard format including:
  - `status`: HTTP status code
  - `message`: human-readable summary based on status
  - `data`: actual payload

## Why?

This provides a consistent response structure for frontend consumption, making debugging and UX improvements easier.

## Example Output:
```json
{
    "status": 200,
    "message": "Success",
    "data": { ... }
}

### DIFF
--- a/ninja/renderers.py
+++ b/ninja/renderers.py
@@ -3,7 +3,7 @@ from typing import Any, Mapping, Optional, Type
 
 from django.http import HttpRequest
 
-from ninja.responses import NinjaJSONEncoder
+from ninja.responses import *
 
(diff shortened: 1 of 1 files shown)


vitalik:
This doesn’t seem like something that should be part of the framework itself, as developers often have varying preferences for JSON response formats and wording conventions.

And it's always possible to make custom renderer like you did inside project

mammhoud:
yes sure with my pleasure for your answer, the renderer function it self based on the programmer needs, i with your side.

PR #1478, opened 2025-06-06
fix(auth): fix sync authentication with async operations

### LINKED ISSUE #1251: [BUG] Sync-only Authentication Callbacks not Working on Async Operations
**Describe the bug**

If an authentication callback can only works in sync context, then it will not work on async views.

Example Code:

```python
from ninja.security import django_auth

api = NinjaAPI()


@api.get("/foobar", auth=django_auth)
async def foobar() -> str:
    return "foobar"
```

Accessing the view will raise: `django.core.exceptions.SynchronousOnlyOperation: You cannot call this from an async context - use a thread or sync_to_async.`

This is because accessing `request.user` involves DB query (when using Django's default session engine) and can only be run in sync context. But in `AsyncOperation._run_authentication()` it didn't switch the context for the authentication callback.

 - Python version: [e.g. 3.6]: 3.12
 - Django version: [e.g. 4.0]: 5.0.7
 - Django-Ninja version: [e.g. 0.16.2]: 1.2.2
 - Pydantic version: [e.g. 1.9.0]: 2.8.2

Note you can quickly get this by runninng in `./manage.py shell` this line:
```
import django; import pydantic; import ninja; django.__version__; ninja.__version__; pydantic.__version__
```

bellini666:
Fix #1251


bellini666:
Hi @vitalik , asking for your kind review 🙏

yorickr:
Hello @vitalik, I understand you must get a lot of review requests. Just checking in, is there any way we would be able to help out?

vitalik:
Thank you @bellini666

bellini666:
Hi @vitalik , any idea when a new version is coming out? Looking forward to be able to use this

Issue #1324, opened 2024-10-22
[BUG] PatchDict errors with inherited schemas


filippomc:
**Describe the bug**

I have a schema hierarchy such as:

```
class ViewableContent(Schema):
    name: str
    description: str = None

class MySchema(ViewableContent):
   other: str # If I don't add a new field the problem does not arise
```

Then add a router like the following:

```
@router.patch('/{uuid}', response={200: MySchema})
@transaction.atomic
def update_study(request: HttpRequest, uuid: str, payload: PatchDict[MySchema]):
   ...
```

When I run my application the following error is raised:

```
  File "/home/user/mnp/applications/neuroglass-research/backend/neuroglass_research/api/__init__.py", line 4, in <module>
    from .studies import router as studies_router
  File "/home/user/mnp/applications/neuroglass-research/backend/neuroglass_research/api/studies.py", line 69, in <module>
    def update_study(request: HttpRequest, study_id: int, payload: PatchDict[UpdateStudyPayload]):
                                                                   ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
  File "/home/user/miniconda3/envs/mnp/lib/python3.12/site-packages/ninja/patch_dict.py", line 45, in __getitem__
    new_cls = create_patch_schema(schema_cls)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/miniconda3/envs/mnp/lib/python3.12/site-packages/ninja/patch_dict.py", line 29, in create_patch_schema
    t = schema_cls.__annotations__[f]
        ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
KeyError: 'name'
```

**Versions (please complete the following information):**
 - Python version: 3.12
 - Django version: 5.1.2
 - Django-Ninja version: 1.3.0
 - Pydantic version: 2.9.2

vitalik:
> If I don't add a new field the problem does not arise

so problem only appears when you ADD some field ? or always ?

filippomc:
If I don't add a field it's like Pydantic is considering the two classes equivalent and what I see is that the annotations are the same from the base class. When adding a field, that field is the only one within the __annotations__ dict.

vitalik:
well I still do not understand you.. can you show two examples working and non working or something...

filippomc:
The not working example is the one I originally posted, and the application breaks on any path with the error above.
These are a few ones that are working for me:

Working -- no inheritance:
```python
[...]

PR #1480, opened 2025-06-10
fix PatchDict errors with inherited schemas #1324

LeandroDeJesus-S:
### Problematic Scenario

Consider the following schemas:

```python
class SomeSchema(Schema):
    some_field: str

class OtherSchema(SomeSchema):
    other_field: str
```

When using these in a route, the method `ninja.patch_dict.create_patch_schema` relies solely on the `__annotations__` attribute of the schema class. This means it only considers the annotations declared in the current schema, ignoring inherited fields. This leads to the following error:

```python
@router.patch('/{uuid}', ...)
def endpoint(request, uuid: str, payload: PatchDict[OtherSchema]):
    ...
    
# Error:
# File ".../ninja/patch_dict.py", line 45, in __getitem__
#   new_cls = create_patch_schema(schema_cls)
#             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
# File ".../ninja/patch_dict.py", line 29, in create_patch_schema
#   t = schema_cls.__annotations__[f]
#       ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
# KeyError: 'some_field'
```

### Proposed Solution

To address this issue, I created a helper function that recursively collects annotations from all parent classes. It traverses the class hierarchy using the `__base__` attribute until it reaches a base schema class (`Schema`, `ModelSchema`, or `BaseModel`), then merges all the collected `__annotations__` in reverse order to preserve field overrides from the child class.

```python
def get_schema_annotations(schema_cls: Type[Any]) -> Dict[str, Any]:
    schema = schema_cls
    schemas = []
    excluded_bases = (
        "<class 'ninja.schema.Schema'>",
        "<class 'ninja.orm.ModelSchema'>",
        "pydantic.BaseModel",
    )
    while True:
        schemas.append(schema)
        schema = schema.__base__
        if str(schema) in excluded_bases:
            break

    annotations = {}
    for schema in reversed(schemas):
        annotations.update(schema.__annotations__)

    return annotations
```

### DIFF
--- a/ninja/patch_dict.py
+++ b/ninja/patch_dict.py
@@ -1,4 +1,12 @@
-from typing import TYPE_CHECKING, Any, Dict, Generic, Optional, Type, TypeVar
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Dict,
+    Generic,
+    Optional,
+    Type,
+    TypeVar,
+)
 
 from pydantic_core import core_schema
 
@@ -22,10 +30,32 @@ class ModelToDict(dict):
         return input_value.model_dump(**cls._wrapped_model_dump_params)
 
 
+def get_schema_annotations(schema_cls: Type[Any]) -> Dict[str, Any]:
+    schema = schema_cls
+    schemas = []
+    excluded_bases = (
+        "<class 'ninja.schema.Schema'>",
+        "<class 'ninja.orm.ModelSchema'>",
+        "pydantic.BaseModel",
+    )
+    while True:
+        schemas.append(schema)
+        schema = schema.__base__
+        if str(schema) in excluded_bases:
+            break
+
+    annotations = {}
+    for schema in reversed(schemas):
+        annotations.update(schema.__annotations__)
+
+    return annotations
+
+
 def create_patch_schema(schema_cls: Type[Any]) -> Type[ModelToDict]:
+    schema_annotations = get_schema_annotations(schema_cls)
     values, annotations = {}, {}
     for f in schema_cls.__fields__.keys():
-        t = schema_cls.__annotations__[f]
+        t = schema_annotations[f]
         if not is_optional_type(t):
             values[f] = getattr(schema_cls, f, None)
             annotations[f] = Optional[t]
--- a/tests/test_patch_dict.py
+++ b/tests/test_patch_dict.py
@@ -1,4 +1,4 @@
-from typing import Optional
+from typing import List, Optional
 
 import pytest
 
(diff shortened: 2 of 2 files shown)


ninja/patch_dict.py
```diff
@@ -22,10 +30,32 @@ def _validate(cls, input_value: Any) -> Any:
         return input_value.model_dump(**cls._wrapped_model_dump_params)
 
 
+def get_schema_annotations(schema_cls: Type[Any]) -> Dict[str, Any]:
+    schema = schema_cls
+    schemas = []
+    excluded_bases = (
+        "<class 'ninja.schema.Schema'>",
+        "<class 'ninja.orm.ModelSchema'>",
+        "<class 'pydantic.main.BaseModel'>",
```
vitalik, line 39: I think you can use actual classes here instead of their repr (which may break in some potential new version)

ninja/patch_dict.py
```diff
@@ -22,10 +33,32 @@ def _validate(cls, input_value: Any) -> Any:
         return input_value.model_dump(**cls._wrapped_model_dump_params)
 
 
+def get_schema_annotations(schema_cls: Type[Any]) -> Dict[str, Any]:
+    schema = schema_cls
```
vitalik, line 37: what you do here with `while True` you can do with `class.__mro__`
some thing like this: 
```python
def get_schema_annotations(schema_cls: Type[Any]) -> Dict[str, Any]:
    excluded_bases = {Schema, ModelSchema, BaseModel}
    annotations = {}

    for base in reversed(schema_cls.__mro__):
        if base in excluded_bases:
            break
        annotations.update(getattr(base, '__annotations__', {}))

    return annotations
```

LeandroDeJesus-S: This approach doesn’t work because when calling `reversed(schema_cls.__mro__)`, it will always return an empty `annotations` dictionary. That’s because it will hit one of the excluded classes first:

```py
class SomeSchema(Schema):
    name: str
    age: int
    category: Optional[str] = None

# reversed(SomeSchema.__mro__)
# [<class 'object'>, <class 'pydantic.main.BaseModel'>, <class 'ninja.schema.Schema'>, <class 'test_patch_dict.SomeSchema'>]
```

What do you think of this solution?
I also believe it would be a good idea to move it to the `ninja.utils.py` file:

```py
def get_schema_annotations(schema_cls: Type[Any]) -> Dict[str, Any]:
    annotations: Dict[str, Any] = {}
    excluded_bases = {Schema, ModelSchema, BaseModel}
    bases = schema_cls.mro()[:-1]
    final_bases = reversed([b for b in bases if b not in excluded_bases])

    for base in final_bases:
        annotations.update(getattr(base, '__annotations__', {}))

    return annotations
```
If you think this looks good, I’ll go ahead and move it and add more test cases.

vitalik:
Thank you @LeandroDeJesus-S

PR #1491, opened 2025-06-22
Add type annotations for `request` in `paginate_queryset`

brianhelba:
This updates the docs and internal code to expect that `request` is always available, which now allows downstreams to rely on this fact.

This is fully backwards-compatible with existing code which does not explicitly expect a `request` argument to `paginate_queryset`; since `request` is annotated (and invoked) as keyword-only, it will always be captured in `**params` as a fallback.

The test case with `CustomPagination` now explicitly tests this backwards-compatibility by using the old-style signature.

### DIFF
--- a/docs/docs/guides/response/pagination.md
+++ b/docs/docs/guides/response/pagination.md
@@ -118,6 +118,7 @@ To create a custom pagination class you should subclass `ninja.pagination.Pagina
 Example:
 
 ```python hl_lines="7 11 16 26"
+from django.http import HttpRequest
 from ninja.pagination import paginate, PaginationBase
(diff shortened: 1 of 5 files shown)


brianhelba:
This completes the changes made in a718d73c75dd49c1717389e05c1df69b6fa6ade5.

brianhelba:
@vitalik PTAL.

brianhelba:
@vitalik I've rebased this to include changes from #1584; note that PR also assumes that `paginate_queryset` definitely has a `request`, so I think this PR is even more important to formalize.

Please take a look.

Issue #1335, opened 2024-11-11
Pyright call issue errors for Field and Query


krims0n32:
```python
from ninja import Query, Field

class InstrumentFilter(FilterSchema):
    symbol: Optional[str] = Field(None, q='symbol__icontains')
```

Here I get "No parameter named 'q'" on Field

```python
@router.get('/instruments', response=List[InstrumentSchema])
@paginate
def instruments(request, filters: InstrumentFilter = Query(...)):
    ...
```

And here I get "Object of type Annotated is not callable" on Query.

I followed the docs on filtering. Am I doing something wrong here, missing a package?

Thanks!

sunfkny:
## No parameter named 'q'
`ninja.Field` is actually `pydantic.Field`, pydantic V2 Field extra kwargs is deprecated.

https://docs.pydantic.dev/2.9/migration/#changes-to-pydanticfield
https://github.com/pydantic/pydantic/blob/8bad227e7557519307a9ea9b0ca129aca1997667/pydantic/fields.py#L1025-L1033
https://github.com/pydantic/pydantic/blob/8bad227e7557519307a9ea9b0ca129aca1997667/pydantic/fields.py#L685-L686

So this will fix this typing error
```python
class InstrumentFilter(FilterSchema):
    symbol: str | None = Field(None, json_schema_extra={"q": "symbol__icontains"})
```

## Object of type Annotated is not callable
https://github.com/vitalik/django-ninja/blob/master/docs/docs/whatsnew_v1.md#shorter--cleaner-parameters-syntax

You can use shorter parameters syntax to fix this typing error

```python
@router.get("/instruments", response=List[InstrumentSchema])
@paginate
def instruments(
    request,
    filters: Query[InstrumentFilter],  # <--- !!!
): ...
```

krims0n32:
Thanks, that fixes the Pyright complaints. I take it the NinjaAPI docs on this topic need a little update then?

marianhlavac:
Came across the same issue.

> I take it the NinjaAPI docs on this topic need a little update then?

Yes, definitely.

PR #1494, opened 2025-06-25
Use json_schema_extra to add FilterSchema options to Field

shaleh:
As documented in https://github.com/vitalik/django-ninja/issues/1335, Pydantic is restricting use of additional parameters to `Field`. See https://docs.pydantic.dev/2.9/migration/#changes-to-pydanticfield for more details.

Move `q` and `ignore_none` into `json_extra_schema` as dictionary arguments.

Also, `Query(...)` does not get along with type checkers so refresh to use the v1 `Query[]` style to be consistent.

### DIFF
--- a/docs/docs/guides/input/filtering.md
+++ b/docs/docs/guides/input/filtering.md
@@ -22,7 +22,7 @@ class BookFilterSchema(FilterSchema):
 Next, use this schema in conjunction with `Query` in your API handler:
 ```python hl_lines="2"
 @api.get("/books")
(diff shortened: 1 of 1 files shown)


l1b3r:
FYI I have submitted a PR which attempts to solve this issue among other things: https://github.com/vitalik/django-ninja/pull/1514

vitalik:
Hi @shaleh - thank you for contribution - we are going in favour of #1514

PR #1497, opened 2025-06-28
pass-through for Schema-based results

andreasnuesslein:
When the response-parameter is filled and the result is already a Schema we don't need to verify it again.

Brief reasoning for the situation:
Wagtail has a [new guide how to integrate Ninja](based on https://docs.wagtail.org/en/v7.0/advanced_topics/api/django-ninja.html).
In it they use this
```python
@api.get("/pages/{page_id}/", response=BlogPageSchema | HomePageSchema)
def get_page(request: "HttpRequest", page_id: int):
    return get_object_or_404(Page, id=page_id).specific
```

This works but has its issues since basically the returned Page type (the django Model) is tried to be fitted against all different `response`-Schemata. Obviously that's a performance issue but it brings other problems too, which I dont want to go too deep into.

So my solution would be:
```python
@api.get("/pages/{page_id}/", response=BlogPageSchema | HomePageSchema)
def get_page(request: "HttpRequest", page_id: int):
    page = get_object_or_404(Page, id=page_id).specific
    if isinstance(page,BlogPage):
        return BlogPageSchema.from_orm(page,context={"request":request})
    return HomePageSchema.from_orm(page,context={"request":request})
```

this does currently not work, because the result is being re-parsed, because I have `response` set in the decorator. Obviously I could remove that but then I'd lose the types.

This PR allows Schemas to be directly returned, assuming that the validation already has happened.

Thanks so much

### DIFF
--- a/ninja/operation.py
+++ b/ninja/operation.py
@@ -275,26 +275,33 @@ class Operation:
             # Empty response.
             return temporal_response
 
-        resp_object = ResponseObject(result)
-        # ^ we need object because getter_dict seems work only with model_validate
-        validated_object = response_model.model_validate(
-            resp_object, context={"request": request, "response_status": status}
+        model_dump_kwargs: Dict[str, Any] = dict(
+            by_alias=self.by_alias,
+            exclude_unset=self.exclude_unset,
+            exclude_defaults=self.exclude_defaults,
+            exclude_none=self.exclude_none,
         )
-
-        model_dump_kwargs: Dict[str, Any] = {}
         if pydantic_version >= [2, 7]:
             # pydantic added support for serialization context at 2.7
             model_dump_kwargs.update(
                 context={"request": request, "response_status": status}
             )
 
-        result = validated_object.model_dump(
-            by_alias=self.by_alias,
-            exclude_unset=self.exclude_unset,
-            exclude_defaults=self.exclude_defaults,
-            exclude_none=self.exclude_none,
-            **model_dump_kwargs,
-        )["response"]
+        if isinstance(result, Schema):
+            # if the result is already a Schema, just return it
+            return self.api.create_response(
+                request,
+                result.model_dump(**model_dump_kwargs),
+                temporal_response=temporal_response,
+            )
+
+        resp_object = ResponseObject(result)
+        # ^ we need object because getter_dict seems work only with model_validate
+        validated_object = response_model.model_validate(
+            resp_object, context={"request": request, "response_status": status}
+        )
+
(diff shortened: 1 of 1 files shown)


ralphdelfs:
+1 This would be helpful

vitalik:
I guess this also need a validation taht returned object is one of the responses (e.g BlogPageSchema | HomePageSchema )

andreasnuesslein:
hi @vitalik 

thank you for looking into this.

to be honest, i dont understand :face_with_peeking_eye: 

isn't that exactly what the `response=BlogPageSchema | HomePageSchema` is doing? I'm pretty sure it was cratering when I return anything else

vitalik:
> isn't that exactly what the `response=BlogPageSchema | HomePageSchema` is doing? I'm pretty sure it was cratering when I return anything else

hm migth be - le'ts also add a test to cover :  
 - return instances taht is not of type from `resopnse`
 - I guess also pydantic basemodel (i think generally code shuld check isinstnace pydantic.BaseModel - not only Schema)

andreasnuesslein:
I changed it to `pydantic.BaseModel`. 

Can you give me a pointer where best to add a test for this?

Cheers

andreasnuesslein:
forgot to mention: i added a test 👍 ready to be merged i think

andreasnuesslein:
hi @vitalik , hope you're okay

just wanted to check if there's something else missing here? if not, could you maybe merge? cheers

andreasnuesslein:
hi @vitalik I think you solved it yourself here: https://github.com/vitalik/django-ninja/pull/1684   so I think we can close this PR.

Can't really verify just yet because of other migration issues

andreasnuesslein:
after a bit of debugging it turns out that this bit: https://github.com/vitalik/django-ninja/pull/1684/changes#diff-bdb1bcc28c7e3fb05084b142b442c8ed70d4ec6f08d24626508bf4047189f316R329

seems to not work with my `Union` type after all :'( 

Namely:
```python
        resp_annotation = response_model.model_fields["response"].annotation
```

is not returning a "type" type. `isinstance(resp_annotation, type)` fails.

I'm still patching it with this:
```python
                # Skip re-validation for pydantic model instances matching the response type
                if isinstance(result, pydantic.BaseModel):
                    return self.api.create_response(
                        request,
                        result.model_dump(**model_dump_kwargs),
                        temporal_response=temporal_response,
                    )
```

Do you need the higher specificity?  I tried to convince your code to accept my Union but the `typing` library seems not to detect the dynamically created Union sadly.

Issue #976, opened 2023-12-03
Support for middleware


jellis18:
**Is your feature request related to a problem? Please describe.**

I know that `django-ninja` is based off of django, obviously, but I'm really not a fan of the way django handles middleware where it applies globally and is somewhat hard to configure since you have to pass in the path to a callable as a string instead of being able to define it in code. 

To me it is always a pain (not just ninja, but django in general) when I want certain middleware per route or even per specific endpoint. I know we can do it with decorators or by subclassing router and using `view_decorator` now but all of those seem a bit tedious.

It seems that django-ninja is missing a key middleware feature like in FastAPI `add_middleware` and express with `use` or something like go Chi `use` as well. In all of those systems you can add middleware directly from the router without needing to use a decorator or equivalent.

**Describe the solution you'd like**
I've already started prototyping this and its somewhat based on how [go Chi](https://github.com/go-chi/chi/blob/master/README.md#examples) does it. Basically what I am proposing is that middleware be defined like
```python
api = NinjaAPI()

api.use(some_middleware)
# or
api.use(some_middleware1, some_middleware2)
```
This would add global middleware. To add router based middleware would look like
```python
router = Router()
router.use(some_router_middleware)
# or
router.use(some_router_middleware1, some_router_middleware2)
```
In this case the middleware from `api` would be inherited and the router would then have its own that comes after the `api` middleware

We could go further and even have endpoint level middleware that could look like
```python
router = Router()
...

@outer.with(some_endpoint_middeware).get(...)
def some_endpoint_handler(request: HttpRequest):
   ...
```
In all cases the middleware would have a signature that obeys
```python
_P = ParamSpec("_P")
ViewFunction = Callable[Concatenate[HttpRequest, _P], HttpResponse]

MiddlewareFunction = Callable[[ViewFunction], ViewFunction]
```
Which all standard django decorators would already obey.

Again, I know there are ways to accomplish the same goal already but this could be really useful and would allow users to more easily add their own middlware without having to subclass router or add decorators everywhere.

What are your thoughts on this. I'd be happy to work on this as I've already started but just want to see what the community thinks.

jkgenser:
I like this very much. I also come from FastAPI so maybe I'm biased

jellis18:
[...]

PR #1500, opened 2025-07-02
Add middleware support for API, Router, and Operation classes

nikatlas:
- Introduced middleware chains for enhanced request processing in sync/async operations.
- Updated `NinjaAPI`, `Router`, and `Operation` to handle middleware configurations.
- Added middleware tests for API and Router setups.

Tackles Issue: https://github.com/vitalik/django-ninja/issues/976

### DIFF
--- a/ninja/router.py
+++ b/ninja/router.py
@@ -393,11 +393,11 @@ class Router:
                 url_name = getattr(operation, "url_name", "")
                 if not url_name and self.api:
                     url_name = self.api.get_operation_url_name(operation, router=self)
-
+                view_func = path_view.get_view()
                 yield django_path(
                     route,
-                    functools.partial(
-                        path_view.get_view(), middlewares=self.middlewares
+                    functools.wraps(view_func)(
+                        functools.partial(view_func, middlewares=self.middlewares)
                     ),
                     name=url_name,
                 )


nikatlas:
@vitalik Please check a POC for middleware support. Added tests for both sync/async. 
Linked issue: https://github.com/vitalik/django-ninja/issues/976

nikatlas:
I added back a removed `type ignore` that was locally reported. seems its needed for older py versions. 
Please re-run. @vitalik

vitalik:
Hi @nikatlas 

Thank you for the time and effort !


I'm actually sitting and thinking on this for a while :)

Baiscally we need some proof of why it should be added and then document it

My concern is that there are currently 3 ways to add some "middleware"-like callbacks

`1)` Django Middlewares - basically standard thing but does not give good control per-endpoint  (and does not have access to validated arguments)

`2)` decorator

```python
@api.get('/some')
@mydeco
def some_method(request, a: int, b: string):
    ...
```
have access to validated data but  only called after validation (and cannot do anything on pre-validation layer))


`3)` [decorate_view](https://django-ninja.dev/whatsnew_v1/?h=decorate_view#decorators)

```python
@api.get("/test")
@decorate_view(cache_page(5))
def test_view(request):
    return str(datetime.now())
```

called before validation, can support all sort of built-in decorators that applied to standard django view functions


---

now what this PR solves is gives a way to apply some wrapper in bulk (like on entire router)

good use cases - apply caching on all get requests for router, or specific permission system check

what I think is lacking is behavior on where I want to apply this middleware before validation or after..

like maybe adding middleware on router can be as simple as some method that will apply decorator on all endpoints (with a flag if it should use decorate_view or work after validation

nikatlas:
@vitalik 

This PR is actually like option 2 (decorating manually a view function) in bulk for both routers and apis in a composable way. The middlewares in this PR are running after validation just before the actual view function would run.

Benefits: 
- Access validated payload
- Run only after all checks including CSRF, auth, throttling (_run_checks in operation)
- Possible to return apart from an HttpResponse a Schema (like views do)
- Composable per router and/or api (e.g. django-middlewares cannot be applied per router)

The option to run before the validation can be added and is actually a simpler use-case as you point out by applying the `decorate_view`.

-------

Regarding option 3 (decorate_view) this is running before the validation and all other checks (it is pretty much outside of the lifecycle of the operation).

-------

My use-case is actually to handle some common authorization errors. We had duplicated a 403 response across many views. The only caveat is that I did not find a way to define router/api common responses in a single place, so I raise an HTTPResponse from the middleware. If we add the `before validation` I will be able to extend the response models as well of the operation to include the generic 403 for the routers I want.

nikatlas:
I could also change the implementation to use the `decorate_view` as suggested for pre-validation, and if the `post_validation` flag is set, wrap the `view_func` of the operation with the middleware. 

This would require to re-construct the operations as all these happen in the constructor while the actual url/route composition happens later. 

I think this is a small pitfall from architectural perspective, ideally the operation declaration(description) and the operation runtime instance could be decoupled.

vitalik:
Hi @nikatlas 

sorry for taking so long - I think I'm more keen to not use "middleware" inside ninja to not confuse users with django middelwares - instead i added 'add_decorator' function that can support both modes (view level, and operation level)

Please take a look and see if that solves your case #1525 

full doc on all decorators - https://github.com/vitalik/django-ninja/blob/309682538514be6cc4d9e4893744ce58de8f47d5/docs/docs/guides/decorators.md

nikatlas:
> Hi @nikatlas
> 
> sorry for taking so long - I think I'm more keen to not use "middleware" inside ninja to not confuse users with django middelwares - instead i added 'add_decorator' function that can support both modes (view level, and operation level)
> 
> Please take a look and see if that solves your case #1525
> 
> full doc on all decorators - https://github.com/vitalik/django-ninja/blob/309682538514be6cc4d9e4893744ce58de8f47d5/docs/docs/guides/decorators.md

@vitalik 
Thank you for the PR.  🎉 🚀 

I did a quick test myself and reviewed the PR as well. It seems to work well and covers my use-cases. 

Please make a release of a new version that includes this feature. 

I agree with you keeping the `decorator` concept is better. Less is more ;)

nikatlas:
Closed in favor of #1525 which Implements middlewares as decorators.

PR #1242, opened 2024-07-29
add union type and subtypes check in schema model signature

fazeelghafoor:
Fixes #1153 Add support for union type field in model schema

PR #1512, opened 2025-07-17
Add union type and subtypes check in schema model signature (with tests)

dan-blanchard:
This PR is a rebased version of https://github.com/vitalik/django-ninja/pull/1242 that also:

- Adds a couple tests
- Fixes some flipped logic in the original PR where it had `if type(arg) is None:` instead of `if arg is type(None):`

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -149,81 +149,12 @@ def view_multi_model_union(request, data: MultiModelUnion):
     return data.model_dump()
 
 
-# Create models that will cause field name collisions during flattening
-class ModelA(BaseModel):
-    name: str
-
-
-class ModelB(BaseModel):
-    name: str  # Same field name as ModelA
-
-
-# Test case to trigger model field collision error
-try:
-    router_collision = Router()
-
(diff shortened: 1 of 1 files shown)


dan-blanchard:
I've rebased this since it got a little stale

vitalik:
Hi @dan-blanchard - could you add missing tests coverage on 

ninja/signature/details.py   lines - 212-215, 355-368

vitalik:
Note: #1483 can be related

ninja/testing/client.py
```diff
@@ -206,7 +206,7 @@ def __init__(self, http_response: Union[HttpResponse, StreamingHttpResponse]):
         if self.streaming:
             self.content = b"".join(http_response.streaming_content)  # type: ignore
         else:
-            self.content = http_response.content  # type: ignore[union-attr]
```
dan-blanchard, line 209: mypy was complaining about this ignore being unnecessary for me

dan-blanchard:
I rebased this on `master`, added more tests, fixed some bugs that those tests revealed, and then got coverage to 100%.

PR #1524, opened 2025-08-08
Remove deprecated csrf argument

vitalik:


ninja/openapi/docs.py
```diff
@@ -103,8 +103,9 @@ def _render_cdn_template(
 
 
 def _csrf_needed(api: "NinjaAPI") -> bool:
```
Xdynix, line 105: This function only checks for global authentication handler settings. If a project relies solely on per-route or per-view `auth=` settings, those won’t be detected. As a result, the OpenAPI docs page won’t include the CSRF token and won’t be able to interact with those endpoints. Is there a way to ensure the OpenAPI docs page always includes the CSRF token, regardless of the authentication configuration?

vitalik: @Xdynix yeah, you are right - this should go deep into each operation and if at least one has csrf requirement should return True

PR #1537, opened 2025-09-03
Promote pydantic 2.x model_dump in docs

pinguin999:
https://docs.pydantic.dev/latest/migration/#changes-to-pydanticbasemodel
dict() is now model_dump()

### DIFF
--- a/docs/docs/guides/response/django-pydantic.md
+++ b/docs/docs/guides/response/django-pydantic.md
@@ -114,7 +114,7 @@ When you process input data, you need to tell Pydantic to avoid setting undefine
 def patch(request, pk: int, payload: PatchGroupSchema):
 
     # Notice that we set exclude_unset=True
-    updated_fields = payload.dict(exclude_unset=True)
+    updated_fields = payload.model_dump(exclude_unset=True)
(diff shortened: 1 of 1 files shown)


vitalik:
Hi @pinguin999 

Well that's an open question actually 

In django ninja we try to avoid "model" word as in django universe it's mainly mean database table

Currently both .dict and model_dump are supported without any deprecation warnings - what should be at the end - i'm not sure yet


and I would like to hear some other opinions...

pinguin999:
For me dict is a python keyword. So I would try to avoid using it for other functions.

pmdevita:
It seems like we'd either have the current situation, where new users like darshan-roon are confused about why there are two methods that do the same thing, or if we drop it, we potentially have confusion about it being named `model` like vitalik said. 

I have sort of come to the idea that as users get comfortable with Ninja, they should also come to be comfortable with Pydantic, just as one would using FastAPI. This makes me less in favor of having an alias then, as it may make for confusion in teaching Pydantic.

EDIT: darshan-roon ended not being a great example because they had prior Pydantic experience

PR #1554, opened 2025-10-01
Add guide for using request headers and improve type hints

skokado:
Hi, I'm very greateful this project!

I ran into a few issues when trying to use custom headers.

First, the function argument `my_header` defined as follows didn't work no matter what I did.

```python
from ninja import Header

@api.get("")
def events(request, my_header: Header[str]):
    return {"received": my_header}
```

I got a 422 status code.

```sh
$ curl localhost:8000
{
  "detail": [
    {
      "type": "missing",
      "loc": [
        "header",
        "my_header"
      ],
      "msg": "Field required"
    }
  ]
}
```

To resolve this, I had to use an alias to separate the header name with hyphens.

```python
@api.get("")
def events(request, my_header: str = Header(alias="my-header"):
    return {"received": my_header}
```

```sh
$ curl -H 'my-header: abc' localhost:8000
{"received": "abc"}
```

I also discovered a "mypy cheat" block in `ninja.params`.
This causes `Header` to always be annotated in the editor, which causes `Header()` to issue a warning.
I'm not entirely sure why "mypy cheat" is necessary today, but it at least seems to work.

I've also added documentation based on the issues I encountered.

Personally, I think it would be better to expect similar usage for parameters like `Query` and `Path` in `ninja.params`, but I won't make any changes in this PR.


skokado:
A little more information:

I added the hyphenated alias example because I was considering integration with OpenAPI.

By default, the header name would be in snake case, which would result in a 422 response, which I thought might confuse developers.

When `alias` not set for argument `my_header`, the header name in OpenAPI also would be `my_header`

image

skokado:
@vitalik Hi, could you please give me any feedback? 🙏

PR #1562, opened 2025-10-14
Add python 3.14 to test matrix

pinguin999:


pinguin999:
I added python 3.14 to the test matrix. 
It fails, and my current projects Ci also fails with errors:

- JSON datetime fields have shorter time
`{'breakfast': '15:33:13.247'} != {'breakfast': '15:33:13.247989'}`

- Optional GET parameters stopped working: search and order can not be left out.
```
class RecipeFilters(Schema):
    tags: list[int] | None = None
    categories: list[int] | None = None
    favorited: bool = False


@router.get("/recipes", response=list[RecipesSchema])
def recipes(
    request: AuthenticatedHttpRequest,
    filters: RecipeFilters = Query(),
    search: str | None = None,
    order: list[str] = Query([]),
) -> QuerySet[Recipe]:
```

vitalik:
Hi @pinguin999 - could you add some tests for stuff that does not work for you 

Note : as it looks like only django 5.2+ suppoprts python3.14 atm

PR #1568, opened 2025-10-16
Handle empty sequence as auth

ziima:
Allow `auth` to be set to an empty sequence and act as if it was set to `None`.


vitalik:
what actually happens when you set empty list (before applying this patch)

ziima:
It produces `TypeError` on access to the view:
```
    def _run_authentication(self, request: HttpRequest) -> Optional[HttpResponse]:
            try:
                    result = callback(request)
                else:
>                   result = callback(request)
                             ^^^^^^^^^^^^^^^^^
E                   TypeError: 'list' object is not callable
```

PR #1569, opened 2025-10-16
Fix custom items_attribute in pagination

ziima:


vitalik:
Hi @ziima - can you check linting errors

ziima:
Fixed.

PR #1570, opened 2025-10-16
Fix paginate_queryset annotations

ziima:


vitalik:
what problme does this solves ?

ziima:
I was trying to include into pagination data based on request, in my case URLs. It wasn't mentioned in the docs nor in the annotations, despite the fact that `request` is always passed to the `(a)paginate_queryset` method.

PR #1574, opened 2025-10-21
Fix use of `PathEx` with `Annotated` types

jceipek:
This PR fixes https://github.com/vitalik/django-ninja/issues/1575: a problem with `PathEx` failing in conjunction with `Annotated` types. The fix will only work for Python 3.9+ because `Annotated[]` doesn't exist at runtime in earlier versions.

`_get_param_type` didn't account for [Annotated flattening](https://docs.python.org/3.11/library/typing.html#typing.Annotated:~:text=Nested%20Annotated%20types%20are%20flattened). To fix this problem, we need to change `_get_param_type` to be aware that `args` might contain more than two values.

I've modified `test_schema_path` to test a `PathEx` param with an annotated type, and I've added test cases for a new `get_path_param_ex_id` that uses `PathEx` with an annotated type.

### DIFF
--- a/ninja/signature/details.py
+++ b/ninja/signature/details.py
@@ -237,7 +237,7 @@ class ViewSignature:
                         # invalid in Python < 3.11 because star expressions
                         # were not allowed in index expressions.
                         annotation, default = Annotated[args[:-1]], args[-1]
-                    else: # pragma: no cover -- requires specific Python versions
+                    else:  # pragma: no cover -- requires specific Python versions
                         raise NotImplementedError(
                             "This definition requires Python version 3.9+"
                         )


ninja/signature/details.py
```diff
@@ -222,9 +222,12 @@ def _get_param_type(self, name: str, arg: inspect.Parameter) -> FuncParam:
...
         if get_origin(annotation) is Annotated:
             args = get_args(annotation)
-            if isinstance(args[1], Param):
+            if isinstance(args[-1], Param):
                 prev_default = default
-                annotation, default = args
+                if len(args) == 2:
+                    annotation, default = args
+                else:
+                    annotation, default = Annotated[*args[:-1]], args[-1]
```
jceipek: It looks like this runtime re-wrapping produces a syntax error in some versions of Python. I'm not sure what a good alternative would be but am open to suggestions, @vitalik

ninja/signature/details.py
```diff
@@ -222,9 +223,24 @@ def _get_param_type(self, name: str, arg: inspect.Parameter) -> FuncParam:
...
+                    if version_info >= (3, 9):
+                        # NOTE: Annotated[args[:-1]] seems to have the same runtime
+                        # behavior as Annotated[*args[:-1]], but the latter is
+                        # invalid in Python < 3.11 because star expressions
+                        # were not allowed in index expressions.
+                        annotation, default = Annotated[args[:-1]], args[-1]
+                    else:  # pragma: no cover -- requires specific Python versions
+                        raise NotImplementedError(
+                            "This definition requires Python version 3.9+"
+                        )
```
jceipek: I added tests that cover this branch, but the tests are only defined for specific Python versions, hence the need for `# pragma: no cover`.

I'm not 100% sure `NotImplementedError` is the appropriate exception here. See https://docs.python.org/3/library/exceptions.html#NotImplementedError. Let me know if there's a different exception you'd prefer.

vitalik: fine with - looking at the stats 3.8 is less then 1% in downloads - chances that some one using this are minimal (and maybe will push to upgrade python version)

vitalik:
Super
Thank you!

Issue #1564, opened 2025-10-15
[BUG] query params do not work


fraser-langton:
Query params behaviour very hard to understand, default / optional values don't work. The open api shows them correcty as optional but ninja fails the validation 

```json
{
  "detail": [
    {
      "type": "string_type",
      "loc": [
        "query",
        "q"
      ],
      "msg": "Input should be a valid string"
    }
  ]
}
```

I have attached a demo, including ones taken from the docs

```python
import sys
from datetime import date

import django
from django.conf import settings
from django.core.management import execute_from_command_line
from django.urls import path
from pydantic import BaseModel

# Configure minimal Django settings
if not settings.configured:
    settings.configure(
        DEBUG=True,
        SECRET_KEY="test-secret-key-for-demo",
        ROOT_URLCONF=__name__,
        ALLOWED_HOSTS=["*"],
        INSTALLED_APPS=["django.contrib.contenttypes", "django.contrib.auth"],
        MIDDLEWARE=["django.middleware.common.CommonMiddleware"],
        APPEND_SLASH=True,
        TEMPLATES=[
            {
                "BACKEND": "django.template.backends.django.DjangoTemplates",
                "DIRS": [],
                "APP_DIRS": True,
                "OPTIONS": {},
            }
        ],
    )
    django.setup()

# Import ninja AFTER Django is configured
from ninja import Field, NinjaAPI, Query, Schema

api = NinjaAPI()


@api.get("/example1")
def example_1(request, q: str | None = None):
    """
    doesn't work
    """
    return {"q": q}


@api.get("/example2")
def example_2(request, q: Query[str | None] = None):
    """
    doesn't work
    """
    return {"q": q}


@api.get("/example3")
def example_3(request, q: str | None = Query(None)):
    """
    doesn't work
    """
    return {"q": q}


class QueryParams1(Schema):
    q: str | None = None


@api.get("/example4")
def example_4(filters: Query[QueryParams1]):
    """
    doesn't work
    """
    return {"q": filters.q}


@api.get("/example5")
def example_5(request, filters: Query[QueryParams1]):
    """
    works (requires request?)
    """
    return {"q": filters.q}


class QueryParams2(BaseModel):
    q: str | None = None


@api.get("/example5")
def example_5(request, filters: Query[QueryParams2]):
    """
    pydnatic doesn't work
    """
    return {"q": filters.q}


[...]

PR #1582, opened 2025-10-26
Fix issues with nested default dicts

### LINKED ISSUE #1581: [BUG] Default attributes within BaseModel not used for ninja.Query
Default attributes are not used when query params are parsed for pydantic `BaseModel`. I faced this issue upgrading a larger codebase from an old django-ninja version to a more recent.

```python
from pydantic import BaseModel

class Filter(BaseModel):
    foo: int = 1

@api.get("/test")
def test(request, filters: Filter = Query(...)):
    ...
```

Here's a one liner showing the issue (showing `BaseModel` vs `Schema`), ready to paste (requiring only `uv`):

```
uv run --with "git+https://github.com/vitalik/django-ninja@master" python <<'PY'
import os
from django.conf import settings

os.environ["NINJA_SKIP_REGISTRY"] = "yes"

settings.configure(
  DEBUG=True,
  SECRET_KEY="demo",
  ROOT_URLCONF=__name__,
  ALLOWED_HOSTS=["*"],
  INSTALLED_APPS=["django.contrib.auth", "django.contrib.contenttypes"],
  MIDDLEWARE=[],
)

import django
django.setup()

from ninja import NinjaAPI, Query, Schema
from ninja.testing.client import TestClient
from pydantic import BaseModel

class FilterBaseModel(BaseModel):
  start: int
  optional: int = 42

class FilterSchema(Schema):
  start: int
  optional: int = 42

api = NinjaAPI(urls_namespace="demo-cli")

@api.get("/basemodel")
def basemodel_view(request, filters: FilterBaseModel = Query(...)):
  return {"filters": filters.dict()}

@api.get("/schema")
def schema_view(request, filters: FilterSchema = Query(...)):
  return {"filters": filters.dict()}

urlpatterns = api.urls

client = TestClient(api)
params = "start=1"

resp_base = client.get(f"/basemodel?{params}")
resp_schema = client.get(f"/schema?{params}")

print("BaseModel:", resp_base.status_code, resp_base.json())
print("Schema:", resp_schema.status_code, resp_schema.json())
PY
```

Output using `master` version:
```
BaseModel: 422 {'detail': [{'type': 'int_type', 'loc': ['query', 'optional'], 'msg': 'Input should be a valid integer'}]}
Schema: 200 {'filters': {'start': 1, 'optional': 42}}
```

Change version from `@master` to `@v0.22.2` to see last functioning version

I saw #1564 but not sure if this is the same thing. I tried with all the versions mentioned there they all have this issue still.

**Versions:**
 - Python version: 3.13.1
 - Django version: 5.2.7
 - Django-Ninja version: All version from `v1.0` to `master`
 - Pydantic version: 2.12.3

HeyHugo:
Resolves #1581 by handling nested default dicts

### DIFF
--- a/ninja/params/models.py
+++ b/ninja/params/models.py
@@ -1,5 +1,4 @@
 from abc import ABC, abstractmethod
-from collections import defaultdict
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -39,10 +38,6 @@ TModel = TypeVar("TModel", bound="ParamModel")
 TModels = List[TModel]
 
 
-def NestedDict() -> DictStrAny:
-    return defaultdict(NestedDict)
-
-
 class ParamModel(BaseModel, ABC):
     __ninja_param_source__ = None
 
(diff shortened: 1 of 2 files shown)


HeyHugo:
I think #1564 is the same issue and would also be resolved by this fix

HeyHugo:
Resolved CI failures

vitalik:
Hi @HeyHugo 

Thank you

I'm wondering - maybe it makes more sense to get rid of NestedDict at all (or subclass dict so that pydantic does not care)

the concern is that at the end there is just extra work to convert from nesteddict to regular dict at runtime that might give some timing penalties

HeyHugo:
@vitalik I managed to drop NestedDict (defaultdicts) but keep same behaviour using dict setdefault instead.

PR #1586, opened 2025-11-05
chore: Remove a character in CSRF code example

makisukurisu:
Hi! I've noticed this typo in the CSRF code example.

On a side note, should there be some migration guide for users that prior to 1.5.0 have used `csrf` in their app config? (From what I see - it's enough to remove the parameter, and it'll _mostly_ work the same, except for anonymous requests where django_auth is required. Previously it returned code 401, and now it says that there's CSRF error, and returns error code 403)

I see that it's mentioned that this parameter is removed with 1.5.0 release (in beta release noes), however it doesn't clearly state that without removing it first, the app simply won't run.

### DIFF
--- a/docs/docs/reference/csrf.md
+++ b/docs/docs/reference/csrf.md
@@ -21,7 +21,7 @@ By default, **Django Ninja** has CSRF protection turned **OFF** for all operatio
 from ninja import NinjaAPI
 from ninja.security import APIKeyCookie
 
-class CookieAuth(APIKeyCookie):å
+class CookieAuth(APIKeyCookie):
     def authenticate(self, request, key):
         return key == "test"
 


makisukurisu:
UPD:
image

So, I'd suggest adding some note that decorators also have to be applied in a view mode, it would seem.
However it doesn't work this way.

Is there any way to use Django authentication with CSRF on?

makisukurisu:
Just to clarify, here's what my code looks like:

```python

from django.views.decorators.csrf import csrf_exempt, ensure_csrf_cookie

from .router import admin_router

@admin_router.post(
    "/department/fetch",
)
@ensure_csrf_cookie
@csrf_exempt
def fetch_departments(request: "HttpRequest") -> bool:
    # There's some actual logic
    return True
```

In `router.py`:
```python
from ninja import Router
from ninja.security import django_auth


admin_router = Router(
    auth=django_auth,
    tags=[
        "Admin",
    ],
)
```

Then this router is added in the app:
```python
from ninja import NinjaAPI

from ...api.decorators import request_id_decorator
from .admin.router import admin_router


app = NinjaAPI()

app.add_decorator(
    request_id_decorator,
    mode="view",
)


app.add_router("/admin/", admin_router)
```

The result is that Django tries to add a cookie to the boolean value (result of the function)
I've tried adding the decorators in view mode to the request, however, it simply yielded me with 403 error and CSRF verification error.

makisukurisu:
After some more tinkering - I need to decorate endpoints as CSRF exempt (not the router), but can apply `ensure_csrf_cookie` to the router in `view` mode.

vitalik:
character fixed
can you move the  issue/question to [Issues](https://github.com/vitalik/django-ninja/issues) ?

PR #1608, opened 2025-11-30
Fixed pagination override in case of RouterPaginated

Lokker29:
This PR fixes errors that happen when RouterPaginated is used. The system crashes when a user overrides the pagination. 

```python
router = RouterPaginated()

@router.get('/items', response=List[int])
@paginate                      #  ninja.errors.ConfigError: Duplicated name: 'limit' in params: 'ninja_pagination' & 'ninja_pagination'
def items(self):
    return ITEMS

# More relevant case where a developer needs to override default pagination
@router.get('/items', response=List[int])
@paginate(PageNumberPagination)     # AttributeError: 'NoneType' object has no attribute 'annotation'  
def items(self):
    return ITEMS
```

The first view:
image

The second view:
image

Solution: Every view can be wrapped into `_inject_pagination` at most once. When RouterPaginated wraps the paginated view, it will skip silently.

### DIFF
--- a/ninja/pagination.py
+++ b/ninja/pagination.py
@@ -219,6 +219,9 @@ def _inject_pagination(
     paginator_class: Type[Union[PaginationBase, AsyncPaginationBase]],
     **paginator_params: Any,
 ) -> Callable[..., Any]:
+    if getattr(func, "_is_paginated", False):
+        return func
+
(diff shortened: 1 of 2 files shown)


vitalik:
Thank you

PR #1614, opened 2025-12-04
Fix optional query params with None default (issue #1607)

### LINKED ISSUE #1607: [BUG] Optional query param doesn't work unless I use Schema
**Describe the bug**

This endpoint below returns this error when I don't pass a query param, which I expect to be optional. It only works when I create a schema class and pass that to the Query class.

```python
@tenants_super_router.get("/")
async def get_farms(request: ASGIRequest, tenant_schema: str = None):
    return tenant_schema
```

```json
{
	"detail": [
		{
			"type": "string_type",
			"loc": [
				"query",
				"tenant_schema"
			],
			"msg": "Input should be a valid string"
		}
	]
}
```

 - Python version: 3.13
 - Django version: 5.2.7
 - Django-Ninja version: 1.4.3
 - Pydantic version: 2.12.0

Note you can quickly get this by runninng in `./manage.py shell` this line:
```
import django; import pydantic; import ninja; django.__version__; ninja.__version__; pydantic.__version__
```

Vincent-Ngobeh:
## Summary
Fix for issue #1607 - Optional query parameters with `= None` default now work correctly.

## Problem
When defining a query parameter like `param: str = None`, Pydantic v2 validation failed because the annotation was `str` but the default was `None`.

## Solution
- Added `_is_optional_type()` helper method to detect if an annotation is already Optional
- Automatically wrap annotation in `Optional[T]` when default is `None` but type is not already Optional

## Test plan
- [x] Added test case for `str` type with `None` default
- [x] All existing tests pass (667 passed)
- [x] Linting passes (`ruff check`)
- [x] Type checking passes (`mypy`)

Fixes #1607

### DIFF
--- a/ninja/signature/details.py
+++ b/ninja/signature/details.py
@@ -25,6 +25,7 @@ from ninja.params.models import (
     _MultiPartBody,
 )
 from ninja.signature.utils import get_path_param_names, get_typed_signature
+from ninja.utils import is_optional_type
 
 __all__ = [
     "ViewSignature",
@@ -294,6 +295,11 @@ class ViewSignature:
                 param_source = Query(...)
             else:
                 param_source = Query(default)
+
+        # If default is None but annotation is not Optional,
+        # wrap it in Optional to allow None values in Pydantic v2
+        if default is None and not is_optional_type(annotation):
+            annotation = Optional[annotation]
 
         return FuncParam(
             name, param_source.alias or name, param_source, annotation, is_collection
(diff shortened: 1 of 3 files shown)


vitalik:
Hi @Vincent-Ngobeh 

can you use already defined function to check if type is optional https://github.com/vitalik/django-ninja/blob/e5806caa8318c258bf2906285088c463f5edd61e/ninja/utils.py#L53

Vincent-Ngobeh:
Yes, no problem. Thanks for the feedback.

Vincent-Ngobeh:
I just made the changes. Let me know if they are fine.

vitalik:
thank you

Issue #1615, opened 2025-12-04
It should be possible to create two APIs with same urls_namespace


anentropic:
Django url namespaces are hierarchical rather than a flat global namespace.

Currently there is some validation to prevent creating two Ninja APIs with the same `urls_namespace`:

```
  File "/home/circleci/project/.venv/lib/python3.11/site-packages/ninja/main.py", line 436, in urls
    self._validate()
  File "/home/circleci/project/.venv/lib/python3.11/site-packages/ninja/main.py", line 577, in _validate
    raise ConfigError(msg.strip())
ninja.errors.ConfigError: Looks like you created multiple NinjaAPIs or TestClients
To let ninja distinguish them you need to set either unique version or urls_namespace
 - NinjaAPI(..., version='2.0.0')
 - NinjaAPI(..., urls_namespace='otherapi')
Already registered: ['reports']
```

But this seems like a mis-feature, because these APIs may be loaded into Django urlconf under different parent namespaces.

Relates to https://github.com/vitalik/django-ninja/issues/1593

https://github.com/vitalik/django-ninja/blob/e5806caa8318c258bf2906285088c463f5edd61e/ninja/main.py#L562-L578

https://github.com/vitalik/django-ninja/blob/e5806caa8318c258bf2906285088c463f5edd61e/ninja/main.py#L428-L442

The registry itself doesn't seem to serve any other purpose except prevent people from creating APIs with same namespace (?) ... but I don't think this should be disallowed in the first place?

anentropic:
For example, this makes it awkward to use same/similar API defs under different app names

(In my case I am also using django-ninja-extra class-based API Controller to have code reuse between similar views in different apps)

anentropic:
as a compromise it could be nice if `NINJA_SKIP_REGISTRY` flag was read from Django settings instead of raw env var

vitalik:
@anentropic you can just use Router instead of NinjaAPI ?

```
api1 = Router()
...

api2 = Router()


api_root = NinjaAPI()
api_root.add_router(api1)
api_root.add_router(api2)
```

or am my missing something ?

anentropic:
I see

I'm not sure that is possible with django-ninja-extra

They have:

```python
from ninja_extra import NinjaExtraAPI

api = NinjaExtraAPI()

api.register_controllers(MyAPIController, MyAPIController2, ...)
```

[...]

PR #1616, opened 2025-12-04
Allow to disable unnecessary API registry checks via Django settings

anentropic:
A minimal change to achieve what I needed in https://github.com/vitalik/django-ninja/issues/1615 without changing any core functionality

I have tried this in my project and seems to work fine, I am able to do:

```python
    path("app-a/", include("apps.a.urls", namespace="a")),
    path("app-b/", include("apps.b.urls", namespace="b")),
```
where `apps.a.urls` and `apps.b.urls` both look like:
```python
api = NinjaAPI(urls_namespace="reports")
...
urlpatterns = [
    path("other", other.view, name="other"),
    path("", api.urls),
]
```
and then I can reverse url names like `a:reports:api-view`, `b:reports:api-view`, or non-ninja views like `a:other`

(I feel like the registry could probably just be removed entirely, AFAICT it's not doing anything useful)

### DIFF
--- a/ninja/main.py
+++ b/ninja/main.py
@@ -1,4 +1,3 @@
-import os
 from typing import (
     TYPE_CHECKING,
     Any,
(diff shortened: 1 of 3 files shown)


brianhelba:
I think this PR is obsoleted by #1622, so this can be closed. `NINJA_SKIP_REGISTRY` no longer exists.

PR #1618, opened 2025-12-05
Added mode to decorate_view

Lokker29:
Hello. 

I added mode (operation, view) to the `@decorate_view`. 

```python
@api.get("/cached")
@decorate_view(cache_page(60 * 15), mode="view") 
def cached_endpoint(request):
    return {"data": "This response is cached"}

def require_feature_flag(flag_name):  # example from the doc about `router.add_decorator(..., mode="operation")`
    def decorator(func):
        @wraps(func)
        def wrapper(request, *args, **kwargs):
            if not request.user.has_feature(flag_name):
                return {"error": f"Feature {flag_name} not enabled"}
            return func(request, *args, **kwargs)
        return wrapper
    return decorator

@api.get("/new-feature")
@decorate_view(require_feature_flag("new_api"), mode="operation") 
def new_feature(request):
    return {"feature": "enabled"}
```

It allows to add decorators to individual endpoints that can do some code after validation. Especially when a developer needs to perform some validation over request, validated path parameters and validated body parameters together.
It's common enough that there is a router with a lot of endpoints, but only a few of them require an "operation" decorator so that they couldn't use `router.add_decorator`.

As well, it looks more consistent with `api.add_decorator` and `router.add_decorator` that make the same sense as `@decorate_view`

P.S. I would like to mark usage `add_decorator` and `decorate_view` without explicit `mode` parameter deprecated, and made them mandatory from 1.6.0 or any other following version.

### DIFF
--- a/docs/docs/guides/decorators.md
+++ b/docs/docs/guides/decorators.md
@@ -7,11 +7,12 @@ Django Ninja provides flexible decorator support to wrap your API operations wit
 Django Ninja supports two modes for applying decorators:
 
 
-### OPERATION Mode (Default)
+### OPERATION Mode
 - Applied **after** Django Ninja's validation
 - Wraps the operation function with validated data
 - Has access to parsed and validated parameters
 - Useful for: business logic, logging with validated data, post-validation checks
+- Default mode in `api.add_decorator` and `router.add_decorator`
 
 ### VIEW Mode
 - Applied **before** Django Ninja's validation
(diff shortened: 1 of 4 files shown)


vitalik:
Just trying to understand the problem here


the current implemenation is the following:


1) this is basically an "operation" decorator - means it will get validated arguments 
```python
@api.get('/some')
@mydecorator
def endpoint(request):
      return ...

```

2) this is a "view" decorator - basically a helper that let's you do some decoration for the django view function even before the pydantic validation kicks in:

```python
@api.get('/some')
@decorate_view(mydecorator)
def endpoint(request):
      return ...

```


so tehnically there is no need for "operation" decorator on per-function basis

Lokker29:
Yes, it makes sense. I've forgotten about the "no @decorate_view" way. I close the request.

PR #1639, opened 2026-01-02
Add path_params to context in ParamModel resolve

nfd:
When validating, it's convenient to get `path_params` as part of context. For example, `PATCH /items/<item id>/` might want to validate that `item id`, if supplied, references a record of a specific type. Obviously this could be done after validation, but it feels like it should be part of the validation process.

This PR adds `path_params` to the context for model validation.


Lokker29:
I endorse the idea of the request. I need this feature time to time too

Issue #1655, opened 2026-01-22
[BUG] pydantics MISSING sentinel is not handled


Ksauder:
**Describe the bug**

I'm experimenting with the use of the MISSING sentinel added in pydantic 2.12. I ran into an issue during serialization where a default value could not be created for the sentinel which is triggered by the overriding of [`default_schema`](https://github.com/vitalik/django-ninja/blob/3c8f7d9d99de59f09867c508d85459e7f6e47f9a/ninja/schema.py#L187) since the MISSING sentinel is not treated as a no-op.

I'm not sure the custom generator or method are needed, at least not for the `allOf` piece since I believe that changed in openapi schema v3.1, but I didn't trace all of that down. It may still be needed to handle how default factories and callables are handled.

**Versions (please complete the following information):**
 - Python version: 3.12.12
 - Django version: 5.28
 - Django-Ninja version: 1.5.0
 - Pydantic version: 2.12.4

PR #1656, opened 2026-01-22
enable MISSING and NoDefault sentinel

Ksauder:
#1655

I've been struggling with model schemas having `null` as an option for fields that I simply want to omit from the response. If the keys don't exist in the response, then there is no data. The current implementation does not allow changing how `create_schema` handles nullable fields and it always assigns a `type | None = None`. This causes all of my generated api clients to type hint incorrectly which is really annoying. `exclude_unset` also does not work for modelschemas since it's technically set since the model instance is interrogated via attributes and returns an explicit `None` for nullable fields and it does not help generate an accurate json schema.

The main feature of this PR is allowing modification to the type and value used when a field is "nullable". This allows [`MISSING`](https://docs.pydantic.dev/latest/concepts/experimental/#missing-sentinel) to be used, which will create the data schema I'm looking for, while also supporting other type/value combos.

```python
from typing import Literal

from django.db import models
from ninja.orm import create_schema
from ninja import NinjaAPI
from pydantic.experimental.missing_sentinel import MISSING

from devtools import debug


class Profile(models.Model):
    name = models.CharField(max_length=10)
    nullable = models.CharField(max_length=10, blank=True, null=True)

    class Meta:
        app_label = "tests"


ProfileModelSchema1 = create_schema(
    Profile,
    name="ProfileModelSchema1",
    fields=["name", "nullable"],
    nullable_type=MISSING,
    nullable_value=MISSING,
)


ProfileModelSchema2 = create_schema(
    Profile,
    name="ProfileModelSchema2",
    fields=["name", "nullable"],
    nullable_type=Literal["MyNullValue"],
    nullable_value="MyNullValue",
)

api = NinjaAPI()


@api.get("/profile1", response=ProfileModelSchema1)
def get_profile1(request):
    return Profile(name="Name")


@api.get("/profile2", response=ProfileModelSchema2)
def get_profile2(request):
    return Profile(name="Name")

api.get_openapi_json() == {
        ...
        'components': {
            'schemas': {
                'ProfileModelSchema1': {
                    'properties': {
                        'name': {
                            'maxLength': 10,
                            'title': 'Name',
                            'type': 'string',
                        },
                        'nullable': {
                            'maxLength': 10,
                            'title': 'Nullable',
                            'type': 'string',
                        },
                    },
                    'required': ['name'],
                    'title': 'ProfileModelSchema1',
                    'type': 'object',
                },
                'ProfileModelSchema2': {
                    'properties': {
                        'name': {
                            'maxLength': 10,
                            'title': 'Name',
                            'type': 'string',
                        },
                        'nullable': {
                            'anyOf': [
                                {
                                    'maxLength': 10,
                                    'type': 'string',
                                },
                                {
                                    'const': 'MyNullValue',
                                    'type': 'string',
                                },
                            ],
                            'default': 'MyNullValue',
                            'title': 'Nullable',
                        },
                    },
                    'required': ['name'],
                    'title': 'ProfileModelSchema2',
                    'type': 'object',
                },
            },
        },
        ...
    }
```

### DIFF
--- a/.github/workflows/test_full.yml
+++ b/.github/workflows/test_full.yml
@@ -13,6 +13,7 @@ jobs:
       matrix:
         python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12', '3.13', '3.14']
         django-version: ['<3.2', '<3.3', '<4.2', '<4.3', '<5.1', '<5.2', '<5.3', '<6.1']
+        pydantic-version: ['']
         exclude:
           - python-version: '3.7'
             django-version: '<5.1'
@@ -46,7 +47,18 @@ jobs:
             django-version: '<5.1'
           - python-version: '3.14'
             django-version: '<5.2'
-    
+
+        include:
+          - python-version: '3.7'
+            django-version: '<6.1'
+            pydantic-version: '<2.11'
+          - python-version: '3.13'
+            django-version: '<6.1'
+            pydantic-version: '<2.11'
+          - python-version: '3.14'
+            django-version: '<6.1'
+            pydantic-version: '<2.11'
+
     steps:
       - uses: actions/checkout@v6
       - name: Set up Python
--- a/ninja/conf.py
+++ b/ninja/conf.py
@@ -1,9 +1,8 @@
 from math import inf
-from typing import Any, Dict, Optional, Set, Tuple, Union
+from typing import Any, Dict, Optional, Set, Tuple
 
 from django.conf import settings as django_settings
 from pydantic import BaseModel, ConfigDict, Field
-from pydantic.experimental.missing_sentinel import MISSING
 
 
 class Settings(BaseModel):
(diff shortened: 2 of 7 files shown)


vitalik:
hi @Ksauder - would you be able to add some matix to test with `pydantic < 2.11` - would not run against all python versions but just one latest

Ksauder:
Sure I can look into that. I'm still working out actually using this in practice. I'm debugging some issues when a schema with  MISSING is actually used as a return schema from an operation.

Ksauder:
@vitalik Okay, I think this is a pretty clean implementation. Added two args to create_schema and relevant child functions `nullable_value` and `nullable_type` with default from conf, NULLABLE_FIELD_*. Had to do this to get MISSING down to the model schema level.
And added MISSING to the check in default_schema so we don't try to encode it.

Ksauder:
> hi @Ksauder - would you be able to add some matix to test with `pydantic < 2.11` - would not run against all python versions but just one latest

@vitalik Actually, I'm not sure how you would like this implemented. I'm not familiar with the github CICD, and I assume you want it there. How exactly would you like this done?

Ksauder:
@vitalik Hi, I just rebased, cleaned up, and got the test matrix done. Any interest in this or anything I can do to move it forward?

PR #1667, opened 2026-02-11
Add default exception handler for `Throttled` to set `Retry-After` HTTP header

### LINKED ISSUE #1666: [BUG] Throttle classes don't set `Retry-After` HTTP header
**Describe the bug**

Throttle classes do not set the [`Retry-After` HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Retry-After) to either indicate the number of seconds before a request can be retried, or the timestamp after which the request can be retried.

 - Python version: 3.14
 - Django version: 6.0.2
 - Django-Ninja version: 1.5.3
 - Pydantic version: 2.12.5

**Possible solution**

One thing we could do is add a default exception handler for the `Throttled` exceptions, which add the number of seconds to retry after to a `Retry-After` header.

This is essentially what Django Rest Framework does:

https://github.com/encode/django-rest-framework/blob/cf38c94ec1376d2b5c875d227cefc2780aa148bd/rest_framework/views.py#L91-L92

This would work well as it appears that the `_check_throttles` method of the `Operation` class already appears to call the API instance's exception handler for the `Throttled` exception:

https://github.com/vitalik/django-ninja/blob/b17dcec55efa094092a965c72511cbe2d5631884/ninja/operation.py#L257-L264

cpolish:
Closes #1666.

This PR adds a default exception handler for the `Throttled` error, meaning that if the exception has a `wait` time defined (in seconds), this will be set to the HTTP `Retry-After` header.

## Testing

I noticed that the `NinjaResponse` object returned by the `NinjaClientBase` testing methods doesn't seem to expose a way to check the responses headers, hence I wasn't sure of the best way to implement tests (and thus why this PR is marked as a draft). It would be great to get some advice on this.


vitalik:
I pushed two commits to the branch:
- merged current `master` (resolved a small import conflict in `ninja/operation.py`)
- added tests covering the sync and async clients, the `math.ceil` rounding of fractional waits, and the no-`wait` case (header must be absent). Also removed the `# type: ignore` on the `Throttled(wait=duration)` call — after your `math.ceil` change the durations are ints, and mypy runs with `warn_unused_ignores` so the stale ignore would fail CI.

vitalik:
Created simpler alternative for now

cpolish:
Thank you so much for taking a look @vitalik! Appreciate your help with the tests + making the simpler fix in #1765 :).

PR #1680, opened 2026-02-17
Fix `TestClient` passing broken `request` objects to endpoints

brianhelba:
Endpoints called via `TestClient` received a `request` where any `HttpRequest` method not explicitly stubbed returned a `Mock` object instead of a real value.

This was causing all the following bugs, which are now resolved:

* `request.build_absolute_uri()` was replaced with a stub that ignored the request path entirely, returning `"http://testlocation/"` for every endpoint. Furthermore, this produced incorrect pagination URLs.
* `request.get_host()` returned a `Mock`. It now returns the value `"testserver"`, which is specifically added to `ALLOWED_HOSTS` by the Django test framework. The use of `testserver` as a host name now allows the `HttpRequest.build_absolute_uri()` to work properly without additional mocking.
* `request.get_full_path()` returned a `Mock` instead of the path with the full query string.
* `request.is_secure()` returned a (truthy) `Mock` instead of False.
* `request.body` returned a `str`, but `HttpRequest.body` returns `bytes`.
* `request.META` was missing required WSGI keys (`SERVER_NAME`, `SERVER_PORT`, `QUERY_STRING`, `PATH_INFO`), so several parts of Django would raise a `KeyError` when the `request` was used.

### DIFF
--- a/ninja/testing/client.py
+++ b/ninja/testing/client.py
@@ -146,8 +146,11 @@ class NinjaClientBase:
             for k, v in data.items():
                 request.POST[k] = v
 
+        query_string = ""
         query_params = request_params.pop("query_params", None)
         if "?" in path:
+            # Store "query_string" verbatim here, for assignment to "request.META.QUERY_STRING",
+            # to ensure empty parameter syntax is preserved
             path, query_string = path.split("?", maxsplit=1)
             request.GET = QueryDict(query_string)
         elif query_params is not None:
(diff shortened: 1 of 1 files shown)


brianhelba:
@vitalik Please take a look. I believe this is a significant bugfix to Django Ninja's `TestClient`. It doesn't change any external APIs or behavior, except to fix broken methods to return usable or more correct data in a test context.

brianhelba:
I've resolved some Mypy errors that were pre-existing, but which the change from `Mock` -> `HttpRequest` had revealed. I also responded to Copilot's feedback, which helped me to fix some additional pre-existing bugs (present before this PR).

@vitalik Could you please re-trigger CI and review this?

PR #1707, opened 2026-04-02
Use a UserWarning instead of print()

gvangool:
I noticed while updating our internal API to set an explicit `operation_id`, that the error of a duplicate `operation_id` is a regular print-statement and not a warning (so we can monitor easily for errors).

### DIFF
--- a/ninja/openapi/schema.py
+++ b/ninja/openapi/schema.py
@@ -1,5 +1,7 @@
+import inspect
 import itertools
 import re
+import warnings
 from http.client import responses
 from typing import TYPE_CHECKING, Any, Dict, Generator, List, Optional, Set, Tuple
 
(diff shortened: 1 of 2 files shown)


mbaragiola (commented):
**Blocker**

- `ninja/openapi/schema.py:121`: duplicate operation IDs can now make OpenAPI generation fail when a view’s source is unavailable. `inspect.getsourcelines(operation.view_func)` raises `OSError` for dynamically generated functions (for example, functions created with `exec`/`eval`), whereas the previous code only printed a warning. Please use a safe filename/line fallback or `warnings.warn(..., stacklevel=...)`, and cover a source-less callable in the regression test.

**Positives**

- Moving this diagnostic to Python’s warnings system lets callers filter or promote it and preserves the useful operation-ID context.
- The existing source-backed case is covered.

**Verification**

I reviewed the changed warning path and reproduced `inspect.getsourcelines()` raising `OSError` for a dynamically created function. GitHub’s full test matrix, coverage, and codestyle checks are green for this head; I did not run the project suite locally.

gvangool:
> **Blocker**
> 
> * `ninja/openapi/schema.py:121`: duplicate operation IDs can now make OpenAPI generation fail when a view’s source is unavailable. `inspect.getsourcelines(operation.view_func)` raises `OSError` for dynamically generated functions (for example, functions created with `exec`/`eval`), whereas the previous code only printed a warning. Please use a safe filename/line fallback or `warnings.warn(..., stacklevel=...)`, and cover a source-less callable in the regression test.

@mbaragiola generating new function with `exec` and `eval` will probably fail in other places too. E.g. in `ViewSignature._validate_view_path_params` https://github.com/vitalik/django-ninja/blob/134869b74b6cba214284faa9f13d54b7247362c0/ninja/signature/details.py#L102-L124

mbaragiola:
Thanks for the pointer to `ViewSignature._validate_view_path_params`.

Agreed that fully dynamic `exec`/`eval` views are already a rough edge elsewhere. The narrower concern on this PR is still a behavior change: the previous `print(...)` path never failed OpenAPI generation, while `warnings.warn_explicit(..., filename=inspect.getfile(...), lineno=inspect.getsourcelines(...)[1])` turns any source-less *or* inspect-hostile callable into a hard error during schema build.

Even without defending exotic view factories, a small `try/except (OSError, TypeError)` fallback (module name only, or plain `warnings.warn(..., stacklevel=...)`) would keep the UserWarning migration from introducing a new failure mode. Happy to drop it if maintainers prefer matching the stricter inspect usage elsewhere.

Issue #1652, opened 2026-01-12
[BUG] Python 3.14 annotations for "custom_fields" in ModelSchemaMetaclass not compatible


simonpercivall:
**Description**
Let's say I have a `ModelSchema`:
```python
class UserSchema(ModelSchema):
    password: SecretStr

    class Meta:
        model = User
        fields = ["username", "password"]
```

What `django-ninja` does, in `ModelSchemaMetaclass.__new__` is:
```python
custom_fields = []
annotations = namespace.get("__annotations__", {})
for attr_name, type in annotations.items():
    ...
    custom_fields.append((attr_name, type, default))
```

However, in Python 3.14, `namespace` will not have `__annotations__`, it will have an `__annotate_func`.

So in Python 3.14, `custom_fields == []` after that loop.

I'm unsure what the "best practice" is for this, what we're _supposed_ to do. But I've seen an least one other library doing:
```python
ann = annotationlib.get_annotate_from_class_namespace(namespace)
annotations = annotationlib.call_annotate_function(
    ann, format=annotationlib.Format.VALUE
) if ann else {}
```

_cf_.: [stdlib:annotationlib](https://docs.python.org/3/library/annotationlib.html)

**Versions:**
 - Python version: 3.14
 - Django version: 6.0
 - Django-Ninja version: 1.5.3
 - Pydantic version: 2.12.5

petrprikryl:
Maybe related with https://github.com/vitalik/django-ninja/issues/1580

Issue #1424, opened 2025-03-14
[BUG] Endpoint with query schema crashes when paginated and `from __future__ import annotations` import included


jjnesbitt:
Apologies for the lengthy title.

**Describe the bug**
I get the following error on application start:
```
NameError: name 'MySchema' is not defined
```

If the following conditions are met:

1. An endpoint function is defined with the `paginate` decorator
2. There is an argument to the function that is a user defined query param schema (`Query[MySchema]`)
3. The `from __future__ import annotations` import is present in the file

If any of these three things are excluded, the bug does not occur.

**Versions (please complete the following information):**
 - Python versions tested: 3.11, 3.12, 3.13
 - Django version: 5.1.6
 - Django-Ninja version: 1.3.0
 - Pydantic version: 2.10.6

Here is an MRE:
```
from __future__ import annotations

from ninja import NinjaAPI, Query, Schema
from ninja.pagination import PageNumberPagination, paginate


class MySchema(Schema):
    foo: str


api = NinjaAPI()


@api.get("/endpoint_path", response=list[int])
@paginate(PageNumberPagination)
def atpath(
    request,
    params: Query[MySchema],
):
    return []

```

jjnesbitt:
This seems similar to https://github.com/vitalik/django-ninja/issues/1357, but I don't think these are duplicate issues.

vitalik:
Hi @jjnesbitt 

there is some dark magic with evaulate forward refs in pydantic that I do not fully understand - but for now you can try this branch #1425 ([1424-evaluate-forward-ref](https://github.com/vitalik/django-ninja/tree/1424-evaluate-forward-ref))

jjnesbitt:
Hi @vitalik, has this been fixed via #1425?

benf710:
I'm still able to reproduce this error with Django Ninja 1.4.1. I was able to work around this issue by overriding globals using a custom `@paginate` decorator:
```python
def repair_wrapped_globalns(
    f: Callable[..., Any],
    globalns: dict[str, Any] | None = None,
    module: str | None = None,
):
    """Use this to generate a new function with an updated global namespace.

    Credit: https://stackoverflow.com/a/49077211
    """

    if globalns is None:
        globalns = f.__globals__

    g = types.FunctionType(
        f.__code__,
        globalns,
        name=f.__name__,
        argdefs=f.__defaults__,
        closure=f.__closure__,
    )

    g = functools.update_wrapper(g, f)

    if module is not None:
[...]

Issue #1357, opened 2024-12-10
[BUG] Annotating the `request` of a paginated route crashes


Jonesus:
## Describe the bug
 When annotating the `request` parameter of a route (in order to type `request.auth`) that has been decorated with `@paginate` the django server crashes with the following exception:

```
  File "/app/.venv/lib/python3.10/site-packages/ninja/router.py", line 268, in decorator
    self.add_api_operation(
  File "/app/.venv/lib/python3.10/site-packages/ninja/router.py", line 319, in add_api_operation
    path_view.add_operation(
  File "/app/.venv/lib/python3.10/site-packages/ninja/operation.py", line 426, in add_operation
    operation = OperationClass(
  File "/app/.venv/lib/python3.10/site-packages/ninja/operation.py", line 82, in __init__
    self.signature = ViewSignature(self.path, self.view_func)
  File "/app/.venv/lib/python3.10/site-packages/ninja/signature/details.py", line 48, in __init__
    self.signature = get_typed_signature(self.view_func)
  File "/app/.venv/lib/python3.10/site-packages/ninja/signature/utils.py", line 38, in get_typed_signature
    typed_params = [
  File "/app/.venv/lib/python3.10/site-packages/ninja/signature/utils.py", line 43, in <listcomp>
    annotation=get_typed_annotation(param, globalns),
  File "/app/.venv/lib/python3.10/site-packages/ninja/signature/utils.py", line 54, in get_typed_annotation
    annotation = make_forwardref(annotation, globalns)
  File "/app/.venv/lib/python3.10/site-packages/ninja/signature/utils.py", line 60, in make_forwardref
    return evaluate_forwardref(forward_ref, globalns, globalns)
  File "/app/.venv/lib/python3.10/site-packages/ninja/signature/utils.py", line 20, in evaluate_forwardref
    return cast(Any, type_)._evaluate(globalns, localns, recursive_guard=set())
  File "/home/app/.local/share/uv/python/cpython-3.10.8-linux-x86_64-gnu/lib/python3.10/typing.py", line 694, in _evaluate
    eval(self.__forward_code__, globalns, localns),
  File "<string>", line 1, in <module>
NameError: name 'AuthenticatedNinjaRequest' is not defined
```

This can be reproduced with the following minimal route:

```py
from ninja.pagination import paginate
from django.http import HttpRequest

class AuthenticatedNinjaRequest(HttpRequest):
    auth: User

@api.get('/users', response=List[UserSchema])
@paginate
[...]

PR #1711, opened 2026-04-10
annotation fixes - python 3.14 compatibility

joergbaier:
These changes allow me to enable the ruff rules TC001/TC002/TC003/UP037 + py314 for projects that use django-ninja.

---

Python 3.14 makes deferred annotation evaluation the default via PEP 649 / PEP 749.

Django Ninja already has a separate compatibility fix for the `ModelSchemaMetaclass` class-namespace case https://github.com/vitalik/django-ninja/pull/1688, but there were still a couple of `inspect.signature()` paths that could force annotation handling too early on Python 3.14:

- `get_typed_signature()` when wrapped view functions are inspected
- `get_args_names()` / `has_kwargs()` when schema resolver methods are inspected during class creation

This PR makes those Python 3.14 signature reads explicit about the annotation format:

- `get_typed_signature()` uses `annotation_format=annotationlib.Format.FORWARDREF`
- `get_args_names()` and `has_kwargs()` use `annotation_format=annotationlib.Format.STRING`

This is the same general failure mode seen in earlier reports such as https://github.com/vitalik/django-ninja/issues/1424 and https://github.com/vitalik/django-ninja/issues/1357 annotations attached to wrapped or not-yet-fully-bound callables can blow up during signature inspection when they are evaluated in the wrong scope.

https://github.com/vitalik/django-ninja/issues/1652 is related background for the broader Python 3.14 annotation transition, but its `ModelSchemaMetaclass` path was already addressed separately. This PR covers the remaining `inspect.signature()` edge cases.

### DIFF
--- a/ninja/signature/utils.py
+++ b/ninja/signature/utils.py
@@ -18,10 +18,23 @@ __all__ = [
     "is_async",
 ]
 
+if version_info >= (3, 14):
+    import annotationlib
+
+    _FORWARDREF_KWARGS: dict[str, Any] = {
+        "annotation_format": annotationlib.Format.FORWARDREF
+    }
+    _STRING_KWARGS: dict[str, Any] = {
+        "annotation_format": annotationlib.Format.STRING
+    }
+else:
+    _FORWARDREF_KWARGS: dict[str, Any] = {}
+    _STRING_KWARGS: dict[str, Any] = {}
+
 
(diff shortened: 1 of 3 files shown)


joergbaier:
cc @vitalik this is ready for workflow approval

joergbaier:
@vitalik can we get a `1.6.3` release? 🙏

PR #1725, opened 2026-04-29
docs: clarify Http404 handler scope in error handling guide

### LINKED ISSUE #1390: [BUG] 404 handler override not working as expected
**Describe the bug**

I have a simple demo Django-Ninja app working thanks to a helpful onboarding documentation on the related docs.

All of the endpoints in my simple demo work as expected.

However, the following code in my urls.py file does not produce the expected result based on the docs:

```
from django.http import Http404
```
```
@api.exception_handler(Http404)
def page_not_found(request, exc):
    return api.create_response(
        request,
        {"message": "Please retry later"},
        status=404,
    ) 
```

I expect my handler to pick up 404s when the server is running, but I get the default

Not Found
The requested resource was not found on this server.

Did a lot of research beforehand and tried a variety of overrides, but none worked-- so am documenting the method from the dn docs:
https://django-ninja.dev/guides/errors/

My code is here:
https://github.com/ErikPohl444/django_ninja_playground

Working control case:
http://127.0.0.1:8000/api/hello_html?name=you

Not working experimental case:
http://127.0.0.1:8000/api/this_endpoint_is_undefined

I've done similar captures of the 404 in Flask and in Django, but I'm hitting a wall here.

It is either:
a) something is buggy with d-n [unlikely]
b) something can be added to for newbies like me to the documentation [potentially]
c) my code is buggy [in which case I might want to make a PR with doc changes --see b -- for folks like me :) ]

 - Python version: [e.g. 3.6]
 Python 3.12.0

 - Django version: [e.g. 4.0]
 Django==5.1.4

 - Django-Ninja version: [e.g. 0.16.2]
 django-ninja==1.3.0

 - Pydantic version: [e.g. 1.9.0]
pydantic==2.10.4

Thiago88pe:
The Http404 exception_handler only catches exceptions raised inside Ninja routes (e.g. via get_object_or_404). It does not intercept requests to undefined URLs, which are handled by Django directly. Fixes #1390

### DIFF
--- a/docs/docs/guides/errors.md
+++ b/docs/docs/guides/errors.md
@@ -72,8 +72,35 @@ Raised when request data does not validate
 Used to throw http error with status code from any place of the code
 
 #### `django.http.Http404`
- 
- Django's default 404 exception (can be returned f.e. with `get_object_or_404`)
+
+Django's default 404 exception (can be returned f.e. with `get_object_or_404`)
+
+> **Note:** The `@api.exception_handler(Http404)` only catches `Http404` exceptions raised **within Ninja route handlers**. It does **not** intercept requests to URLs that are simply not registered in Django's URL configuration — those never reach Ninja and are handled directly by Django.
+
+```python
+from django.shortcuts import get_object_or_404
+from django.http import Http404
+
+@api.exception_handler(Http404)
+def handle_404(request, exc):
+    return api.create_response(request, {"message": "Not found"}, status=404)
+
+# This WILL trigger the handler above
+@api.get("/items/{item_id}")
+def get_item(request, item_id: int):
+    return get_object_or_404(Item, id=item_id)
+
+# Requests to undefined URLs (e.g. /api/undefined-route)
+# will NOT trigger this handler — Django handles them before reaching Ninja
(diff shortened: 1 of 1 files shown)


mbaragiola (commented):
**Warnings**

- `docs/docs/guides/errors.md:78`: “only catches” exceptions raised “within Ninja route handlers” is narrower than the implementation. Exceptions raised during other Ninja operation-processing phases, including authentication, are also sent through `api.on_exception`. The boundary is Ninja operation processing versus a URL that Django never resolves to a Ninja view.
- `docs/docs/guides/errors.md:97`: please mention that Django uses a custom `handler404` only when `DEBUG=False`; with `DEBUG=True`, Django serves its technical 404 response instead.

**Positives**

- The main distinction is correct: an unresolved Django URL never enters Ninja’s exception machinery.
- The `get_object_or_404` example correctly exercises the registered Ninja `Http404` handler.

**Verification**

I reviewed the diff against Django Ninja’s operation/exception flow and the surrounding documentation. GitHub currently reports no checks for this head; I did not run local tests for this documentation-only change.

Issue #1713, opened 2026-04-12
[BUG] `ninja.streaming.SSE` and `JSONL` are not `Generic[T]` - mypy reports "type is not generic and not indexable"


adonig:
### Describe the bug

The streaming type containers introduced in v1.6 (`ninja.streaming.SSE`,
`ninja.streaming.JSONL`) support `SSE[Schema]` syntax at runtime via
`__class_getitem__`, but the underlying `StreamFormat` base class does not
inherit from `typing.Generic`. As a result, mypy (and any other type checker
that doesn't special-case `__class_getitem__`) rejects the documented usage:

```
error: The type "type[SSE]" is not generic and not indexable  [misc]
```

This forces users with strict mypy configurations to add `# type: ignore[misc]`
to every streaming endpoint, which weakens the type-safety story that
django-ninja otherwise has.

### Versions

- django-ninja: 1.6.2
- mypy: 1.20.0
- Python: 3.14

### Reproduction

```python
from ninja import NinjaAPI, Schema
from ninja.streaming import SSE

api = NinjaAPI()


class Event(Schema):
    message: str


@api.get("/events", response=SSE[Event])  # mypy error here
async def stream(request):
    yield {"message": "hello"}
```

Run `mypy` on the file:

```
error: The type "type[SSE]" is not generic and not indexable  [misc]
```

Runtime works correctly - only static type checking fails.

### Root cause

In `ninja/streaming.py`:

```python
class StreamFormat:
    media_type: str

    def __class_getitem__(cls, item_type: type) -> "_StreamAlias":
        return _StreamAlias(cls, item_type)
```

`__class_getitem__` enables the runtime syntax, but mypy only treats a class
as subscriptable when it inherits from `typing.Generic[T]` (or one of the
special-cased stdlib generics).

### Suggested fix

Make `StreamFormat` generic over the item type. The change is small and
shouldn't affect runtime behavior:

```python
from typing import Generic, TypeVar

T = TypeVar("T")


class StreamFormat(Generic[T]):
    media_type: str
    # ... rest unchanged
```

`SSE` and `JSONL` then inherit the generic parameter automatically, and
`SSE[Event]` becomes a valid parameterized type for mypy.

Happy to send a PR if this approach is acceptable.

PR #1726, opened 2026-04-29
fix(streaming): make SSE and JSONL generic

depuk98:
## What does this PR do?

Makes `ninja.streaming.StreamFormat`-based response containers generic so static type checkers accept documented usage like `SSE[Schema]` and `JSONL[Schema]`.

## Why is this change needed?

Tracked in #1713.

`SSE[Schema]` and `JSONL[Schema]` work at runtime via `__class_getitem__`, but mypy rejects them because the streaming containers were not modeled as generics. This forces users with strict type checking to add ignores around valid code.

## How was this tested?

- Added mypy regression coverage in `tests/mypy_test.py` for both `JSONL[Payload]` and `SSE[Payload]`
- Ran `mypy ninja tests/mypy_test.py`
- Ran `pytest tests/test_streaming.py`

## Checklist

- [x] I've read the CONTRIBUTING.md
- [x] Tests pass locally
- [x] I've added tests where applicable
- [ ] I've updated documentation where applicable

### DIFF
--- a/ninja/streaming.py
+++ b/ninja/streaming.py
@@ -1,29 +1,32 @@
 import json
-from typing import Any, Dict
+from typing import Any, Dict, Generic, Type, TypeVar
 
 from ninja.responses import NinjaJSONEncoder
 
 __all__ = ["StreamFormat", "SSE", "JSONL"]
 
 
+T = TypeVar("T")
+
+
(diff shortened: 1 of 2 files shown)


vitalik:
Looks like this is done - please check and reopen if you still see issue

PR #1727, opened 2026-04-30
fix: handle aliases with dashes in DjangoGetter attribute resolution

### LINKED ISSUE #1660: TemplateSyntaxError: Invalid character ('-') in variable name: 'CloudFront-Policy'
Hi I want to connect some S3 service and the JSON needs Felds with - in the name.

```python
class CookiesSchema(Schema):
    CloudFront_Policy: str = Field(..., alias="CloudFront-Policy")
```

How can I return {"CloudFront-Policy": "abcde"}

I get this Error:
Error extracting attribute: TemplateSyntaxError: Invalid character ('-') in variable name: 'CloudFront-Policy' [type=get_attribute_error, input_value=<DjangoGetter: CookiesSch...air_Id='KEDIJMZ6J0APU')>, input_type=DjangoGetter]
    For further information visit https://errors.pydantic.dev/2.12/v/get_attribute_error

Thiago88pe:
Fix TemplateSyntaxError when a Schema field uses an alias containing a dash (e.g. alias="CloudFront-Policy").

DjangoGetter.__getattr__ was falling through to Django's template Variable(key).resolve() even for keys that are invalid template variable names. Django's parser rejects dashes, raising TemplateSyntaxError instead of a catchable error.

The fix adds a guard using re.compile(r"^[\w.]+$") to skip Variable.resolve for keys that contain characters outside the set Django Template Variables accept (word chars and dots). Dotted paths like "boss.name" continue to work normally.

Fixes: aliases with - in name (e.g. AWS CloudFront cookies headers). Fixes #1660

### DIFF
--- a/ninja/router.py
+++ b/ninja/router.py
@@ -151,8 +151,8 @@ class BoundRouter:
                         )
 
                 # Apply tags inheritance
-                if operation.tags is None and self.tags is not None:  # type: ignore[has-type]
-                    operation.tags = self.tags  # type: ignore[has-type]
+                if operation.tags is None and self.tags is not None:
+                    operation.tags = self.tags
 
                 # Apply decorators (fresh application - no tracking needed)
                 for decorator, mode in effective_decorators:
--- a/ninja/schema.py
+++ b/ninja/schema.py
@@ -18,6 +18,7 @@ dotted attributes and resolver methods. For example::
 
 """
 
+import re
 import warnings
 from typing import (
     Any,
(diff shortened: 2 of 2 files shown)


mbaragiola (commented):
**Blockers**

- `ninja/schema.py:81`: Django’s `Variable` accepts dashed lookup components and can resolve `Variable("display-name")` through `obj["display-name"]` on a mapping-style object. The new guard raises before that valid fallback, so dashed aliases regress for non-`dict` objects implementing `__getitem__`.
- `ninja/schema.py:49`: the regex is not an accurate safety predicate for `Variable`. `\w` permits a leading underscore, but Django rejects variables or lookup components beginning with `_` by raising `TemplateSyntaxError`; `_private` therefore still reaches the exception this change is intended to avoid.

**Test gap**

- This behavioral change adds no regression coverage. Please cover the original dashed-alias report, mapping-style fallback, supported dotted aliases, and the leading-underscore failure mode.

**Positive**

- Avoiding uncaught template-parser errors for unsupported aliases is a useful goal, but it should preserve Django’s valid lookup semantics rather than duplicating them with a partial regex.

**Verification**

I reviewed the getter path and directly confirmed with Django 5.2 that `Variable("display-name")` resolves a mapping key while `Variable("_private")` raises `TemplateSyntaxError`. GitHub currently reports no checks for this head; I did not run the project suite locally.

PR #1729, opened 2026-05-15
update NinjaGenerateJsonSchema to handle the MISSING sentinel

Ksauder:
Small update to the custom schema generator method `NinjaGenerateJsonSchema.default_schema` so it can handle the new pydantic `MISSING` sentinel. The changes here reflect the [changes in pydantic](https://github.com/pydantic/pydantic/blob/ebc4652df990affb79bb3c3db960fb8d2a2b4308/pydantic/json_schema.py#L1183).

This does not address any integration of `MISSING` into `ModelSchema` django fields as done here: https://github.com/vitalik/django-ninja/pull/1656

### DIFF
--- a/.github/workflows/test_full.yml
+++ b/.github/workflows/test_full.yml
@@ -13,6 +13,7 @@ jobs:
       matrix:
         python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', '3.12', '3.13', '3.14']
         django-version: ['<3.2', '<3.3', '<4.2', '<4.3', '<5.1', '<5.2', '<5.3', '<6.1']
+        pydantic-version: ['']
         exclude:
           - python-version: '3.7'
             django-version: '<5.1'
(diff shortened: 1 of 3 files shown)


alexandre-fundcraft:
@Ksauder will you continue with this feature?

Ksauder:
@alexandre-fundcraft I have not seen any movement from the maintainer, so no. To get around this issue I converted all my `ModelSchemas` to explicit `Schemas`, use a custom type to mark fields as `MISSING`, and monkeypatched `ninja.schema.NinjaGenerateJsonSchema` and `ninja.openapi.schema.NinjaGenerateJsonSchema` with a permutation of the changes in this PR.

alexandre-fundcraft:
@Ksauder Thanks for your reply!

vitalik:
Hi @alexandre-fundcraft @alexandre-fundcraft - I'm a bit hesitant that it's inside experimental module

Maybe make import like 

'''
try:
    from pydantic import MISSING
except ImportError:
    from pydantic.experimental.missing_sentinel import MISSING
'''

?

Ksauder:
That's a good point, I can try both imports before defaulting to `MISSING = None`

Ksauder:
And as far as it being experimental, the underlying [PEP 661](https://peps.python.org/pep-0661/#add-a-single-new-sentinel-value-such-as-missing-or-sentinel) looks to have an implementation underneath it now with [sentinel](https://docs.python.org/3.15/builtins/functions.html#sentinel)

PR #1735, opened 2026-06-04
Fix empty multipart body form fields

### LINKED ISSUE #1703: _MultiPartBodyModel raises IndexError on empty form fields
params/models.py around line 185 does data[0] and data[-1] on the raw form value. When a field is present but empty (field=), data is an empty string and indexing raises IndexError.

Happens on any multipart POST where a string-annotated field is submitted empty. A len check before indexing would fix it.

cyphercodes:
## Summary
- Avoid indexing multipart body form data when a submitted string field is empty.
- Preserve the existing quote-wrapping behavior while allowing empty strings to parse as JSON string bodies.
- Add regression coverage for a multipart request with a `Body` string field and file upload.

## Tests
- `pytest -q tests/test_forms_and_files.py tests/test_params_models.py`
- `ruff format --check ninja tests`
- `ruff check ninja tests`
- `git diff --check`

Fixes #1703

### DIFF
--- a/ninja/params/models.py
+++ b/ninja/params/models.py
@@ -183,7 +183,11 @@ class _MultiPartBodyModel(BodyModel):
         for name, annotation in cls.__ninja_body_params__.items():
             if name in request.POST:
                 data = request.POST[name]
-                if annotation is str and data[0] != '"' and data[-1] != '"':
+                if (
+                    annotation is str
+                    and not data.startswith('"')
+                    and not data.endswith('"')
+                ):
                     data = f'"{data}"'
                 req.body = data.encode()
                 results[name] = get_request_data(req, api, path_params)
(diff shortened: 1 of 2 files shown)


mbaragiola (commented):
**Positives**

- Replacing direct first/last-character indexing with `startswith()`/`endswith()` removes the empty-string crash while preserving the existing quoting behavior.
- The regression test exercises the real multipart path with both an empty body field and a file upload, and verifies the decoded response.

**Residual risk**

- The test is focused on the reported empty-string case; one-sided quotes and other escaping cases retain the pre-existing behavior and are outside this patch.

**Verification**

I reviewed the multipart parsing path and regression test against `master`. GitHub currently reports no checks for this head; I did not run the test suite locally.

PR #1737, opened 2026-06-09
Confirm Django 6.1 support.

felixxm:


felixxm:
Something that we "early adopters" need to fix dependency resolution for Django 6.1 :package:  Let me know if I can help in any way and thanks for creating such a great package :1st_place_medal:

felixxm:
@vitalik Wow that was fast :racing_car: thanks :+1:

PR #1746, opened 2026-07-15
fix: create_schema crashes on models with a ForeignObject reverse relation

### LINKED ISSUE #1530: Django Ninja fails with ForeignObject fields: AttributeError 'ForeignObjectRel' object has no attribute 'help_text'
Django Ninja fails to generate ModelSchema when using Django's `ForeignObject` field, which is the recommended workaround for composite primary key relationships in [Django 5.2 documentation](https://docs.djangoproject.com/en/5.2/topics/composite-primary-key/#composite-primary-keys-and-relations).

**Error:**
```
AttributeError: 'ForeignObjectRel' object has no attribute 'help_text'
```

**The models:**

Note:  I'm don't using the  composite primary key in this example.

```python
from django.db import models

class Order(models.Model):
    # ... (fields can be omitted for brevity)
    pass

class Product(models.Model):
    # ... (fields can be omitted for brevity)
    pass

class OrderDetail(models.Model):
    order = models.ForeignObject(
        Order,
        on_delete=models.CASCADE,
        from_fields=["order_id"],
        to_fields=["id"],
        related_name="details"
    )
    product = models.ForeignKey(Product, on_delete=models.CASCADE)
    quantity = models.PositiveIntegerField()
```

**The endpoints:**

```python
from ninja import ModelSchema, NinjaAPI
from .models import Order

api = NinjaAPI()


class OrderSchema(ModelSchema):
    class Meta:
        model = Order
        fields = "__all__"


@api.get("/orders", response=list[OrderSchema])
def list_orders(request):
    return Order.objects.all()
```

Django Ninja cannot handle `ForeignObject` fields because `ForeignObjectRel` objects don't have the same attributes as `ForeignKey` relations.

A workaround is to explicitly define the fields in the ModelSchema, but I want to use the `fields = "__all__"`  in the schema definition.

## Reproduction

A minimal reproduction case is available at:  https://github.com/blasferna/django-ninja-foreignobject-issue

To reproduce:
1. Clone the reproduction repository
2. Run `uv sync`
3. Run `uv run manage.py runserver`

The error occurs immediately when Django tries to import the API endpoints.

## Versions:

 - Python version: 3.12
 - Django version: 5.2.5
 - Django-Ninja version: 1.4.3
 - Pydantic version: 2.11.7

shivanij1203:
## Summary

Generating a `ModelSchema` / calling `create_schema()` for a model that is the target of a `ForeignObject` relation crashes with:

```
AttributeError: 'ForeignObjectRel' object has no attribute 'help_text'
```

`SchemaFactory._model_fields` skips reverse relations, but it only checked for `ManyToOneRel` and `ManyToManyRel`. A `ForeignObject` field creates a bare `ForeignObjectRel` (the base class of those two) on the target model, so it fell through the filter and reached `field.help_text` in `get_schema_field`, which a reverse relation does not have.

Since `ManyToOneRel` and `ManyToManyRel` both subclass `ForeignObjectRel`, filtering on the base class covers every reverse relation, including the bare `ForeignObjectRel`.

Fixes #1530

## Changes

- `ninja/orm/factory.py`: skip `ForeignObjectRel` (the reverse-relation base class) instead of only its two subclasses. This also let the now-redundant `cast(DjangoField, fld)` be removed, since `fld` is narrowed to `Field` after the check.
- `tests/test_orm_relations.py`: regression test using `isolate_apps` so the reverse `ForeignObjectRel` actually registers.

## Testing

- `pytest tests/test_orm_relations.py tests/test_orm_metaclass.py tests/test_models.py tests/test_django_models.py` passes.
- The new test fails against `main` with the exact `AttributeError: 'ForeignObjectRel' object has no attribute 'help_text'`, and passes with the fix.
- `ruff format --check`, `ruff check`, and `mypy ninja/orm/factory.py` are clean.

Minimal reproduction (from the issue): a model `OrderDetail` with `models.ForeignObject(Order, ..., related_name="details")`, then `create_schema(Order)`.

### DIFF
--- a/ninja/orm/factory.py
+++ b/ninja/orm/factory.py
@@ -1,8 +1,8 @@
 import itertools
-from typing import Any, Dict, Iterator, List, Optional, Set, Tuple, Type, Union, cast
+from typing import Any, Dict, Iterator, List, Optional, Set, Tuple, Type, Union
 
 from django.db.models import Field as DjangoField
-from django.db.models import ManyToManyRel, ManyToOneRel, Model
+from django.db.models import ForeignObjectRel, Model
 from pydantic import create_model as create_pydantic_model
 
 from ninja.errors import ConfigError
@@ -158,10 +158,11 @@ class SchemaFactory:
     def _model_fields(self, model: Type[Model]) -> Iterator[DjangoField]:
         "returns iterator with all the fields that can be part of schema"
         for fld in model._meta.get_fields():
-            if isinstance(fld, (ManyToOneRel, ManyToManyRel)):
-                # skipping relations
+            if isinstance(fld, ForeignObjectRel):
+                # skipping reverse relations (ManyToOneRel, ManyToManyRel, and the
+                # bare ForeignObjectRel produced by a ForeignObject field)
                 continue
-            yield cast(DjangoField, fld)
+            yield fld
 
 
 factory = SchemaFactory()
--- a/tests/test_orm_relations.py
+++ b/tests/test_orm_relations.py
@@ -1,4 +1,5 @@
 from django.db import models
+from django.test.utils import isolate_apps
 
 from ninja import NinjaAPI
 from ninja.orm import create_schema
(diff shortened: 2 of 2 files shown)


mbaragiola (commented):
**Positives**

- Generalizing the reverse-relation filter to `ForeignObjectRel` covers the bare reverse object while retaining the existing `ManyToOneRel` and `ManyToManyRel` behavior through inheritance.
- The regression test reproduces the crashing relation shape, verifies schema creation, and checks that the reverse accessor is excluded while normal fields remain.

**Residual risk**

- The new test covers the default depth path only, although the shared field-selection path makes a depth-specific regression unlikely.

**Verification**

I reviewed the relation hierarchy, schema field-selection path, and regression test against `master`. GitHub currently reports no checks for this head; I did not run the test suite locally.

PR #1747, opened 2026-07-15
fix: ModelSchema doesn't apply alias_generator to foreign key fields

### LINKED ISSUE #1691: [BUG] camelCase not supported for relations on ModelSchema
camelCase not converted for foreignkeys in ModelSchema

 - Python version: 3.12.8
 - Django version: 5.2.8
 - Django-Ninja version: 1.4.3
 - Pydantic version: 2.12.5

```python
from django.db import models
class DbModelA(models.Model):
    id = models.BigAutoField(auto_created=True, primary_key=True)

class DbModelB(models.Model):
    id = models.BigAutoField(auto_created=True, primary_key=True)
    model_a = models.ForeignKey(DbModelA, on_delete=models.CASCADE)
    my_other_property = models.CharField()

from pydantic import ConfigDict
from pydantic.alias_generators import to_camel
CAMEL_CONFIG = ConfigDict(
    alias_generator=to_camel,
    validate_by_name=True,
    validate_by_alias=True,
)


from ninja import ModelSchema
class SchemaModelB(ModelSchema):
    model_config = CAMEL_CONFIG
    class Meta:
        model = DbModelB
        fields = [
            "id",
            "model_a",
            "my_other_property"
        ]

from ninja import Router
router = Router(by_alias=True)
@router.get("/endpoint/{model_id}", response=SchemaModelB)
def get_model_b(request, model_id: int):
    return DbModelB.objects.filter(id=model_id).first()
```

when the `/endpoint/{model_id}` is called, the response returns a:

```
{
    "id": "...",
    "model_a_id": ..., # this is in snake_case
    "myOtherProperty": "...", # this is in camelCase
}
```

Attempting to put a custom field with the camelCase alias results in a serialization error since we're passing in the model object

```python
from pydantic import Field
class SchemaModelB(ModelSchema):
    model_config = CAMEL_CONFIG
    class Meta:
        model = DbModelB
        fields = [
            "id",
            "model_a",
            "my_other_property"
        ]
    model_a: int = Field(alias="modelAId") # invalid since object is not id
    # implicitly it's the following when the schema is generated:
    # model_a: int = Field(alias="model_a_id")

# "Pydantic validation error for /endpoint/1"
# 'msg': 'Input should be a valid integer', 'input': <DbModelA: DbModelA object (1)>
```

when the endpoint is called, it should return:

```
{
    "id": "...",
    "modelAId": ..., # this is in snake_case
    "myOtherProperty": "...", # this is in camelCase
}
```

shivanij1203:
## Summary

When a `ModelSchema` sets an `alias_generator` (e.g. `to_camel`), regular fields are aliased but a foreign key's generated `_id` field is not — it stays snake_case while everything else is transformed:

```json
{
  "id": "...",
  "model_a_id": ...,        // stays snake_case
  "myOtherProperty": "..."  // camelCased
}
```

The reason is that relation fields set their alias **explicitly** (to the `_id` attname) in `get_schema_field`. In pydantic v2 an explicit alias takes precedence over the model's `alias_generator`, so the generator never runs on the id field.

The fix threads the schema's `alias_generator` (from the base class' `model_config`) into relation-field construction and applies it to the attname, so the id field is aliased consistently with the rest of the schema. When no generator is configured, behavior is unchanged (`model_a_id`).

Fixes #1691

## Changes

- `ninja/orm/fields.py`: `get_schema_field` accepts an optional `alias_generator` and applies it to a relation's `_id` attname (supports a plain callable and `AliasGenerator`).
- `ninja/orm/factory.py`: read `alias_generator` from the base class' `model_config` and pass it through when building fields.
- `tests/test_orm_relations.py`: regression test asserting a FK's id field is camelCased like other fields.

## Testing

- `pytest tests/test_orm_relations.py tests/test_orm_metaclass.py tests/test_models.py` passes.
- The new test fails against `main` (the FK alias is `target_id` instead of `targetId`) and passes with the fix.
- `ruff format --check`, `ruff check`, and `mypy ninja` are all clean.

### DIFF
--- a/ninja/orm/factory.py
+++ b/ninja/orm/factory.py
@@ -60,12 +60,15 @@ class SchemaFactory:
             if optional_fields == "__all__":
                 optional_fields = [f.name for f in model_fields_list]
 
+        alias_generator = getattr(base_class, "model_config", {}).get("alias_generator")
+
         definitions = {}
         for fld in model_fields_list:
             python_type, field_info = get_schema_field(
                 fld,
                 depth=depth,
                 optional=optional_fields and (fld.name in optional_fields),
+                alias_generator=alias_generator,
             )
             definitions[fld.name] = (python_type, field_info)
 
--- a/ninja/orm/fields.py
+++ b/ninja/orm/fields.py
@@ -5,7 +5,7 @@ from uuid import UUID
 
 from django.db.models import ManyToManyField
 from django.db.models.fields import Field as DjangoField
-from pydantic import IPvAnyAddress
+from pydantic import AliasGenerator, IPvAnyAddress
 from pydantic.fields import FieldInfo
 from pydantic_core import PydanticUndefined, core_schema
 
@@ -113,9 +113,25 @@ def create_m2m_link_type(type_: Type[TModel]) -> Type[TModel]:
     return M2MLink
 
 
+def _apply_alias_generator(alias_generator: Any, attname: str) -> str:
+    """Apply the schema's ``alias_generator`` to a relation's ``_id`` attname.
+
+    Relation fields set their alias explicitly (to the ``_id`` attname), which
+    otherwise bypasses the schema's ``alias_generator`` so the generated field
+    keeps its snake_case name while every other field is transformed (see #1691).
+    """
+    if isinstance(alias_generator, AliasGenerator):
+        return alias_generator.alias(attname) if alias_generator.alias else attname
+    return str(alias_generator(attname))
+
+
 @no_type_check
 def get_schema_field(
-    field: DjangoField, *, depth: int = 0, optional: bool = False
+    field: DjangoField,
+    *,
+    depth: int = 0,
+    optional: bool = False,
+    alias_generator: Any = None,
 ) -> Tuple:
     "Returns pydantic field from django's model field"
     alias = None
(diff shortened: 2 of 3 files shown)


mbaragiola (commented):
**Blocker**

- `ninja/orm/fields.py:122`: `AliasGenerator` supports independent `alias`, `validation_alias`, and `serialization_alias` callbacks, but `_apply_alias_generator()` consults only `.alias`. If `.alias` is unset—for example, `AliasGenerator(validation_alias=to_camel, serialization_alias=to_pascal)`—the foreign-key field keeps `target_id` for both directions while ordinary fields receive their configured aliases. Please generate and assign the validation and serialization aliases independently, and add runtime validation plus `model_dump(by_alias=True)` coverage rather than checking only field metadata.

**Positives**

- The common callable `alias_generator=to_camel` case is addressed.
- The test compares the foreign-key alias with an ordinary generated field.

**Verification**

I reviewed the factory/field path and Pydantic’s `AliasGenerator.generate_aliases()` behavior. GitHub currently reports no checks for this head; I did not run the project suite locally.

shivanij1203:
Good catch, thanks. Reworked this so the `_id` field goes through pydantic's own `AliasGenerator.generate_aliases()` instead of only reading `.alias`. `alias`, `validation_alias`, and `serialization_alias` are now generated and assigned independently, and each direction the generator leaves unset falls back to the `_id` attname (so a single-callback generator like `alias_generator=to_camel` still covers both directions, unchanged from before).

For the `AliasGenerator(validation_alias=to_camel, serialization_alias=to_pascal)` case you flagged, the `author_id` field now gets `authorId` for validation and `AuthorId` for serialization rather than staying `author_id` in both.

Also swapped the test from metadata-only checks to a behavioral one: it validates via the camelCase validation alias (`model_validate`) and asserts the PascalCase serialization alias comes back out of `model_dump(by_alias=True)`, alongside a plain field for comparison. `mypy ninja` and `ruff` (format + check) are clean.

PR #1481, opened 2025-06-11
Update CI and packaging for Django 5.2 and Python 3.13 support

mahdirahimi1999:
## Description

This PR adds support for Django 5.2 and Python 3.13:

- Updated GitHub Actions test matrix:
  - Added Python 3.13
  - Added Django `<5.3`
- Set default installed Django version to `>=5.2` in CI workflows
- Updated `pyproject.toml`:
  - Added Python 3.13 and Django 5.2 to classifiers
  - Set Django version range to `>=3.1, <6.0`

Keeps the project up to date with the latest stable versions of Python and Django.

PR #1749, opened 2026-07-27
Drop Django upper version constraint

Stormheg:
The upper version constraint for Django was introduced in https://github.com/vitalik/django-ninja/pull/1481 and shipped in release v1.5.0. I don't believe there was an intentional reason why it was added - there are no remarks specifically calling out that change and why it was made.

This constraint makes it unnecessarily difficult to test projects and other re-usable Django libraries against upcoming Django versions, because dependency resolution will fail or [install v1.4.5 which was the last version that had no upper bound](https://github.com/vitalik/django-ninja/blob/b7ac09820573f0a1802000134b1e49684cc8d7d1/pyproject.toml#L45).

See also this blog post about the topic: https://iscinumpy.dev/post/bound-version-constraints/

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,7 @@ classifiers = [
     "Topic :: Internet :: WWW/HTTP",
 ]
 
-requires = ["Django >=3.1, <6.2", "pydantic >=2.0,<3.0.0"]
+requires = ["Django >=3.1", "pydantic >=2.0,<3.0.0"]
 description-file = "README.md"
 requires-python = ">=3.7"
 


joergbaier (approved):

tony (approved):
Disregard my earlier message here. #1737 has Django 6.1 handled 🙇.

anderspetersson:
With Django 6.1 released, this should be merged ASAP and a new release should be cut.

Upgrading django to 6.1 in a project that depends on django-ninja>=1.5.1 gives the following error when running `uv sync`:


```
  × No solution found when resolving dependencies:
  ╰─▶ Because django-ninja>=1.5.1 depends on django>=3.1,<6.1 and your project depends on django==6.1, we can
      conclude that your project and django-ninja>=1.5.1 are incompatible.
      And because your project depends on django-ninja>=1.5.1, we can conclude that your project's requirements are
      unsatisfiable.
```

PR #1753, opened 2026-08-07
Fix Flit PEP 621 metadata & pytest setup for local contributions

LexGlu:
Running flit install --deps develop --symlink on newer versions of
Flit (3.8+) fails because [tool.flit.metadata] is deprecated in favor
of standard PEP 621 metadata.

This PR updates pyproject.toml to PEP 621 standards and fixes a
couple of minor local development setup issues:

• PEP 621 Migration: Converted [tool.flit.metadata] to standard
[project] fields in pyproject.toml and specified [tool.flit.module]
name = "ninja".
• Pytest Isolation: Added -p no:postgresql to pytest addopts so
running tests locally doesn't fail when unrelated third-party plugins
(like pytest-postgresql) are present in global site-packages.
• Type Annotations & Pre-commit: Added explicit tags attribute typing
on Operation to fix a mypy pre-commit hook failure, and added a
coverage pragma on is_async to keep coverage at 100%.

### Testing

• Ran flit install --deps develop --symlink (passes cleanly).
• Ran make lint, make test, and make test-cov (797 tests passing,
100% coverage).

### DIFF
--- a/ninja/operation.py
+++ b/ninja/operation.py
@@ -51,6 +51,8 @@ __all__ = ["Operation", "PathView", "ResponseObject"]
 
 
 class Operation:
+    tags: Optional[List[str]] = None
+
     def __init__(
         self,
         path: str,
--- a/ninja/router.py
+++ b/ninja/router.py
@@ -151,8 +151,8 @@ class BoundRouter:
                         )
 
                 # Apply tags inheritance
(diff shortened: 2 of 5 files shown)


loqs:
Please consider splitting the three different changes types into three separate atomic commits for easier review and cherry-picking.

PR #1754, opened 2026-08-12
Add HTTP QUERY method support (RFC 10008)

caspel26:
Adds support for the HTTP `QUERY` method, standardized by [RFC 10008](https://www.rfc-editor.org/rfc/rfc10008) — safe and idempotent like `GET`, but carries a request body like `POST`. OpenAPI 3.2 added an official `query` PathItem field to match.

This is useful for search/filter endpoints whose criteria don't fit cleanly into a query string (nested filter trees, large ID lists, etc.) without giving up GET's safe/idempotent/cacheable semantics the way a `POST /search` workaround does.

## Usage

```python
class SearchRequest(Schema):
    tags: Optional[List[str]] = None
    min_price: Optional[float] = None
    ids: Optional[List[int]] = None
    limit: int = 50

@api.query("/products/search", response=List[Product])
def search_products(request, payload: SearchRequest = Body(...)):
    ...
```

```python
# test client
response = client.query("/products/search", json={"tags": ["chair"], "limit": 20})
```

## Changes

- `Router.query()` / `NinjaAPI.query()` — new decorator, mirrors `.patch()`
- CSRF: QUERY requests are now exempted from CSRF checks (same mechanism as `@csrf_exempt`), since Django's `CsrfViewMiddleware` doesn't know about QUERY and would otherwise wrongly block it under cookie-based auth
- `Body()`/`Form()`/`File()` params now work on QUERY requests (added to the existing `FIX_REQUEST_FILES_METHODS` compatibility shim used for PUT/PATCH/DELETE)
- OpenAPI schema version bumps from `3.1.0` to `3.2.0` only when a QUERY route is registered — zero impact on APIs that don't use it
- `.query()` added to the sync and async test clients
- Tests for dispatch, CSRF exemption, and schema versioning
- Docs updated in the HTTP methods guide

## Testing

- `pytest .` — full suite green
- `pytest --cov=ninja --cov-report term-missing tests` — 100% coverage maintained
- `ruff format --check`, `ruff check`, `mypy ninja` — clean

### DIFF
--- a/docs/docs/guides/input/operations.md
+++ b/docs/docs/guides/input/operations.md
@@ -9,10 +9,11 @@ An `operation` can be one of the following [HTTP methods](https://developer.mozi
 - PUT
 - DELETE
 - PATCH
+- QUERY
 
 **Django Ninja** comes with a decorator for each operation:
 
-```python hl_lines="1 5 9 13 17"
+```python hl_lines="1 5 9 13 17 21"
 @api.get("/path")
 def get_operation(request):
     ...
@@ -32,7 +33,15 @@ def delete_operation(request):
 @api.patch("/path")
 def patch_operation(request):
     ...
+
+@api.query("/path")
+def query_operation(request):
+    ...
 ```
+
+`QUERY` is a safe, idempotent method (like `GET`) that also carries a request
+body (like `POST`), making it a good fit for complex search criteria that
+don't fit cleanly into a URL's query string.
 
 See the [operations parameters](../../reference/operations-parameters.md)
 reference docs for information on what you can pass to any of these decorators.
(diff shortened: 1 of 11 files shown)


vitalik:
Hi @caspel26 

```
    if self.uses_query_method:
            self["openapi"] = "3.2.0"
```

Can this be not hardcoded?

caspel26:
Hi @vitalik,
thank you for your feedback first of all! You're right, the previous version had two problems: the 3.1.0/3.2.0 strings were duplicated literals, and openapi_extra could never actually override the "openapi" key since it was already set on the dict before the extras-merge loop ran.

Pushed a fix: the version is now computed via get_openapi_version() — it returns 3.2.0 only when the schema actually has a QUERY operation registered, 3.1.0 otherwise — and an explicit NinjaAPI(openapi_extra={"openapi": "..."}) now genuinely overrides the auto-detected value (previously silently ignored). Added a regression test (test_openapi_version_explicit_override_wins) covering that override path. you can check the fix by yourself. Let me know if there's something else worth flagging!

PR #1756, opened 2026-08-14
Fix deferred annotations on paginated routes

### LINKED ISSUE #1717: [BUG] pydantic.errors.PydanticUserError: `QueryParams` is not fully defined; you should define `ModelFilterSchema`, then call `QueryParams.model_rebuild()`.
**Describe the bug**
There seems to be an issue with paginated and filtered endpoints.

While this works:
```python
@api.get("/models", response=List[ModelSchema])
def list_models(request, filters: ModelFilterSchema = Query(...)) -> List[ModelSchema]:
    models = Model.objects.all()
    models = filters.filter(models)
    return models
```

And this also works:
```python
@api.get("/models", response=List[ModelSchema])
@paginate
def list_models(request):
    models = Model.objects.all()
    return models
```

This one fails:
```python
@api.get("/models", response=List[ModelSchema])
@paginate
def list_models(request, filters: ModelFilterSchema = Query(...)) -> List[ModelSchema]:
    models = Model.objects.all()
    models = filters.filter(models)
    return models
```

```pydantic.errors.PydanticUserError: `QueryParams` is not fully defined; you should define `ModelFilterSchema`, then call `QueryParams.model_rebuild()`.```

(Note, "Model" is just a placeholder here)

**Versions:**
 - Python version: 3.10
 - Django version: 5.2.13
 - Django-Ninja version: 1.6.2
 - Pydantic version: 2.12.5

tunglambk:
## Summary

- Resolve deferred endpoint annotations using the original callable's globals when decorators wrap the endpoint.
- Preserve the decorated signature, including pagination's contributed parameters.
- Add regression coverage for a paginated route with a deferred `FilterSchema` query annotation.

## Root cause

`inspect.signature()` follows `__wrapped__` and returned the original endpoint annotations, but `get_typed_signature()` resolved them against the pagination wrapper's `__globals__`. User-defined schema names are absent from that module, so Pydantic received an unresolved forward reference and raised `QueryParams is not fully defined` at request time.

## Validation

- `python -m pytest -q tests/test_pagination_annotations.py tests/test_pagination.py tests/test_pagination_async.py tests/test_pagination_router.py tests/test_utils.py tests/test_decorators.py -k 'not test_test_async_pagination'` (45 passed, 1 deselected)
- `ruff format --check ninja/signature/utils.py tests/test_pagination_annotations.py`
- `ruff check ninja/signature/utils.py tests/test_pagination_annotations.py`
- `mypy ninja/signature/utils.py`
- `git diff --check`

The deselected database-backed async pagination test hangs in my local test environment; all five async pagination tests that do not use the database pass individually.

Fixes #1717


tunglambk:
Rebased onto current `master` (was 27 commits behind).

Re-ran the pagination, pagination-router, utils and decorators tests (45 passed), plus `ruff format --check`, `ruff check` and `mypy`: all pass. The full suite passes too (808 passed, 3 skipped).

Ready for review when you have time.

PR #1758, opened 2026-08-14
fix: skip all reverse relation types when building ModelSchema fields

### LINKED ISSUE #1530: Django Ninja fails with ForeignObject fields: AttributeError 'ForeignObjectRel' object has no attribute 'help_text'

juneja-varun:
Fixes #1530.

Any model that's the target of a plain `ForeignObject` relation (Django's documented pattern for composite-key relations) crashes when django-ninja builds its schema fields: `AttributeError: 'ForeignObjectRel' object has no attribute 'help_text'`. The reverse-relation filter only checked for `ManyToOneRel`/`ManyToManyRel`, which covers `ForeignKey`-based relations but not the bare `ForeignObjectRel` a plain `ForeignObject` creates.

Fixed by checking `isinstance(fld, ForeignObjectRel)` instead, which covers every case the old check did plus the one it missed. Added a regression test using a real `ForeignObject` model pair, confirmed it fails with the original error on unpatched code.


juneja-varun:
Just checking in on this one, still green and up to date with master. Happy to make any changes if useful.

PR #1742, opened 2026-07-08
fix: combine router operations for matching paths

mbaragiola:
## Summary

- combine bound router operations that normalize to the same Django route before creating URL patterns
- preserve one URL pattern name per operation while sharing a combined dispatcher for matching paths
- add regression coverage for API-level `POST /users` plus router-level `PUT /users`
- keep `BoundRouter.urls_paths()` covered as a direct internal compatibility surface

Fixes #1546.

## Tests

- `pytest tests/test_api_instance.py tests/test_app.py -q` → 17 passed
- `pytest tests/test_app.py tests/test_api_instance.py tests/test_auth_global.py -q` → 19 passed
- `pytest . -q` → 798 passed, 3 skipped
- Python 3.12: `pytest --cov=ninja --cov-report term-missing tests -q` → 799 passed, 3 skipped; 100.00% coverage
- `mypy ninja` → success
- `ruff format --check ninja tests` → 157 files already formatted
- `ruff check ninja tests` → all checks passed

## Review

- Independent Codex review: APPROVED, no blocking findings.
- Independent Codex review after coverage fix: APPROVED, no findings.

## Notes

Multiple URL names for the same route still resolve to the same path, matching existing same-router behavior. Future non-blocking hardening could add explicit coverage for reverse URL names and mixed sync/async operations on the same merged route.

PR #1763, opened 2026-08-17
fix: dispatch same-path API and router operations by method

### LINKED ISSUE #1546: [BUG] Router api can not visited
**Describe the bug**
When I'm using the interface written by Ninja Router, I can't access it properly.

Demo is here: https://github.com/guozhenyi/django-ninja-bug-20250919

 - Python version: [e.g. 3.8]
 - Django version: [e.g. 4.2]
 - Django-Ninja version: [e.g. 1.4.3]

Note you can quickly get this by runninng this command:
```
git clone https://github.com/guozhenyi/django-ninja-bug-20250919.git
cd django-ninja-bug-20250919
python manage.py runserver
```
Use Postman or [Bruno](https://www.usebruno.com/), and visit  POST /api/users and PUT /api/users.

Create a Ninja router:

`apps/core/api.py`:

```python
from typing import Optional
from ninja import Router, Schema

router = Router()


class UserPutSchema(Schema):
    id: int
    username: str
    email: Optional[str] = None


@router.put("/users")
def user_update(request: HttpRequest, rdata: UserPutSchema):
    data = {"code": 200}

    return data

```

Create mysite/api.py:

```python
from typing import Optional
from ninja import NinjaAPI, Schema
from apps.core.api import router as core_router

api = NinjaAPI()


class UserPostSchema(Schema):
    username: str
    password: str
    email: Optional[str] = None


@api.post("/users")
def user_create(request: HttpRequest, rdata: UserPostSchema):
    data = {"code": 200}
    return data

api.add_router("", core_router)

```

Request result:

Image

Image

Normally, these should all OK.

mmdverse:
## Summary

Fixes #1546.

When a `NinjaAPI` operation and an operation from a mounted `Router` share the same normalized path but handle different HTTP methods, Django resolves the first URL pattern and never reaches the later router pattern. The first `PathView` therefore returns 405 for the other router's method.

This change merges bound `PathView` operations by normalized route before creating Django URL patterns. Each operation keeps its own named pattern for URL reversing, while every duplicate route pattern delegates to the same merged dispatcher.

## Regression coverage

Adds a regression test for:

- `POST /users` on the API
- `PUT /users` on a router mounted at the same path
- `DELETE /users` returning 405 with both `POST` and `PUT` in `Allow`

## Validation

- `pytest tests/test_router_reuse.py -q` → 27 passed
- `ruff format --check ninja tests`
- `ruff check ninja tests`
- `mypy ninja`
- `pytest . -q` → 798 passed, 3 skipped

### DIFF
--- a/ninja/main.py
+++ b/ninja/main.py
@@ -14,6 +14,7 @@ from typing import (
 
 from django.http import HttpRequest, HttpResponse
 from django.urls import URLPattern, URLResolver, reverse
+from django.urls import path as django_path
 from django.utils.module_loading import import_string
 
 from ninja.constants import NOT_SET, NOT_SET_TYPE
@@ -28,6 +29,7 @@ from ninja.openapi import get_schema
 from ninja.openapi.docs import DocsBase, Swagger
 from ninja.openapi.schema import OpenAPISchema
 from ninja.openapi.urls import get_openapi_urls, get_root_url
+from ninja.operation import PathView
 from ninja.parser import Parser
 from ninja.renderers import BaseRenderer, JSONRenderer
 from ninja.router import BoundRouter, Router, RouterMount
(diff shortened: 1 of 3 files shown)


mmdverse:
Hi! Following up on this fix for the same-path API/router method dispatch issue. The regression is covered by focused tests, and I’m happy to make any adjustments if you have feedback. Thanks for your time.

mmdverse:
Closing this in favour of #1742 — it covers the same case (API-level and router-level operations on the same path) and was opened earlier, so there's no reason to keep two PRs on it.

PR #1764, opened 2026-08-18
Preserve model metadata on custom ModelSchema fields

### LINKED ISSUE #1682: ModelSchema custom fields lose verbose_name title
## Bug description

When a `ModelSchema` field is explicitly overridden (e.g., to add validation or change defaults), the auto-generated `title` from the Django model's `verbose_name` is lost. This breaks translated field titles in the OpenAPI schema.

## Minimal reproduction

```python
from ninja import Field, ModelSchema

class Book(models.Model):
    title = models.CharField(_("Book title"), max_length=200)

# Without override — title is correctly set from verbose_name
class BookSchema(ModelSchema):
    class Meta:
        model = Book
        fields = ["title"]

BookSchema.model_fields["title"].title  # "Book title" ✓

# With override — title is lost
class BookCreateSchema(ModelSchema):
    title: str = Field(min_length=1)

    class Meta:
        model = Book
        fields = ["title"]

BookCreateSchema.model_fields["title"].title  # None ✗
```

## Root cause

In `ninja/orm/factory.py:72-76`, `custom_fields` completely overwrite the auto-generated `definitions` entry (which contains the `FieldInfo` with `title` from `verbose_name`) with no merging:

```python
if custom_fields:
    for fld_name, python_type, field_info in custom_fields:
        definitions[fld_name] = (python_type, field_info)  # title is lost here
```

The model-derived `FieldInfo` (with `title` and `description`) is already computed in `definitions[fld_name]` at this point, but is discarded entirely.

## Proposed fix

Preserve the model-derived `title` (and `description`) when the custom field doesn't explicitly set them:

```python
if custom_fields:
    for fld_name, python_type, field_info in custom_fields:
        if fld_name in definitions and isinstance(field_info, FieldInfo):
            _, model_field_info = definitions[fld_name]
            if field_info.title is None:
                field_info.title = model_field_info.title
            if field_info.description is None:
                field_info.description = model_field_info.description
        definitions[fld_name] = (python_type, field_info)
```

This is backward-compatible — fields with an explicit `title=` or `description=` are unaffected.

## Current workaround

Manually duplicating the `verbose_name` on every overridden field:

```python
class BookCreateSchema(ModelSchema):
    title: str = Field(min_length=1, title=_("Book title"))  # manual workaround
```

This is tedious and error-prone at scale (we have 170+ such overrides in our codebase).

---

Would you accept a PR for this fix? Happy to submit one.

tunglambk:
## Summary

- preserve a Django model field's `verbose_name` and `help_text` when a
  `ModelSchema` overrides that field with Pydantic `Field` constraints
- keep an explicitly configured custom title or description authoritative
- add a regression test covering both inherited and explicitly overridden
  metadata

## Root cause

`SchemaFactory.create_schema()` replaced the model-derived `FieldInfo` with the
custom field's `FieldInfo`. This correctly kept custom validation constraints,
but discarded the title and description already derived from the Django model.

The two metadata values are now merged into the custom `FieldInfo` before the
schema is created. Pydantic's merge semantics ensure explicitly provided custom
values still win.

## Validation

- `.venv/bin/pytest -q tests/test_orm_metaclass.py tests/test_orm_schemas.py`
  (28 passed)
- `.venv/bin/ruff format --check ninja/orm/factory.py tests/test_orm_metaclass.py`
- `.venv/bin/ruff check ninja/orm/factory.py tests/test_orm_metaclass.py`
- `git diff --check`

Fixes #1682


tunglambk:
Rebased onto current `master` (was 19 commits behind).

Re-ran the ORM schema tests (30 passed), plus `ruff format --check`, `ruff check` and `mypy`: all pass. The full suite passes too (808 passed, 3 skipped).

Ready for review when you have time.

